PR #61, opened 2020-06-19
Fix couple of typos

mandoonandy:
Little typos before I forget

### DIFF
--- a/golang/example.go
+++ b/golang/example.go
@@ -121,7 +121,7 @@ func main() {
 		return
 	}
 
-	fmt.Printf("Fetcehd %s\n", network.Name)
+	fmt.Printf("Fetched %s\n", network.Name)
 	networkUuid := networks[0].UUID
 
 	// --------------------------------------------------------------------------
--- a/shakenfist/client/main.py
+++ b/shakenfist/client/main.py
@@ -199,7 +199,7 @@ def network_events(ctx, network_uuid=None):
         filtered_events = []
         for e in events:
             filtered_events.append(filter_dict(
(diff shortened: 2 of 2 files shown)


golang/example.go
```diff
@@ -121,7 +121,7 @@ func main() {
 		return
 	}
 
-	fmt.Printf("Fetcehd %s\n", network.Name)
+	fmt.Printf("Fetched %s\n", network.Name)
```
mikalstill, line 124: I presume this example should move into the golang API client eventually?

shakenfist/client/main.py
```diff
@@ -199,7 +199,7 @@ def network_events(ctx, network_uuid=None):
         filtered_events = []
         for e in events:
             filtered_events.append(filter_dict(
-                n, ['timestamp', 'fqdn', 'operation', 'phase', 'duration', 'message']))
+                e, ['timestamp', 'fqdn', 'operation', 'phase', 'duration', 'message']))
```
mikalstill, line 202: Oh, good catch.

shakenfist/client/main.py
```diff
@@ -488,7 +488,7 @@ def instance_events(ctx, instance_uuid=None):
         for e in events:
             filtered_events.append(filter_dict(
                 e, ['timestamp', 'fqdn', 'operation', 'phase', 'duration', 'message']))
-        print(json.dumps({'networks': filtered_events},
+        print(json.dumps({'events': filtered_events},
```
mikalstill, line 491: Again, nice.

golang/example.go
mandoonandy: yes. doing that now.

PR #62, opened 2020-06-19
Move golang to separate repos

mandoonandy:


mikalstill:
I don't know how you end up with a conflict on a delete, so that's nice. That said, I think we should add a section to the readme file saying where to find the client library and terraform provider.

mandoonandy:
My bad... disregard

Issue #58, opened 2020-06-17
We need the concept of users, groups, and API auth


mikalstill:
So that we can implement resource ownership.

PR #66, opened 2020-06-19
Production web serving, auth, and ownership

mikalstill:
This pull request will eventually implement production grade web serving, authentication, and resource ownership. There are quite a few steps to get there though. This isn't ready for merge yet, but I'll upload it as I go along in case someone wants to make comments and I need to course correct.

First off: add a uniqifier to GCP instance names.
* This allows me to have more than one SF cluster on GCP at once for testing, which will be useful for CI when we get there. Its also needed at the moment now that we have more than one person working on SF.

Then: fix deployments.
* I have no idea how the grafana dashboard directory being missing got this far, given its been through CI a bunch of times.
* Add a missing file. This is a good example of where github based CI would have caught something which worked on my machine. My bad.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ If there is an existing library which does a thing, we use it. OpenStack suffere
 Deployment choices
 ------------------
 
-libvirt is the only supported hypervisor. Instances are specified to libvirt with simple templated XML. If your local requirements are different to what's in the template, you're welcome to change the template to meet your needs. If you're template changes break things, you're also welcome to debug what went wrong for yourself.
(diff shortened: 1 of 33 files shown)


mikalstill:
I considered using apache2 as the WSGI container for the API server, but having travelled through half of the conversion to a non-root account for the SF components I've decided it violates the "simplicity rule". Effectively you need to give SF so much control over hypervisor nodes that there is no point not running the API service as root for now. I guess this might change in the future, we'll see.

mikalstill:
Ok, it was a bit of a saga, but I now have flask APIs being served by gunicorn and CI passing locally. Next step, add JWT auth tokens.

mikalstill:
Note the comment in the commit message for 1ceb604... This is incomplete at this point as deployments will be broken for now.

mikalstill:
This chain of patches also Fixes #58.

PR #84, opened 2020-06-24
Set default API URL

mandoonandy:
This fixes an error when a namespace is specifed without an API URL.

  File "/usr/local/lib/python3.6/dist-packages/shakenfist/client/apiclient.py", line 142, in create_instance
    r = self._request_url('POST', self.base_url + '/instances',
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'


mikalstill:
LGTM.

PR #102, opened 2020-06-29
Api interface request

mandoonandy:
Primarily adding the /interface/<uuid> endpoint
Python client updated to utilise the endpoint.

Discovered that CLI autocomplete does not work. Work started to fix, but I think this is pushing Click beyond it's design. The CLI will now list the UUID options rather than barf an exception. (Will raise an issue to track this problem.)


mikalstill:
LGTM

PR #119, opened 2020-07-02
Separate namespace creation and metadata

mandoonandy:
Couple of issues caught in this branch, as per the commit messages.


mandoonandy:
I broke it. I'll fix tomorrow....

mandoonandy:
Couldn't leave broken tests. Too embarrassed.

PR #120, opened 2020-07-03
Namespace changes, metadata fixes, sf-client changes

mandoonandy:
This contains the commits from #119 since I have kept working.
I assume you can accept #119 then this one.


mandoonandy:
This now has the soft/hard deletion of metadata fix.

PR #137, opened 2020-07-08
Instance memory is measured in MB

mandoonandy:
Discussion topic: Should the CLI specify memory in MB or GB??

My previous changes regarding this assumed that the CLI would use MB. I even changed the README. I liked MB because it differentiated the CPU and memory spec to the casual user.

This PR discovered the CLI usage help was not changed. Also found the memory utilisation numbers was also broken for memory_mb.

As a compromise, the CLI will accept GB memory specification with a warning.

Take aim! Discuss!

### DIFF
--- a/shakenfist/client/main.py
+++ b/shakenfist/client/main.py
@@ -592,7 +592,7 @@ def _parse_spec(spec):
                   help=('Create an instance.\n\n'
                         'NAME:      The name of the instance.\n'
                         'CPUS:      The number of vCPUs for the instance.\n'
-                        'MEMORY:    The amount of RAM for the instance in GB.\n'
+                        'MEMORY:    The amount of RAM for the instance in MB.\n'
                         '\n'
                         'Options (may be repeated, must be specified at least once):\n'
                         '--network/-n:   The uuid of the network to attach the instance to.\n'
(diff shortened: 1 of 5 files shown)


shakenfist/client/main.py
```diff
@@ -634,6 +634,10 @@ def instance_create(ctx, name=None, cpus=None, memory=None, network=None, networ
     if len(disk) < 1 and len(diskspec) < 1:
         print('You must specify at least one disk')
 
+    if memory < 256:
+        print('WARNING: Assuming you have specified memory in GB\n')
+        memory *= 1024
+
```
mikalstill, line 640: This seems reasonable to me, but maybe we should file an issue to remove the special casing in a later release? v0.3 perhaps?

mikalstill: In fact, have https://github.com/shakenfist/shakenfist/issues/140

shakenfist/daemons/resources.py
```diff
@@ -48,12 +48,12 @@ def _get_stats():
...
 
-    # Memory info
+    # Memory info - libvirt returns memory in KiB
     memory_status = conn.getMemoryStats(
         libvirt.VIR_NODE_MEMORY_STATS_ALL_CELLS)
     retval.update({
-        'memory_max': memory_status['total'],
-        'memory_available': memory_status['free'],
+        'memory_max': memory_status['total'] // 1024,
+        'memory_available': memory_status['free'] // 1024,
```
mikalstill, line 56: Oh good catch. This will make the dashboards nicer. I wonder how it affects the scheduler (if at all)?

shakenfist/client/main.py
mandoonandy: I only put it in to avoid you wasting another sunday wondering why VM's don't start. :-)

PR #138, opened 2020-07-09
Standardise usage of PUT and POST API endpoints

mandoonandy:
Refs:
https://tools.ietf.org/html/rfc2616
https://www.restapitutorial.com/lessons/httpmethods.html


shakenfist/external_api/app.py
```diff
@@ -260,6 +260,22 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def metadata_putpost(meta_type, owner, key, value):
```
mikalstill, line 263: This one should have a leading underscore too?

PR #139, opened 2020-07-09
Do not write to etcd directly, go via db.py.

### LINKED ISSUE #124: Namespaces write directly to etcd from the API code
Everything else uses the db.py abstraction, which I think has advantages in terms of keeping the format of what is in etcd consistent. We should decide if we value db.py and if we do ensure that all etcd access flows through it.

mikalstill:
Fixes #124.


mikalstill:
Good catch, that was a rebase error. I will fix.

PR #167, opened 2020-07-16
Various API fixes

mandoonandy:


shakenfist/client/main.py
```diff
@@ -704,7 +705,12 @@ def instance_create(ctx, name=None, cpus=None, memory=None, network=None, networ
 
     diskdefs = []
     for d in disk:
-        size, base = _parse_spec(d)
+        p = _parse_spec(d)
+        if not p[1]:
+            print('Disk short-hand specification is <size>@<base>')
+            return
```
mikalstill, line 711: I don't think this is true. We should allow short hand blank disks. For example, ephemeral data disks. This is a valid start command:

```
sf-client instance create foo 1 1024 -d 8@cirros -d 42 -n ...
```

shakenfist/client/main.py
```diff
@@ -730,8 +739,11 @@ def instance_create(ctx, name=None, cpus=None, memory=None, network=None, networ
     for n in networkspec:
         defn = {}
         for elem in n.split(','):
-            key, value = elem.split('=')
-            defn[key] = value
+            s = elem.split('=')
+            if len(s) != 2:
+                print("Error in network specification. Missing equal sign: %s" % elem)
```
mikalstill, line 744: Should this error message be consistent with the one above?

shakenfist/daemons/main.py
```diff
@@ -53,7 +53,7 @@ def restore_instances():
                 LOG.info('%s Restoring instance' % i)
                 i.create()
             except Exception as e:
-                LOG.error('%s Failed to restore instance: %s' % (i, e))
+                LOG.error('Failed to restore instance: %s' % (e))
```
mikalstill, line 56: So, we should be consistent about using instance at the front as a string, that's because instance has a __str__() method which returns "instance(uuid)" in a nice form. The pattern you're removing here is in the code base in many places.

mikalstill: That said, maybe what we actually need is a logging helper to assist us with outputting high quality consistent messages so admins don't hate us?

shakenfist/external_api/app.py
```diff
@@ -608,6 +608,26 @@ def post(self, name=None, cpus=None, memory=None, network=None,
...
         name = re.sub(r'([^a-zA-Z0-9_\-])', '', name)
 
+        # Sanity check
+        if not disk:
+            return error(400, "Instance must specify at least one disk")
+        for d in disk:
+            if not isinstance(d, dict):
+                return error(400, "Disk specification should contain JSON objects")
+            if not all (k in d for k in ('size', 'base', 'bus', 'type')):
+                return error(400, "Disk specification is missing parameters: %s" % d)
```
mikalstill, line 618: Some of these are optional though...

* size if not set defaults to the size of the base image
* base is only needed for the boot disk
* bus defaults to the configuration item if not set
* type defaults to disk (the other option is CDROM)

shakenfist/external_api/app.py
```diff
@@ -608,6 +608,26 @@ def post(self, name=None, cpus=None, memory=None, network=None,
...
+
+        if network:
+            for n in network:
+                if not isinstance(n, dict):
+                    return error(400,
+                        "Network specification should contain JSON objects")
+
+                net_params = ('network_uuid', 'address', 'macaddress', 'model')
+                if not all (k in n for k in net_params):
```
mikalstill, line 627: Similarly here, some of these are optional...

* network_uuid: definitely required
* address: if not set we pick one for you
* macaddress: if not set we randomly generate one
* model: if not set you get the default from config

shakenfist/client/main.py
mandoonandy: True. Good save.

shakenfist/daemons/main.py
mandoonandy: So the variable "i" cannot be reported since the exception can occur before initialisation.
The bug is actually that "instance" (not "i") should be in the log statement.
Yes?

shakenfist/external_api/app.py
mandoonandy: Got carried away here obviously. Reversing.
I got so many 500 errors I started looking for culprits.

shakenfist/daemons/main.py
mikalstill: Oh I see. Fair.

mandoonandy:
Let me know if I missed any changes.

Issue #156, opened 2020-07-15
We should prefer not to start instances on the network node


mikalstill:
...until the other nodes are capacity constrained. Its already doing extra work so we should perhaps leave it as our last choice?

PR #169, opened 2020-07-16
Prefer scheduling instances on non-network nodes

mandoonandy:
Is this what you were thinking in issue #156?

### DIFF
--- a/shakenfist/db.py
+++ b/shakenfist/db.py
@@ -50,6 +50,13 @@ def get_nodes():
     return etcd.get_all('node', None)
 
 
+def get_network_node():
+    see_this_node()
+    for n in get_nodes():
+        if n['ip'] == config.parsed.get('NETWORK_NODE_IP'):
+            return n
+
+
 def get_network(network_uuid):
     see_this_node()
     return etcd.get('network', None, network_uuid)
--- a/shakenfist/scheduler.py
+++ b/shakenfist/scheduler.py
@@ -209,6 +209,15 @@ class Scheduler(object):
             db.add_event('instance', instance.db_entry['uuid'],
                          'schedule', 'Have most matching images', None, str(candidates))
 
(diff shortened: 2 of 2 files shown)


shakenfist/scheduler.py
```diff
@@ -209,6 +209,15 @@ def place_instance(self, instance, network, candidates=None):
...
                          'schedule', 'Have most matching images', None, str(candidates))
 
+            # Avoid allocating to network node if possible
+            net_node = db.get_network_node()
+            if len(candidates) > 1 and net_node['fqdn'] in candidates:
+                candidates.remove(net_node['fqdn'])
+                LOG.info('Scheduling %s, %s are non-network nodes'
+                            % (instance, candidates))
+                db.add_event('instance', instance.db_entry['uuid'],
+                             'schedule', ' are non-network nodes', None, str(candidates))
```
mikalstill, line 219: Sorry to be a pain, but I think that should be "Are non-network nodes" to be consistent with the others here?

shakenfist/db.py
```diff
@@ -50,6 +50,13 @@ def get_nodes():
     return etcd.get_all('node', None)
 
 
+def get_network_node():
+    see_this_node()
+    for n in get_nodes():
+        if n['ip'] == config.parsed.get('NETWORK_NODE_IP'):
+            return n
+
+
```
mikalstill, line 59: This is fine for here, but it leaves me feeling like the current way of specifying if a node is a network node or not is clunky. I should think more about that.

shakenfist/scheduler.py
mandoonandy: Absolutely.

mikalstill:
Oh sorry, this was on the wrong branch. We should propose these against master and then cherry pick them to v0.2-devel. Fixing that now.

PR #170, opened 2020-07-16
Implement power states, and expose power and instance state in CLI.

### LINKED ISSUE #159: The powerstates of instances are not exposed
Currently the only method of determining the powerstate of an instance is to look at the event log (and even then, that's not entirely accurate for a real-time response). From a usability perspective, this seems to be incompatible with allowing instances to be powered off.

mikalstill:
Fixes #159, and should be cherry picked to v0.2-devel once merged.


mandoonandy:
Cherry picked to v0.2-devel.
Pls check I got it right...

PR #185, opened 2020-07-18
Chase memory issues with scheduling under load

mikalstill:


mikalstill:
Still needs: unit tests for reworked instance starts; unit tests for reworked scheduler. Probably other stuff I've forgotten too.

mikalstill:
Oh, and code to tune KSM if its enabled. I think maybe we should just own the state of KSM on our hypervisor nodes instead of requiring deployers to handle it.

PR #189, opened 2020-07-20
Reject instance create requests for invalid names

mandoonandy:
Resolves shakenfist/terraform-provider-shakenfist#12

### DIFF
--- a/shakenfist/external_api/app.py
+++ b/shakenfist/external_api/app.py
@@ -613,8 +613,9 @@ class Instances(Resource):
              instance_uuid=None):
         global SCHEDULER
 
-        # We need to sanitise the name so its safe for DNS
-        name = re.sub(r'([^a-zA-Z0-9_\-])', '', name)
+        # Check that the instance name is safe for use as a DNS host name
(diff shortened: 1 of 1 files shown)


mandoonandy:
I meant to add that this is my *proposed* solution to TF issue 12.

mikalstill:
I feel a bit like we should solve this in SF as described in my comment on the issue. We should talk about this before merging.

shakenfist/external_api/app.py
```diff
@@ -607,8 +607,9 @@ def post(self, name=None, cpus=None, memory=None, network=None,
              instance_uuid=None):
         global SCHEDULER
 
-        # We need to sanitise the name so its safe for DNS
-        name = re.sub(r'([^a-zA-Z0-9_\-])', '', name)
+        # Check that the instance name is safe for use as a DNS host name
+        if name != re.sub(r'([^a-zA-Z0-9_\-])', '', name) or len(name) > 63:
+            return error(400, "Instance name must be useable as a DNS host name")
```
mikalstill, line 612: Quotes! I see quotes!

mandoonandy: Rebased to current master and fixed the quotes

PR #199, opened 2020-07-22
Update power state after operations which change power state.

### LINKED ISSUE #197: Power status updates should be faster for new instances
Either we should increase the frequency of power status updates in the cleaner, or add a power status check to the end of the instance boot cycle.

mikalstill:
This should provide more immediate feedback to callers. Fixes #197.

### DIFF
--- a/shakenfist/util.py
+++ b/shakenfist/util.py
@@ -121,15 +121,17 @@ def extract_power_state(libvirt, domain):
     state, _ = domain.state()
     if state == libvirt.VIR_DOMAIN_SHUTOFF:
         return 'off'
-    elif state == libvirt.VIR_DOMAIN_CRASHED:
+
+    if state == libvirt.VIR_DOMAIN_CRASHED:
         return 'crashed'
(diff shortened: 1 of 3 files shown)


shakenfist/util.py
```diff
@@ -117,6 +117,21 @@ def get_libvirt():
...
+def extract_power_state(libvirt, domain):
+    state, _ = domain.state()
+    if state == libvirt.VIR_DOMAIN_SHUTOFF:
+        return 'off'
+    elif state == libvirt.VIR_DOMAIN_CRASHED:
+        return 'crashed'
+    elif state in [libvirt.VIR_DOMAIN_PAUSED,
+                   libvirt.VIR_DOMAIN_PMSUSPENDED]:
+        return 'paused'
+    else:
```
mandoonandy, line 129: This else is not really necessary now it's a function. (Nor the elif.)

mikalstill: Nah, I'll fix it.

mikalstill:
Please to be re-reviewing?

PR #221, opened 2020-07-24
Catch libvirt unsupported config errors in API

mandoonandy:
Now prints:
```
root@sf-1:/srv/shakenfist# sf-client instance create rubbishvideo 1 1024 -d8 -V model=rubbish,memory=888 -n a6b3669e-a771-43d5-9061-4d1aac7dcb74 
ERROR: Malformed Request: unsupported configuration: unknown video model 'rubbish'
```

There might be more to this? I think using error_with_cleanup() should be sufficient.

The next step could be to actually remove the instance from the database but could compromise other cleanup steps?

*Testing for this should probably be done through integration testing via the API. Once next-gen CI replaces ansible-ci?? Start another issue?*

### DIFF
--- a/shakenfist/external_api/app.py
+++ b/shakenfist/external_api/app.py
@@ -822,8 +822,17 @@ class Instances(Resource):
                     n.ensure_mesh()
                     n.update_dhcp()
 
-            with util.RecordedOperation('instance creation', instance) as _:
-                instance.create(lock=lock)
+            libvirt = util.get_libvirt()
+            try:
+                with util.RecordedOperation('instance creation',
+                                            instance) as _:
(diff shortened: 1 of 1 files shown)


mikalstill:
I agree this is a test for CI, given we mock away libvirt in unit tests. Tracking the gap for now with a bug makes sense to me. I also agree we need better error roll backs (delete interfaces etc), but I also suspect we should have the cleaner reap orphan interfaces to provide "defence in depth", although it would mask errors elsewhere.

mikalstill:
I filed https://github.com/shakenfist/deploy/issues/38 to track the CI bit.

PR #251, opened 2020-08-01
Continued adventures in broken mesh networking.

### LINKED ISSUE #250: Ubuntu 18.04's cloud init logs errors
```
[    7.105336] cloud-init[546]: 2020-08-01 06:25:29,547 - openstack.py[WARNING]: Unknown network_data link type (physical); treating as physical
```

mikalstill:
We needed to change the interface type. Fixes #250.

### DIFF
--- a/shakenfist/client/main.py
+++ b/shakenfist/client/main.py
@@ -597,7 +597,7 @@ def instance_list(ctx, all=False):
 def _pretty_data(row, space_rules):
     ret = ''
     for key in space_rules:
-        ret += key + '=' + str(row[key]).ljust(space_rules[key]) + '  '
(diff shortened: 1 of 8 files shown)


shakenfist/client/main.py
```diff
@@ -597,7 +597,7 @@ def instance_list(ctx, all=False):
 def _pretty_data(row, space_rules):
     ret = ''
     for key in space_rules:
-        ret += key + '=' + str(row[key]).ljust(space_rules[key]) + '  '
+        ret += key + '=' + str(row.get(key, '')).ljust(space_rules[key]) + '  '
```
mandoonandy, line 600: The get() was to enable use of the function when the data row didn't contain the entire set of keys in space_rules.
It turns out we don't need that at the moment.
So, yes. I guess that was cruft...

PR #279, opened 2020-08-06
We create networks too frequently as part of mesh maintenance.

mikalstill:


mikalstill:
Andy's patch is better.

PR #280, opened 2020-08-06
Stop daemon:net recreating networks unnecessarily

mandoonandy:
Net daemon runs every 30 seconds on every node.
Each run the daemon recreates every network on it's respective node then sends the network node a request to recreate network.
Network node receives a request to recreate every network every 30 seconds from every node (including itself).

To avoid the storm, the daemon should do some checks of the network before recreation.

If the "network ok" checks are not sufficient then we add more checks.
This will let network creation bugs surface rather hiding them in the log storm.

### DIFF
--- a/shakenfist/daemons/net.py
+++ b/shakenfist/daemons/net.py
@@ -26,7 +26,7 @@ class monitor(object):
         host_networks = []
         seen_vxids = []
 
-        if not net.is_network_node():
+        if not util.is_network_node():
             # For normal nodes, just the ones we have instances for
             for inst in list(db.get_instances(only_node=config.parsed.get('NODE_NAME'))):
                 for iface in db.get_instance_interfaces(inst['uuid']):
@@ -104,7 +104,7 @@ class monitor(object):
             time.sleep(30)
 
             try:
-                LOG.info("Daeman:net: Checking networks")
+                LOG.info("Daemon:net: Checking networks")
                 self._maintain_networks()
             except Exception as e:
                 util.ignore_exception('network monitor', e)
--- a/shakenfist/daemons/net.py
+++ b/shakenfist/daemons/net.py
@@ -26,7 +26,7 @@ class monitor(object):
         host_networks = []
         seen_vxids = []
 
-        if config.parsed.get('NODE_IP') != config.parsed.get('NETWORK_NODE_IP'):
+        if not net.is_network_node():
             # For normal nodes, just the ones we have instances for
             for inst in list(db.get_instances(only_node=config.parsed.get('NODE_NAME'))):
                 for iface in db.get_instance_interfaces(inst['uuid']):
@@ -46,7 +46,10 @@ class monitor(object):
                         LOG.info('Hard deleted stray network interface %s '
                                  'associated with absent instance %s'
                                  % (ni['uuid'], ni['instance_uuid']))
-                    elif inst.get('state', 'unknown') in ['deleted', 'error', 'unknown']:
+
+                    elif inst.get('state', 'unknown') in ['deleted',
+                                                          'error',
+                                                          'unknown']:
                         db.hard_delete_network_interface(ni['uuid'])
                         LOG.info('Hard deleted stray network interface %s '
                                  'associated with %s instance %s'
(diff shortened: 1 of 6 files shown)


mandoonandy:
So maybe my solution is over-engineered compared to @mikalstill PR?

shakenfist/net.py
```diff
@@ -40,6 +40,11 @@ def from_db(uuid):
                    namespace=dbnet['namespace'])
 
 
+def is_network_node():
```
mikalstill, line 43: I wonder if this should be in shakenfist.util?

shakenfist/daemons/net.py
```diff
@@ -60,14 +63,17 @@ def _maintain_networks(self):
                 if not n:
                     continue
 
-                n.create()
+                if not n.is_okay():
+                    LOG.info("Daemon:net: maintain_networks() recreating %s",
```
mikalstill, line 67: This "Daemon:net" bit is a little odd to me. We should be getting the filename from python logging already. I think I'd prefer something matching the style of the other log lines, and we should rethink logging in general in v0.3.

shakenfist/net.py
```diff
@@ -99,7 +104,33 @@ def subst_dict(self):
...
+        if is_network_node():
+            okay &= self.is_dnsmasq_running()
+
+        return okay
+
+    def is_created(self):
+        """Attempt to ensure network has been created successfully"""
+
+        subst = self.subst_dict()
+        out, _ = processutils.execute('ip link show %(vx_bridge)s' % subst,
```
mikalstill, line 122: Why not use util.check_for_interface here?

shakenfist/net.py
```diff
@@ -99,7 +104,33 @@ def subst_dict(self):
...
+                                      shell=True)
+        net_created = re.match(r'.*[<,]UP[,>].*', out)
+        if not net_created:
+            LOG.warning('network(%s) is_created() bridge is not up: %s',
```
mikalstill, line 126: network has a __str__ which should give a nicely formatted description of the network (uuid and vxid). I think that's what we use for the rest of this file? I do think logging needs a rethink though.

mandoonandy:
Should we add `util.check_for_interface('vxlan-%s' % n.vxlan_id)` to the `is_okay()`?

shakenfist/net.py
```diff
@@ -276,11 +309,28 @@ def delete(self):
                         ipm.release(self.floating_gateway)
                         db.persist_ipmanager('floating', ipm.save())
 
+    def is_dnsmasq_running(self):
+        """Determine if dnsmasq process is running for this network"""
```
mikalstill, line 313: We should only do this if DHCP is enabled for the network.

shakenfist/net.py
```diff
@@ -276,11 +309,28 @@ def delete(self):
...
+        try:
+            out, _ = processutils.execute('pgrep -a dnsmasq', shell=True)
```
mikalstill, line 315: So, there should be lots of dnsmasq processes (one per network, they're in a network namespace). So this wont work? That said, we also have the pid of the dnsmasq process at /srv/shakenfist/dhcp/...net...uuid.../pid so we could do a more direct "is pid running" check with psutils?

shakenfist/net.py
mandoonandy: Sure. Will do.

mandoonandy: I didn't replace all the other uses to have a minimal touch for patching CT.
Can use it more widely in a later PR.

shakenfist/net.py
mandoonandy: I want to check that it exists and is up. Since the "up" is a separate step.

shakenfist/net.py
mandoonandy: true!

shakenfist/net.py
mandoonandy: The process is affecting the network namespace but the linux process list is unaffected. 

"Works on my cluster" (TM)

shakenfist/net.py
mandoonandy: Sorted.

shakenfist/net.py
mikalstill: When we are big boys we should add the up as an option on that util method. Not blocking you for now though.

shakenfist/net.py
mikalstill: We've just talked about this and I think we are aligned now.

PR #291, opened 2020-08-07
Fix sf-upgrade fail while deleting mesh events

### LINKED ISSUE #290: Deploy fails on sf-upgrade
etcdctl still fails due to gRPC size limitation.

mandoonandy:
Resolves #290 

Is this deletion code still a thing that's required? Anyway this should avoid failures during deploy.

### DIFF
--- a/shakenfist/client/upgrade.py
+++ b/shakenfist/client/upgrade.py
@@ -1,16 +1,49 @@
 # Copyright 2020 Michael Still
 
 import etcd3
-import json
-import string
+
+from oslo_concurrency import processutils
 
 from shakenfist import db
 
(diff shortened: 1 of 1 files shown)


mandoonandy:
Sadly it blocked my deploy test.

mandoonandy:
How about I change it to delete all currently targeted events and then delete the remain events keeping the last 50 events per network?

mikalstill:
I like that idea.

mandoonandy:
Since sf-syd is probably the only cloud and it does not have ensure mesh events any more, changed the spec to just last 50 events.
Also, we may not even need to this at all, I just made a quick code change.

PR #297, opened 2020-08-09
Git work flow doc

mandoonandy:
# Git Work Flow
After many pleasant afternoons underneath a tree watching the world pass - and reading many web pages - it appears that our current method is *close to the best* for right *now.*

Our major problem might be that we are doing it slightly wrong. Development has been "bug driven". This has lead to lots of hot-fixes and desired changes being cherry-picked to an existing release.

We are now be moving to Feature Improvement and New Feature Implementation. This will reduce the flow of 'must be in the current release' bug fixes.

If there is a serious desire to have a number of features or fixes in a release then make a new release.

### DIFF
--- a/docs/git_usage.md
+++ b/docs/git_usage.md
@@ -1,0 +1,52 @@
+# Development Work Flow
+
+### Short Lesson
+The majority of teams using git have a work flow that looks similar to the four well known work flows:
+* [Git Flow](https://datasift.github.io/gitflow/IntroducingGitFlow.html)
+* [GitHub Flow](https://guides.github.com/introduction/flow/)
+* [GitLab Flow](https://about.gitlab.com/blog/2014/09/29/gitlab-flow/)
+* [Trunk Based Development](https://trunkbaseddevelopment.com/)
+
+## Git Development - the Shaken Fist Way
+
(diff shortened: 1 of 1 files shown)


mandoonandy:
Answering in order.

Master should always be deployable so multiple features can be developed on the same feature branch - preferably the commits that are not squashed should be complete. I quite like the idea of longer running feature branches if they are reviewed during development. (Not so long that it's a merge hell.)

Leave the devel branch until it's historically only a curiosity. Or if keen, delete each tag and re-tag in `master`.

We have the v1.0.0 when the product is trusted with production workloads. Then the usual, incrementing the major version number on breaking changes. (I think a breaking change includes 'not recommended to upgrade' situations. This could occur with an OS change.) Even a release every fortnight will not increment the minor version beyond 2 digits for years.

I agree, that a minor release could just be lots of hot fixes. This avoids cherry-picking hot-fixes for a patch release. Patch releases are reserved for serious bug fixes.

mikalstill:
This all sounds reasonable to me, except for moving tags. I think we just leave them where they are now.

mandoonandy:
Yep. Leave the history as is.

PR #302, opened 2020-08-23
Use a simple etcd based queue for network node tasks.

mikalstill:
Instead of making real time REST calls to a special API, instead
queue jobs and have a daemon on the network node process them as
fast as it can. This is the pattern I want to move to for instance
starts as well.

### DIFF
--- a/shakenfist/daemons/net.py
+++ b/shakenfist/daemons/net.py
@@ -102,12 +102,13 @@ class Monitor(daemon.Daemon):
             return
 
         if 'network_uuid' not in workitem:
+            LOG.warn('Network workitem %s lacks network uuid.' % workitem)
             return
 
         n = net.from_db(workitem.get('network_uuid'))
         if not n:
             LOG.warn(
-                'Received work item for non-existant network: %s' % workitem)
+                'Received workitem for non-existant network: %s' % workitem)
             return
 
         # NOTE(mikal): there's really nothing stopping us from processing a bunch
--- a/shakenfist/daemons/net.py
+++ b/shakenfist/daemons/net.py
@@ -1,3 +1,4 @@
+import etcd3
 import logging
 import time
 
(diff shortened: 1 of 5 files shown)


shakenfist/daemons/net.py
```diff
@@ -94,12 +95,57 @@ def _maintain_networks(self):
         db.persist_node_vxid_mapping(
             config.parsed.get('NODE_NAME'), vxid_to_mac)
 
+    def _process_network_node_workitems(self):
+        workitem = db.dequeue('networknode')
+        if not workitem:
+            time.sleep(0.2)
```
mandoonandy, line 101: Could the SF etcd library use watch_prefix() to avoid this busy wait?
I acknowledge this would mean a separate thread for _maintain_networks() but better than a etcd request every 0.2 secs?

shakenfist/daemons/net.py
```diff
@@ -94,12 +95,57 @@ def _maintain_networks(self):
...
+            return
+
+        if 'network_uuid' not in workitem:
+            return
```
mandoonandy, line 105: Probably should warn here since this would be a code bug

shakenfist/external_api/app.py
```diff
@@ -1275,18 +1275,11 @@ def post(self, netblock=None, provide_dhcp=None, provide_nat=None, name=None,
                 n.create()
```
mandoonandy, line 1275: To avoid code duplication what do you think about the network node queuing the network change?
Also removes the unnecessary check of network uuid validity.
You will also notice that db.enqueue() snuck into the lock context when it doesn't need the lock.

mandoonandy (changes requested):

mandoonandy:
I like this change, it makes inter process communication much cleaner.

shakenfist/daemons/net.py
mikalstill: I think this is a good idea. I am hesitant to do it now though because the code feels quite instable to me... Later in this series (which you haven't seen yet) I discover that we are using different names for locks and are clashing, and move instance create / delete to queues too. I think I'd like to get to somewhere a little more stable before optimising this code -- especially because we might abandon this approach entirely when we discover we can't make it stable or something. Is that ok to defer for now with a bug or something?

shakenfist/daemons/net.py
mikalstill: Done.

shakenfist/external_api/app.py
mikalstill: That's a good idea. Done.

mikalstill:
Just running a test run on this before uploading a tweaked version.

mikalstill:
Sorry, I should have said the tests passed, but I went to a meeting instead.

shakenfist/daemons/net.py
mandoonandy: sounds fair to me: "bug or something"

mandoonandy (approved):
LGTM

PR #306, opened 2020-09-03
Continued queue maturity work

mikalstill:


mandoonandy (approved):
LGTM

mandoonandy (approved):
Looks fine. With the caveat, that without explanation of what this should do I can't compare the code to the intent.

mikalstill:
Well, the end goal is "works as well as it did before queues, but hopefully better" if that makes sense.

PR #307, opened 2020-09-05
Run python 3.8 unit tests as well as 3.7

### LINKED ISSUE #294: API app.py breaking change to gunicorn3
Commit f93078d changed `gunicorn3` to `gunicorn` to support Ubuntu 20.04 but breaks 18.04 support. We could write a bunch of stuff to change the code depending on version, but 20.04 is a LTS version anyway.

I propose that we ONLY support 20.04 for v0.3 (this is actually the current situation). Once we hit v1.0 we can maintain previous LTS  versions.

Discuss.

mikalstill:
Fixes #294 remaining issues.

### DIFF
--- a/.github/workflows/python-unit-tests.yml
+++ b/.github/workflows/python-unit-tests.yml
@@ -14,7 +14,25 @@ on:
       - v*-devel
 
 jobs:
-  build:
+  lint:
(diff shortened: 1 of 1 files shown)


mandoonandy (approved):
cool

PR #308, opened 2020-09-07
Bug fixes for issues found in testing queues

mikalstill:
Make sure we don't infinitely attempt to place instances. Fix a queue reporting bug. Don't infinitely attempt to delete instances.

### DIFF
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -86,8 +86,9 @@ def image_fetch(url, instance_uuid):
         instance = None
         if instance_uuid:
             instance = virt.from_db(instance_uuid)
-        images.get_image(url, [], instance,
-                         timeout=images.IMAGE_FETCH_LOCK_TIMEOUT)
+
+        img = images.Image(url)
+        img.get([], instance)
     except exceptions.LockException:
         pass
 
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -1,5 +1,4 @@
 import copy
-import logging
 import multiprocessing
 import os
 import setproctitle
@@ -10,16 +9,24 @@ from shakenfist.daemons import daemon
 from shakenfist import db
 from shakenfist import exceptions
 from shakenfist import images
+from shakenfist import logutil
 from shakenfist import net
 from shakenfist import scheduler
 from shakenfist import util
 from shakenfist import virt
 
-LOG = logging.getLogger(__name__)
+
+class JobName(object):
+    def __init__(self, name):
+        self.name = name
+
+    def get_describing_tuple(self):
+        return 'workitem', self.name
 
 
 def handle(jobname, workitem):
-    LOG.info('Worker for workitem %s has pid %d' % (jobname, os.getpid()))
+    j = JobName(jobname)
+    logutil.info([j], 'Processing workitem')
     setproctitle.setproctitle(
         '%s-%s' % (daemon.process_name('queues'), jobname))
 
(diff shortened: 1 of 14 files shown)


shakenfist/db.py
```diff
@@ -238,6 +239,7 @@ def place_instance(instance_uuid, node):
         return
```
mandoonandy, line 239: Does this check block the recording of a placement attempt? If not, maybe it shouldn't be called?

mandoonandy (commented):
I'm unclear (without more investigation) how the instance start queuing works.

_preflight and _start are queued at the same time, but _preflight can be requeued? Does that put the _preflight after the _start?

shakenfist/db.py
mikalstill: So, a new attempt shouldn't be allowed to use the current node, so it shouldn't? That said, there might be a bug in the caller then?

shakenfist/daemons/queues.py
```diff
@@ -103,6 +108,10 @@ def instance_preflight(instance_uuid, network):
                      'schedule', 'retry', None,
                      'insufficient resources: ' + str(e))
 
+    if instance.db_entry.get('placement_attempts') > 3:
+        raise exceptions.AbortInstanceStartException(
+            'Too many start attempts.')
+
     try:
         if instance.db_entry.get('requested_placement'):
             candidates = [instance.db_entry.get('requested_placement')]
```
mikalstill, line 117: Yes, there is a bug in this logic. The list of candidate should exclude the current node. I will fix.

mikalstill:
Sorry, I missed your greater comment earlier. The preflight is a re-check that the node is still capable of scheduling the instance. If its not, the instance is re-scheduled on another node and the entire workitem should be re-queued there. Does that make sense?

shakenfist/util.py
```diff
@@ -260,3 +261,22 @@ def ignore_exception(processname, e):
         msg += '\n%s' % traceback.format_exc()
 
     LOG.error(msg)
+
+
+METHOD_RE = re.compile('.* in (.*)')
+
+
+def log(level, relatedobjects, message):
```
mandoonandy, line 269: I think you should have a command for each log level.

mandoonandy (commented):
My general comment is that if we go down the path of high information logging, then we should do it properly with parameterised fields. Then it can be ingested easily into any of the log processors.
It will also be less verbose then using various ( [ : to indicate a field.

shakenfist/util.py
mikalstill: Mutter.

mikalstill:
Tweaks based on our slack conversation.

mikalstill:
Logs look like:

```
Sep 10 10:56:18 cbr-sf-3 INFO sf-queues[92312] Processing workitem; method=daemons/queues.py:29:handle(); workitem=1599735378.7058043-000
Sep 10 10:56:18 cbr-sf-3 INFO sf-queues-1599735378.7058043-000[92312] Executing task instance_preflight; instance=ebbc70c7-ce6e-4eb2-8371-4d9313011d23; method=daemons/queues.py:51:handle(); workitem=1599735378.7058043-000
Sep 10 10:56:18 cbr-sf-3 INFO sf-queues-1599735378.7058043-000[92312] Executing task image_fetch; instance=ebbc70c7-ce6e-4eb2-8371-4d9313011d23; method=daemons/queues.py:51:handle(); workitem=1599735378.7058043-000
Sep 10 10:56:18 cbr-sf-2 INFO sf-queues-1599735378.4955704-000[2050934] Executing task image_fetch; instance=64d1f8d9-4b3b-486b-8422-e25c9af28147; method=daemons/queues.py:51:handle(); workitem=1599735378.4955704-000
```

mandoonandy (approved):

PR #313, opened 2020-09-08
Schedule instance on node resulting from placement

### LINKED ISSUE #312: Instances incorrectly start on node that received API call

mandoonandy:
Resolves #312


mikalstill (approved):
Good catch. Thanks.

PR #315, opened 2020-09-08
Avoid instance allocation to network node

mandoonandy:
Networks are now fully created on the network node. On instance allocation the scheduler always chooses the network node since it has the network.

As discussed previously, the scheduler will probably be changed in the future. Only adding tests for this change.


mikalstill (approved):
LGTM

PR #322, opened 2020-09-18
Refresh locks during long running external processes

mikalstill:


mandoonandy (commented):
LGTM once you convince the linter

My favourite part was of the movie was  is_network_node()  :-)

PR #325, opened 2020-09-26
Allow numeric configuration values to be set.

### LINKED ISSUE #324: You cannot set numeric config flags
They are always pulled from the environment variable as a string.

mikalstill:
Fixes #324.

### DIFF
--- a/shakenfist/config.py
+++ b/shakenfist/config.py
@@ -4,16 +4,14 @@ import copy
 import os
 import socket
 
+from shakenfist import exceptions
 from shakenfist import util
 
 
-node_name = socket.getfqdn()
-try:
-    node_ip = socket.gethostbyname(node_name)
-except Exception as e:
-    # Only for localhost development environments
-    node_ip = '127.0.0.1'
-    util.ignore_exception('config parser', e)
+# NOTE(mikal): Dear future developer. Thanks ftheor dropping by! Remember that the type
+# of values in this set of defaults matters, as we use it to decide how to parse
+# override values from the user. Your current options are strings (the default), integers,
+# and floats. Integers and floats differ in that a float has values after a decimal point.
 
 CONFIG_DEFAULTS = {
     # Deployment wide options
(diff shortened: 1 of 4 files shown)


shakenfist/config.py
```diff
@@ -111,7 +111,14 @@ def __init__(self):
         for var in os.environ:
             if var.startswith('SHAKENFIST_'):
                 flag = var.replace('SHAKENFIST_', '')
-                self.config[flag] = os.environ[var]
+                value = os.environ[var]
+
+                try:
+                    value = float(value)
```
mandoonandy, line 117: How will this handle non floating values? eg. API_PORT

mandoonandy (changes requested):

shakenfist/config.py
mikalstill: They'll get a valueerror, and fall through to using the original value. For example:

```
$ python3
Python 3.8.2 (default, Jul 16 2020, 14:00:26) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> foo="1"
>>> try:
...   foo=float(foo)
... except ValueError:
...   pass
... 
>>> print(foo)
1.0
>>> foo="gerkin"
>>> try:
...   foo=float(foo)
... except ValueError:
...   pass
... 
>>> print(foo)
gerkin
```

mandoonandy: I understand the processing for text/numerics.

I should have been a little clearer. Integer values such API_PORT are now floating-point values - as you actually demonstrated above. This breaks other parts of the code. Some parts still work because they are formatting with %d but others are using %s such as the gunicorn API_TIMEOUT.

In the future the value may be directly applied and an integer is expected.

Maybe test for an integer first?

mikalstill: Oh right. I see now. Its interesting. I wonder if we need a type system for configuration values? I guess I want to avoid that complexity if possible.

mikalstill: The problem being, that a string of "123" is a valid int and float. As is "123.123". Both casts will work and not raise a ValueError.

mandoonandy: If you keep the same logic, I was thinking of something simple like
```
try:
    foo = int(foo)
except:
    try:
        foo=float(foo)
    except ValueError:
        pass
```

mikalstill:
Please don't merge this yet. I want to add some unit tests first, but I'd be interested in your thoughts on this approach.

mandoonandy:
That looks quite elegant. Maybe add a nice error message if it triggers an exception. (Also, maybe a note in `config.py` that the correct type is important ie. Use a float with trailing ".0" if the value should be a floating point value.)

mikalstill:
Will do if I ever escape meetings.

mandoonandy (approved):
LGTM

PR #329, opened 2020-09-28
Add tests for config.py

mikalstill:
I forgot the unit tests in the previous commit. Sorry.


mandoonandy (approved):
We love tests - early or late does not matter

PR #335, opened 2020-09-29
Automate website publication

mikalstill:


.github/workflows/publish-website.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - master
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
```
jackadamson, line 13: Do we need submodules? But otherwise looks good

mikalstill: We do not. I shall remove that.

jackadamson (approved):
Looks good to me

mandoonandy (approved):
Looks believable.

PR #338, opened 2020-09-30
Clarify releases for developer installs.

mikalstill:


mandoonandy (approved):
LGTM

I use `RELEASE=git:` to deploy my current repo state. I don't need to specify the branch (so I can reuse a single deploy script) and I also get any uncommitted changes. 

Is that worth mentioning or just adding too much confusion.

PR #343, opened 2020-10-03
Always print exceptions even when uuid is unknown

### LINKED ISSUE #342: Queues drops exceptions
Queues drops exceptions if it doesn't know the instance_uuid.

This should not be significant except that I'm stupid and the missing exceptions had me questioning my ability to computer.

This is a memorial to my lost time.

mandoonandy:
I should have worked this out with less effort...

Resolves #342


mikalstill (approved):
LGTM

PR #344, opened 2020-10-04
Queuing tasks objects

mandoonandy:
I don't trust dict's for passing objects around. Classes verify the data on creation and can be relied upon to contain the correct data. Task creation can cause bugs if the wrong keys are used. 

This removes reliance on correct keys used in task dict's. Check `external/app.py` for the neat code improvement.

Plan:
1. Get consensus on this change
2. Add tests for this new code
3. Further improvements taking advantage of these objects eg. inter alia `next_state` removal

This PR branches from #343.

### DIFF
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -76,8 +76,8 @@ def handle(jobname, workitem):
             log_i.info("Task complete")
 
     except Exception as e:
+        util.ignore_exception(daemon.process_name('queues'), e)
         if instance_uuid:
-            util.ignore_exception(daemon.process_name('queues'), e)
             db.enqueue_instance_delete(config.parsed.get('NODE_NAME'),
                                        instance_uuid, 'error',
                                        'failed queue task: %s' % e)
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -13,6 +13,12 @@ from shakenfist import net
 from shakenfist import scheduler
 from shakenfist import util
 from shakenfist import virt
+from shakenfist.tasks import (DeleteInstanceTask,
+                              FetchImageTask,
+                              InstanceTask,
+                              PreflightInstanceTask,
+                              StartInstanceTask,
+                              )
 
 
 LOG, _ = logutil.setup(__name__)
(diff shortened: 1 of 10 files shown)


shakenfist/daemons/queues.py
```diff
@@ -29,55 +35,68 @@ def handle(jobname, workitem):
     task = None
     try:
         for task in workitem.get('tasks', []):
-            instance_uuid = task.get('instance_uuid')
-            log_i = log.withInstance(instance_uuid)
-            log_i.info("Starting task")
+            if InstanceTask.__subclasscheck__(type(task)):
+                instance_uuid = task.instance_uuid()
+                log_i = log.withInstance(instance_uuid)
```
mikalstill, line 40: I think I'd rather just replace log here instead of creating log_i.

shakenfist/daemons/queues.py
```diff
@@ -29,55 +35,68 @@ def handle(jobname, workitem):
...
+            else:
+                instance_uuid = None
+                log_i = log
```
mikalstill, line 43: And then this would go away.

shakenfist/daemons/queues.py
```diff
@@ -29,55 +35,68 @@ def handle(jobname, workitem):
...
                         db.update_instance_error_message(
-                            instance_uuid, task.get('next_state_message'))
+                            instance_uuid, task.next_state_message())
                     db.enqueue('%s-metrics' %
                                config.parsed.get('NODE_NAME'), {})
                 except Exception as e:
                     util.ignore_exception(daemon.process_name('queues'), e)
 
+            else:
+                log_i.withField('task', task).error("Unhandled task - dropped")
```
mikalstill, line 93: golang quotes! My eyes bleed!

mikalstill (commented):
I like this approach. Proceed!

shakenfist/daemons/queues.py
mandoonandy: `log` cannot be replaced because there is no method to remove a log field. A task may not have an `instance_uuid`.

mikalstill: But everyone below uses log_i right? So wouldn't we just be replacing the value of log in this scope?

mandoonandy: The `withInstance()` adds the `instance_uuid` field. So if - as you suggest - we store that logger with the added instance field in `log`. Next loop we call `withInstance()` again with the new instance_uuid and it will overwrite the field in the logger. All good. What happens when we loop a third time and do not have an instance_uuid (because the task doesn't have an instance_uuid). Overwrite the field with `None`? Then we have a incorrect log output implying that the field was supplied with None.

I'm wondering if you are assuming that instance_uuid is always the same in `workitem`? Not an incorrect assumption but if true we should maybe change the data structure?

Am I being totally obtuse?

[Sneak peak] In either case, it will be moot because I will be suggesting in a near-future PR that `workitem`s only contain single tasks and cause single state transitions.

mandoonandy:
Test cases added as promised.

Once the above issue is resolved, this is ready to merge. Promised step 3 is better in a separate PR.

shakenfist/daemons/queues.py
mikalstill: I think this is me being asleep at the wheel and forgetting this code is in a loop. I blame the github PR viewer thing not showing enough context.

mikalstill (approved):
LGTM, thanks.

PR #357, opened 2020-10-07
Re-enable CI in Jenkins

mikalstill:


shakenfist/daemons/queues.py
```diff
@@ -35,6 +35,10 @@ def handle(jobname, workitem):
     task = None
     try:
         for task in workitem.get('tasks', []):
+            if isinstance(task, dict):
+                raise exceptions.UnknownTaskException(
```
mandoonandy, line 39: Dict's and any other unknown type are handled at the end of the loop by https://github.com/shakenfist/shakenfist/blob/52ba17eeffbd7e61148debd71f13e8d46ef91dc4/shakenfist/daemons/queues.py#L92-L93

I didn't raise an exception since the rest of the system may run. (Future changes having more tasks than the current critical items will probably appreciate continuation.)

Should we change the Jenkins stage  `stage('Assert that the logs look sensible')` to also look for `ERROR`?

mandoonandy: Just having a second think about my above comment. My code doesn't catch the exact problem your code is catching.

The current code base does not place anything but QueueTasks on the queue. It is a major coding error if anything other than QueueTasks reaches handle() and therefore deserves the resulting fireworks.

For belts and braces you could keep it and modify to: `if not QueueTask.__subclasscheck__(type(task))`

mikalstill: So, I think you're saying this code should stay as is? Or am I misunderstanding you?

mikalstill: Oh, and I'll add a log check for ERROR now.

mandoonandy: If you change the `dict` check to `if not QueueTask.__subclasscheck__(type(task))`, we will be guaranteed to catch any problem.

mikalstill: Done.

mikalstill:
Ok, that's enough churn for one PR. I've disabled the failing CI again and would like to see this merged for now. Note that Jenkins might fail -- the grafana apt repository seems to be broken at the moment.

mikalstill:
Yeah, Jenkins failures are grafana apt repository being broken right now.

mandoonandy (approved):

Issue #359, opened 2020-10-07
Network has no ipblock


mikalstill:
Another issue from CI:

```
Oct  7 06:51:43 sfci-lxayws sh[21926]: Traceback (most recent call last):
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/bin/sf-daemon", line 8, in <module>
Oct  7 06:51:43 sfci-lxayws sh[21926]:     sys.exit(main())
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 163, in main
Oct  7 06:51:43 sfci-lxayws sh[21926]:     _audit_daemons()
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 156, in _audit_daemons
Oct  7 06:51:43 sfci-lxayws sh[21926]:     _start_daemon(d)
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 100, in _start_daemon
Oct  7 06:51:43 sfci-lxayws sh[21926]:     DAEMON_IMPLEMENTATIONS[d].Monitor(d).run()
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 141, in run
Oct  7 06:51:43 sfci-lxayws sh[21926]:     self._process_network_node_workitems()
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 116, in _process_network_node_workitems
Oct  7 06:51:43 sfci-lxayws sh[21926]:     n.create()
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/net.py", line 122, in create
Oct  7 06:51:43 sfci-lxayws sh[21926]:     subst = self.subst_dict()
Oct  7 06:51:43 sfci-lxayws sh[21926]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/net.py", line 88, in subst_dict
Oct  7 06:51:43 sfci-lxayws sh[21926]:     'ipblock': self.ipblock,
[...]

PR #388, opened 2020-10-08
Continued attempts to get CI passing

mikalstill:


mikalstill:
Ignoring https://github.com/shakenfist/shakenfist/issues/359 occassionally popping up and GCP sometimes giving us nodes so slow things timeout, this appears to be reliable now. Reviews appreciated.

shakenfist/config.py
```diff
@@ -137,6 +137,8 @@ def parse(self):
                     value = int(value)
                 elif isinstance(CONFIG_DEFAULTS[flag], float):
                     value = float(value)
+                elif isinstance(CONFIG_DEFAULTS[flag], bool):
+                    value = (value == 'true')
```
mandoonandy, line 141: Given that some people may use `True`, maybe
```
from distutils.util import strtobool
value = strtobool(value)
```

mandoonandy (changes requested):
One annoying suggestion because I know you love them.

shakenfist/config.py
mikalstill: Done.

mandoonandy (approved):
LGTM

PR #390, opened 2020-10-09
Bug 370 instance power off net

mandoonandy:
The failing test failed mainly because the triggers process can send multiple login_prompt events for the same login prompt. This tricked the test to continue when it shouldn't. Fixed by only looking at prompts after the test command has been issued.

Made faster by removing the reliance on sleep() and just allowing for a few initial test_ping fails.

The unpause() test always failed because it waited for login prompt.

You can see on my test cluster that the ping fails maximum of 1 despite no sleep() delays.

`test_state_changes.TestStateChanges.test_lifecycle_events` now looks like:

```
Started test instance d5da216e-be8e-4cb4-be66-928161b937ac
Started keep network alive instance
  ping test...
    _test_ping()  attempts=10
Instance Soft reboot
  ping test...
    _test_ping()  attempts=10
Instance Hard reboot
  ping test...
    _test_ping()  attempts=10
    _test_ping()  attempts=9
Power off
  ping test...
    _test_ping()  attempts=10
Instance Power on
  ping test...
    _test_ping()  attempts=10
    _test_ping()  attempts=9
Instance Pause
  ping test...
    _test_ping()  attempts=10
Instance Unpause
  ping test...
    _test_ping()  attempts=10
{0} shakenfist_ci.tests.test_state_changes.TestStateChanges.test_lifecycle_events [216.856081s] ... ok
```

Not sure this resolves all of #370 since it also uncovered the 5 min delay on the queue.

### DIFF
--- a/deploy/shakenfist_ci/base.py
+++ b/deploy/shakenfist_ci/base.py
@@ -103,12 +103,20 @@ class BaseTestCase(testtools.TestCase):
         raise TimeoutException(
             'Instance %s never triggered a login prompt after %s' % (instance_uuid, after))
 
-    def _test_ping(self, instance_uuid, network_uuid, ip, expected):
-        out, err = processutils.execute(
-            'sudo ip netns exec %s ping -c 1 %s' % (network_uuid, ip),
-            shell=True, check_exit_code=[0, 1])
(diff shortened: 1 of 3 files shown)


mikalstill (approved):
LGTM, thanks for digging into this for me.

PR #399, opened 2020-10-13
Update docs for v0.3.

mikalstill:


mandoonandy (commented):
The /docs location is okay, but do we still need a file called README?

mikalstill:
So, the README.md file is rendered on github.com, so I do think we need something there which redirects people to the website.

mandoonandy:
That was particularly unclear of me. I was trying to say, I think we should have more of a sales piece in the README but I'll merge this now and leave that for further discussion.

mandoonandy:
Also, the Source code link is self-referencing and unnecessary.

PR #405, opened 2020-10-14
394 power transition locks

### LINKED ISSUE #394: Reject poweron requests during poweroff cycle.
In general, instance commands should reject incompatible state requests.

Each state change should be analysed. Maybe we might even queue the request?

mandoonandy:
API should give a hint about server errors.
Take a lock during libvirt instance state changes.

Resolves #394

### DIFF
--- a/shakenfist/external_api/app.py
+++ b/shakenfist/external_api/app.py
@@ -862,9 +862,11 @@ class InstanceRebootSoft(Resource):
     @arg_is_instance_uuid_as_virt
     @redirect_instance_request
     def post(self, instance_uuid=None, instance_from_db=None, instance_from_db_virt=None):
-        db.add_event('instance', instance_uuid,
-                     'api', 'soft reboot', None, None)
-        return instance_from_db_virt.reboot(hard=False)
+        with db.get_lock(
+                'instance', None, instance_uuid, ttl=120, timeout=120):
+            db.add_event(
+                'instance', instance_uuid, 'api', 'soft reboot', None, None)
+            return instance_from_db_virt.reboot(hard=False)
 
 
 class InstanceRebootHard(Resource):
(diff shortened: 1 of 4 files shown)


shakenfist/external_api/app.py
```diff
@@ -139,8 +139,8 @@ def wrapper(*args, **kwargs):
                 ) as e:
             return error(401, str(e))
 
-        except Exception:
-            return error(500, 'server error')
+        except Exception as e:
+            return error(500, 'server error: %s' % e.__repr__())
```
mikalstill, line 143: Normally you'd just use e here:

```'server error: %s' % e```

Even if the repr is needed you'd normally write it like this:

```'server error: %s' % repr(e)```

File: shakenfist/virt.py [hunk 2]
```diff
@@ -471,49 +470,52 @@ def power_on(self):
                 config.parsed.get('NODE_NAME'), self.db_entry['uuid'], 'error',
                 'missing domain file in power on')
 
-        libvirt = util.get_libvirt()
-        with open(self.xml_file) as f:
-            xml = f.read()
+        with db.get_lock('virt', None, self.db_entry['uuid'], ttl=120):
```
mikalstill, line 473: This should be 'instance' instead of 'virt' shouldn't it?

shakenfist/virt.py
```diff
@@ -471,49 +470,52 @@ def power_on(self):
...
+            db.add_event('instance', self.db_entry['uuid'],
+                         'poweron', 'complete', None, None)
+            return True
 
     def power_off(self):
-        libvirt = util.get_libvirt()
-        instance = self._get_domain()
-        if not instance:
-            return
+        with db.get_lock('virt', None, self.db_entry['uuid'], ttl=120):
```
mikalstill, line 505: Ditto.

mikalstill (changes requested):

shakenfist/virt.py
mandoonandy: The problem is that `queues.py instance_delete()` takes the `instance` lock and then eventually calls `instance.power_off()`. Cleaner also calls but without a lock. That is fine since the VM is in state `deleted` anyway.

So I chose a non-colliding name that pertained only to virt.py making power state changes.

I'll take the locks higher and use `instance` locks. `virt.py` will have to trust any callers to it that they have a lock.

mandoonandy: done

shakenfist/virt.py
mandoonandy: Done.

shakenfist/virt.py
mikalstill: So yeah, the other issue is we've talked about moving to using the object to generate the name of the lock and that wouldn't have worked here.

mikalstill (approved):
LGTM

PR #417, opened 2020-10-17
Fix confusing ansible-ci failure message

mandoonandy:
Small change to clarify test failure

### DIFF
--- a/deploy/shakenfist_ci/base.py
+++ b/deploy/shakenfist_ci/base.py
@@ -129,7 +129,8 @@ class BaseTestCase(testtools.TestCase):
         self._log_console(instance_uuid)
         self._log_instance_events(instance_uuid)
         raise TimeoutException(
-            'Instance %s never triggered a login prompt after %s' % (instance_uuid, after))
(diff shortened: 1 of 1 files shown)


deploy/shakenfist_ci/base.py
```diff
@@ -120,7 +121,8 @@ def _await_event(self, instance_uuid, operation, message=None, after=None):
         self._log_console(instance_uuid)
         self._log_instance_events(instance_uuid)
         raise TimeoutException(
-            'Instance %s never triggered a login prompt after %s' % (instance_uuid, after))
+            'Instance %s never triggered event "%s":"%s" after %s' % (
+                instance_uuid, operation, message, after))
```
mikalstill, line 125: Isn't "after" here the horizon for ignored events, not the timeout? This should just say five minutes, right?

mandoonandy: [stream of thought]
I just changed "login prompt" in the existing message.

The `after` is to clarify when the test started looking for the event. In the case of login prompt it occurs many times in the event log.

I'll add a `within` to the output.

mikalstill (approved):
LGTM

PR #421, opened 2020-10-18
Revert log colourisation

mandoonandy:
Revert accidental commit that enabled colour logging... 

I'm assuming we don't want colour logs?


mikalstill (approved):
Yeah, I don't think we want that.

PR #423, opened 2020-10-19
419 fix missing exception

### LINKED ISSUE #419: Instance during CI still in preflight after timeout
shakenfist_ci.base.TimeoutException: Instance cc2ddb97-c543-428c-a49b-a64060423342 was not created in a reasonable time ({'block_devices': None, 'console_port': 32707, 'cpus': 1, 'disk_spec': [{'base': 'https://nosuch.shakenfist.com/foo', 'size': 8, 'type': 'disk'}], 'error_message': None, 'memory': 1024, 'name': 'cirros', 'namespace': 'ci-cacheimage-tidrwrzb', 'node': 'sfci-mcamyc.c.mikal-269605.internal', 'node_history': [], 'placement_attempts': 0, 'power_state': 'initial', 'requested_placement': None, 'ssh_key': None, 'state': 'preflight', 'state_updated': 1602991219.5415065, 'user_data': None, 'uuid': 'cc2ddb97-c543-428c-a49b-a64060423342', 'vdi_port': 35261, 'video': {'memory': 16384, 'model': 'cirrus'}})

mandoonandy:
Find missing image fetch exception and add event
Fix image task bugs and add image events endpoint

Requires https://github.com/shakenfist/client-python/pull/9

Resolves #419

### DIFF
--- a/deploy/shakenfist_ci/base.py
+++ b/deploy/shakenfist_ci/base.py
@@ -56,22 +56,29 @@ class BaseTestCase(testtools.TestCase):
 
     def _log_instance_events(self, instance_uuid):
         # If we've failed, log all events and then raise an exception
+        self._log_events(instance_uuid,
+                         self.system_client.get_instance_events(instance_uuid))
+
+    def _log_image_events(self, url):
+        self._log_events(url, self.system_client.get_image_events(url))
+
+    def _log_events(self, uuid, event_source):
         x = PrettyTable()
         x.field_names = ['timestamp', 'node',
                          'operation', 'phase', 'duration', 'message']
-        for e in self.system_client.get_instance_events(instance_uuid):
+        for e in event_source:
             e['timestamp'] = datetime.datetime.fromtimestamp(e['timestamp'])
             x.add_row([e['timestamp'], e['fqdn'], e['operation'], e['phase'],
                        e['duration'], e['message']])
 
         sys.stderr.write(
             '----------------------- start %s events -----------------------\n'
-            % instance_uuid)
+            % uuid)
         sys.stderr.write(str(x))
         sys.stderr.write('\n')
         sys.stderr.write(
             '----------------------- end %s events -----------------------\n'
-            % instance_uuid)
+            % uuid)
 
     def _log_netns(self):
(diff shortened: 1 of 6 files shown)


deploy/shakenfist_ci/base.py
```diff
@@ -132,6 +140,33 @@ def _await_event(self, instance_uuid, operation, message=None, after=None):
...
                 after, instance_uuid, operation, message))
 
+    def _await_image_download_success(self, url):
+        return self._await_image_event(
+            url, 'fetch', 'success')
+
+    def _await_image_download_error(self, url):
+        return self._await_image_event(
+            url, 'fetch', 'Name or service not known')
+
```
mikalstill, line 150: The only catch with this approach is it means failing tests always take 300 seconds to timeout, whereas if we asserted that waiting for a success means that there are no failures then we could fail faster?

mikalstill (commented):
Small comment.

deploy/shakenfist_ci/base.py
mandoonandy: Yes, there is a better way. I sort that out after dinner.

mikalstill:
CI here should now pass because the client change has been landed. We're not mature enough to be able to express CI dependancies yet.

mikalstill:
CI is still unhappy... https://jenkins.shakenfist.com/job/shakenfist/view/change-requests/job/PR-423/2/execution/node/19/log/

mandoonandy:
A few changes:

Fixed the tests to error immediately upon the wrong event.

Fixed queue handling: during image fetch exception handling, I think there was confusion of urllib HTTPError with shakenfist.exception.HTTPError

**Major change:** virt.py create() now only creates the virtual instance, it does not fetch any disk images. This removes the potential download errors from that code section. Instances are only created after a fetch so this step was redundant. (Let me know if I made a gross error?!)

Future change for Images - upon instantiation: check files exist or flag that fetch is required.

mikalstill:
Nice. I thought I'd already removed the fetch from virt.Instance.create(), so I am fine with that.

mikalstill (approved):
LGTM, thanks!

PR #424, opened 2020-10-19
GCP Terraform Refactor

jackadamson:
- Use terraform petnames for uniquifier
- Use terraform loop to reduce duplication
- Remove `gcp-xl` cloud and replace with `NODE_COUNT` environment variable


deploy/ansible/terraform/gcp/terraform.yml
```diff
@@ -5,80 +5,26 @@
...
   add_host:
-    hostname: sf-1
-    ansible_ssh_host: "{{terraform_out.outputs.sf_1_external.value[0]}}"
-    ansible_ssh_private_key_file: "{{ssh_key_filename}}"
+    hostname: "sf-{{node_idx+1}}"
+    node_name: "sf-{{node_idx+1}}"
+    ansible_ssh_host: "{{item}}"
+    node_ip: "{{item}}"
+    ansible_ssh_private_key_file: "{{ssh_key_filename|default(omit)}}"
     groups: hypervisors, primary, etcd_master
```
mikalstill, line 20: So this is a subtle thing in gcp-xl which has been missed... Only the first three nodes over there are etcd_masters.

deploy/ansible/terraform/gcp/terraform.yml
```diff
@@ -5,80 +5,26 @@
...
-
-- name: Add sf-3 node_ip
-  set_fact:
-    node_name: "sf-3"
-    node_ip: "{{terraform_out.outputs.sf_3_internal.value[0]}}"
-  delegate_to: sf-3
-  delegate_facts: true
+  loop: "{{terraform_out.outputs.sf_nodes_external_ip.value}}"
+  loop_control:
+    index_var: node_idx
```
mikalstill, line 23: gcp-xl also has the concept of a "primary node", which is where syslog and auth tokens go. It was done because that node is localhost in multinode CI and made log collection easier. I wonder if there's some conditional way we can do that here as well?

deploy/ansible/terraform/gcp/terraform.yml
mikalstill, line 20: We definitely can't have every node be a primary. Its used for things like the syslog collector. Only one node at a time!

deploy/ansible/terraform/gcp/terraform.yml
```diff
@@ -5,80 +5,26 @@
...
 
 - name: Log terraform hosts
   debug:
     msg:
       - "sf-1: {{hostvars['sf-1']['ansible_ssh_host']}}, {{hostvars['sf-1']['node_ip']}}"
       - "sf-2: {{hostvars['sf-2']['ansible_ssh_host']}}, {{hostvars['sf-2']['node_ip']}}"
```
mikalstill, line 29: Update or drop this?

docs/installation.md
```diff
@@ -216,13 +215,14 @@ with real users.
 | BOOTDELAY | All | How long to wait for terraform deployed instances to boot before continuing with install, in minutes |
 | SKIP_SF_TEST | All | Set to 1 to skip running destructive testing of the cloud |
 | KSM_ENABLED | All | Set to 1 to enable KSM, 0 to disable |
+| NODE_COUNT | gcp | Number of SF hypervisor nodes to deploy |
```
mikalstill, line 218: This should be down with the other GCP specific options unless we're going to add it to AWS etc.

mikalstill (changes requested):
This is cool and I appreciate you cleaning up yet another of my dirty hacks. Some small comments inline.

PR #425, opened 2020-10-19
Switch to pydantic for config management

jackadamson:
Change to using Pydantic for config management / validation.  
Aside from requiring changes to unit tests to assist stubbing, non-breaking change.  
Benefits:
- Allows easy validation of settings at launch time, including checking that IPs are valid IPs etc.
- Self documenting, putting Shakenfist in a good position to start auto generating parts of the docs from the code or checking the docs are correct
- Allows gradual adoption, as while `config.parsed.get('NODE_NAME')` will still work, we can switch to `config.parsed.NODE_NAME` to get free type checking


shakenfist/config.py
```diff
@@ -14,142 +13,100 @@
...
+    node_name = socket.getfqdn()
+    try:
+        return socket.gethostbyname(node_name)
+    except Exception as e:
+        # Only for localhost development environments
+        util.ignore_exception('config parser', e)
+        return '127.0.0.1'
+
+
+class SfConfig(BaseSettings):
```
mikalstill, line 26: Am I weird for thinking this class should be called SFConfig? The lower case 'f' jars me for some reason.

shakenfist/config.py
```diff
@@ -14,142 +13,100 @@
...
+    ###################
+    # Deployment Wide #
+    ###################
+
+    # API Options
+    API_PORT: int = Field(13000, description="Port for the REST API")
```
mikalstill, line 32: Definitely the nittiest nit ever, but I almost exclusively use single quote strings, i.e. 'Port for REST API'. Its an OpenStack thing. Andy v2 struggles with this too. Maybe I'm wrong?

shakenfist/config.py
```diff
@@ -14,142 +13,100 @@
...
+
+    class Config:
+        env_prefix = 'SHAKENFIST_'
+
+    @cached_property
+    def _dict(self):
+        return self.dict()
+
+    def get(self, key):
+        return self._dict.get(key)
```
mikalstill, line 109: This bit is really neat!

shakenfist/model.py
```diff
@@ -0,0 +1,43 @@
+# generated by datamodel-codegen:
```
mikalstill, line 1: This is obviously generated, is there a thing we need to do if we add a new config option? I feel like we need to document that somewhere because I will forget...

mikalstill (commented):
I like this idea. I especially like that it gives us type safety with config options in a way which isn't a hack like what I did, but is externally invisible to users. I'd be interested in what Andy thinks?

mandoonandy:
Very cool. I like it!

One request, can code lines be kept to 79 chars? (PEP8)

jackadamson:
> Very cool. I like it!
> 
> One request, can code lines be kept to 79 chars? (PEP8)

Happy to, I was basing the 120 lines off tools/flake8wrapper which overrides max line len to 120. Are we sticking to 80 chars per line or 120?

mandoonandy:
We try to be PEP8 (plus Michael's inverted comma thing/issue). In practice, we go past 72 columns and generally stop at 79.

That 120 limit is to allow ugly test files to be committed.

You do a raise a good point. We should probably change that since it misleads people.

shakenfist/config.py
jackadamson: Changed

shakenfist/config.py
jackadamson: Changed

shakenfist/model.py
jackadamson: Ah, this file snuck in. It was a reference while I wrote the config.py changes.

jackadamson:
All unit tests are now passing. Ready for merging!

shakenfist/config.py
jackadamson: Super sad, but just realized that `functools.cached_property` was introduced with python3.8, so have had to remove that

shakenfist/configuration.py
```diff
@@ -1,164 +1,170 @@
...
-    'STORAGE_PATH': '/srv/shakenfist',
-
-    # LOGGING
-    # -------
-    'SLOW_LOCK_THRESHOLD': 5.0,
-    'LOGLEVEL_API': 'info',
-    'LOGLEVEL_CLEANER': 'info',
-    'LOGLEVEL_MAIN': 'info',
-    'LOGLEVEL_NET': 'info',
-    'LOGLEVEL_QUEUES': 'info',
```
mandoonandy, line 101: Just a heads up. A missing log level `LOGLEVEL_QUEUES` sneaked into master in the last few days.

jackadamson: Good catch, fixed

mikalstill:
LGTM

PR #441, opened 2020-10-21
Store Image metadata in DB

mandoonandy:


mandoonandy:
This draft currently works quite well. Still to do is utilise the DB in the scheduling decision.

shakenfist/daemons/queues.py
```diff
@@ -135,9 +135,14 @@ def image_fetch(url, instance_uuid):
         instance = virt.from_db(instance_uuid)
 
     try:
-        img = images.Image(url)
-        img.get([], instance)
-        db.add_event('image', url, 'fetch', None, None, 'success')
+        # TODO(andy): Wait up to 15 mins for another queue process to download
+        # the required image. This will be changed to queue on a
+        # "waiting_image_fetch" queue but this works now.
```
mikalstill, line 140: Yes, we could inspect the lock I suppose.

shakenfist/db.py
```diff
@@ -587,3 +588,24 @@ def get_queue_length(queuename):
...
+def get_image_metadata(url_hash, node=None):
+    return etcd.get('image', url_hash, node)
+
+
+def get_image_metadata_all(only_node=None):
+    key_val = etcd.get_all_dict('image')
+
+    if only_node:
+        for k in copy.copy(key_val):
+            if not k.endswith('/'+only_node):
```
mikalstill, line 604: Spaces around the + in a perfect world.

shakenfist/images.py
```diff
@@ -40,46 +36,83 @@ def _get_cache_path():
...
-            with open(self.hashed_image_path + '.info') as f:
-                return json.loads(f.read())
-
-    def _persist_info(self):
-        with open(self.hashed_image_path + '.info', 'w') as f:
-            f.write(json.dumps(self.info, indent=4, sort_keys=True))
+        h.update(url.encode('utf-8'))
+        return h.hexdigest()
+
+    @staticmethod
```
mikalstill, line 78: This is interesting. In Instance and Network this would be done as a module level method called "from_url" or whatever. I wonder if this is a better approach and we should mvoe the others?

mikalstill: I do suspect it should be called from_url, but whatevers.

shakenfist/tests/test_db.py
```diff
@@ -7,6 +7,8 @@
 
 
 class DBTestCase(testtools.TestCase):
+    maxDiff = None
```
mikalstill, line 10: You variable name weird.

shakenfist/tests/test_etcd.py
```diff
@@ -337,3 +337,49 @@ def test_dequeue_image_fetch(self, m_delete, m_put, m_get_lock, m_get_prefix):
         ]
         self.assertCountEqual(expected, workitem['tasks'])
         self.assertSequenceEqual(expected, workitem['tasks'])
+
+
+#
+# General ETCD operations
+#
+class GeneralETCDtestCase(testtools.TestCase):
+    maxDiff = None
```
mikalstill, line 346: Ditto.

mikalstill (changes requested):
Small nits.

shakenfist/tests/test_db.py
mandoonandy: LOL. Not me bro. That's configuring testtools to output the full diff when an expect fails.

shakenfist/images.py
mandoonandy: I agree. Consistency says I should have used snake case. Fixed.

I was hoping you would like this method to roll out to the other classes. Then we can guarantee data correctness in the object more easily.

mikalstill: I would accept a patch to the other two examples to make them look like this... I do think the from_() methods should be straight after __init__ so we find them.

shakenfist/tests/test_db.py
mikalstill: Oh nice.

PR #443, opened 2020-10-22
Stop refreshing image fetch locks every chunk

### LINKED ISSUE #442: Image fetch refreshes locks too often

mandoonandy:
Resolves #442


mikalstill (approved):
Sigh.

PR #450, opened 2020-10-23
Fast path the "queue is empty" flow.

### LINKED ISSUE #449: We hold heaps of locks to check if there is any work in the queue
Instead, lets see if there is anything in the queue without a lock, and if there is then take the lock and find the work.

mikalstill:
Fixes #449.


mandoonandy (approved):
LGTM

PR #457, opened 2020-10-23
Don't copy base images during resize, create a COW chain instead.

### LINKED ISSUE #453: Don't copy image files before resize, build a cow chain instead.
We do this:

```
    util.execute(locks,
                 'cp %s %s' % (image_path + '.qcow2', backing_file))
    util.execute(locks,
                 'qemu-img resize %s %sG' % (backing_file, size))
```

That cp should probably be a COW layer instead? That would save a lot of IO and reduce block storage use as well.

mikalstill:
For large downloaded images, this is a significant performance and
IO workload improvement.

Fixes #453.

### DIFF
--- a/shakenfist/images.py
+++ b/shakenfist/images.py
@@ -148,7 +148,7 @@ class Image(object):
             diff_field = self._new_image_available(resp)
             if diff_field:
                 self.log.withField('diff_field', diff_field).info(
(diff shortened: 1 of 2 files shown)


mandoonandy (approved):
LGTM

PR #459, opened 2020-10-24
Introduce an image resize option to the cache API call.

### LINKED ISSUE #451: We should be able to prime the image cache with resized images

mikalstill:
Fixes #451.

### DIFF
--- a/shakenfist/tasks.py
+++ b/shakenfist/tasks.py
@@ -25,7 +25,8 @@ class QueueTask(object):
 
     def __eq__(self, other):
         if not QueueTask.__subclasscheck__(type(other)):
-            raise NotImplementedError('Objects must be subclasses of QueueTask')
+            raise NotImplementedError(
+                'Objects must be subclasses of QueueTask')
         return self.__hash__() == other.__hash__()
 
     def __hash__(self):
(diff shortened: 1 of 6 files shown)


shakenfist/tasks.py
```diff
@@ -175,6 +177,35 @@ def __hash__(self):
...
+    _name = 'image_resize'
+
+    def __init__(self, url, size, instance_uuid=None):
+        super(ResizeImageTask, self).__init__(url)
+        self._size = size
+        self._instance_uuid = instance_uuid
+
+    def __hash__(self):
+        return hash((super(ResizeImageTask, self).__hash__(),
+                     self._instance_uuid, self._size))
```
mandoonandy, line 199: Should include the size in the hash

mandoonandy (changes requested):
Minor change.

After we merge this, I'll change QueueTask to avoid this requirement.

shakenfist/tasks.py
mandoonandy: Next time I'll use my glasses.
Nothing to see here except "my bad".

mandoonandy (approved):
LGTM

PR #461, opened 2020-10-25
Include original qcow in backing store chain.

### LINKED ISSUE #460: CI fails to start instances
Likely a domain XML problem because of the cow-for-resize change?

mikalstill:
When we introduced qcow2 chains for resizes to reduce IO, we omitted
to add them to the domain XML as a backing store entry. This stops
VMs from starting.

Fixes #460.


mandoonandy (approved):
LGTM

PR #464, opened 2020-10-25
Fix broken get_object_lock

### LINKED ISSUE #463: New get_object_lock() is borked

mandoonandy:
Resolves #463

`object` is a type - should not be used as a variable name
No util function called unique_label()
If statement was logically flawed


mikalstill (approved):
Wow, I did really well here. Thanks.

PR #466, opened 2020-10-26
465 broken libvirt xml

### LINKED ISSUE #465: Instances do not start because libvirt xml is broken

mandoonandy:
Add missing XML forward slash and reformat to be more readable.

Resolves #465


mikalstill (approved):
I really can't have nice things.

PR #467, opened 2020-10-26
Find an available vxlan without a lock

LudvikGalois:
The old implementation takes a lock and then tries successively higher
vxlan ids until it finds one which is free. A side effect of this is
that finding an available is needlessly serial, so two networks can't
be made in parallel.

Instead of taking a lock, this uses `create` which should attempt to
atomically create the vxlan entry in etcd with successively higher
vxland ids until it succeeds. Whilst still the same inefficient
algorithm, it should allow networks to be created in parallel.


mikalstill (approved):
I think this is a good change, thanks.

PR #474, opened 2020-10-26
Remove locks in mac addr and console port generation

LudvikGalois:
Previously both of these functions took locks to avoid race conditions
at the cost of losing concurrency. These now both use an atomic create
to make sure there aren't any race conditions.

Allocating console ports changed fairly substantially - `consumed`
changed from a list to a set for faster lookups and to more strongly
hint at the usage for this variable. The `_port_free` function was
merged in, because the `try/catch/finally` of the socket bind was
needed to be part of the control flow to allow us to hold the socket
open until our port is in etcd (although this is perhaps a pre-mature
optimisation since the cost of querying etcd is not too high, and it
doesn't affect correctness, only performance).


mandoonandy (approved):
LGTM

PR #478, opened 2020-10-26
Fix bug blocking non-networked instances

### LINKED ISSUE #477: Instances without a network fail to start

mandoonandy:
Resolves #477

### DIFF
--- a/deploy/shakenfist_ci/tests/test_cirros.py
+++ b/deploy/shakenfist_ci/tests/test_cirros.py
@@ -1,0 +1,75 @@
+from shakenfist_ci import base
+
+
+class TestCirros(base.BaseNamespacedTestCase):
+    '''Make sure instances boot under various configurations.
+    '''
+    def __init__(self, *args, **kwargs):
+        kwargs['namespace_prefix'] = 'cirros'
+        super(TestCirros, self).__init__(*args, **kwargs)
+
+    def setUp(self):
+        super(TestCirros, self).setUp()
+        self.net = self.test_client.allocate_network(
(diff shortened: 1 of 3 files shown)


shakenfist/tasks.py
```diff
@@ -41,10 +41,12 @@ def json_dump(self):
 # Instance Tasks
 #
 class InstanceTask(QueueTask):
-    def __init__(self, instance_uuid, network=None):
+    def __init__(self, instance_uuid, network=[]):
```
mikalstill, line 44: You can't actually do this in python. You must use None in the method signature and then turn that into an empty list below. https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists has discussion of why, but it comes down to when the interpreter executes the method signature (once on initial parse).

mikalstill (changes requested):

shakenfist/tasks.py
mandoonandy: Very good point.

deploy/shakenfist_ci/tests/test_cirros.py
```diff
@@ -0,0 +1,75 @@
+from shakenfist_ci import base
+
+
+class TestCirros(base.BaseNamespacedTestCase):
+    '''Make sure instances boot under various configurations.
```
mikalstill, line 5: Should be """, but that's because I am a people person.

mikalstill (approved):
LGTM, except for the people person bit.

PR #482, opened 2020-10-26
Remove superfluous image resize step

### LINKED ISSUE #481: Resized COW layer could be removed
The resized COW layer for an image never stores any data. After the image resizing step changed from straight copy to a COW layer, the resized COW layer appears superfluous. It's current purpose is to save time during instance start. Creating the layer takes tenths of a second - so no real time saved.

The resizing can be done during the `qemu-img create`. The benefits are a reduction in code base size and reduced number of redundant files on disk. The only files in the image directory will the original image and the QCOW converted file.

This will also make Image handling simpler. Future image commands to delete images will be simpler due to less files that require reference counting. (Also, removal of a COW layer from libvirt could have a modicum of disk access speed improvement??)

mandoonandy:
The majority of images use QCOW backed images. Requiring a resized COW layer adds unnecessary code complexity.

Resolves #481

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,5 @@ dmypy.json
 
 # Deployer git repos
 deploy/gitrepos/
+
+.vscode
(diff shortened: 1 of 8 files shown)


mikalstill (approved):
I wonder if we need the qcow_flat and flat options at all any more?

PR #486, opened 2020-10-27
Remove withLabel from logger

mandoonandy:
`withLabel()` duplicates `withField()`

The `SFCustomAdapter` function had a bug that dropped existing fields on logger. So log 'Added event' had no useful data.


mikalstill (approved):
LGTM

PR #488, opened 2020-10-28
Check console port availability on allocated node

mandoonandy:
Check console port availability on allocated node rather than the API node.

Tidy up comments.


mikalstill (approved):
LGTM

PR #495, opened 2020-10-29
Update docs to new order of arguments for network create.

mikalstill:


mandoonandy (approved):
LGTM. Thanks. I should have thought to fix that...

PR #497, opened 2020-10-29
Actually check host networks in net daemon

### LINKED ISSUE #496: False warnings of "'Extra vxlans present!"
```
WARNING sf-net[17830] Extra vxlans present!; vxids={1, 2, 3, 4}
```
Issued when all are valid networks. They are present on the network node by design.

mandoonandy:
Resolves #496
False warnings of "'Extra vxlans present!"

Fix reversed time comparison.
Record all vxids even if new.
Use correct vxid value.

### DIFF
--- a/shakenfist/daemons/net.py
+++ b/shakenfist/daemons/net.py
@@ -56,7 +56,9 @@ class Monitor(daemon.Daemon):
                 if not n:
                     continue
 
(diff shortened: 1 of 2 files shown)


mikalstill (approved):
Sigh. LGTM.

PR #501, opened 2020-10-30
Network delete collision

### LINKED ISSUE #500: Net daemon does not check if network exists before create steps
The network creation steps can take several seconds. The API can delete the network during this period.

Due to VXLAN ID reuse this also leads to log confusion.

mandoonandy:
Stop the `net` and `queues` daemons creating networks that are simultaneously being deleted by the API daemon.

This issue manifested itself in many ways. The worst outcome was new networks reusing an old VXLAN ID whilst `sf-net` tried to recreate the old network components. Whilst the API deleted other parts of the networking infrastructure.

This PR introduces a new network state of `deleting` and the concept of a `dead` network. A network is `dead` to the system if it is in error, deleted or being deleted. A network that is `dead` to the system can not be re-created nor redeploy NAT.

Remove noisy log msgs, reduce client caused API errors to info.
Regular 'daemons running' logs are removed.
Exceptions that cause an API 500 code are logged as errors but client caused API errors are now logged as info.

Note there are two commits. The first are the networking changes. The second is lots of reformatting.

Further note, this PR comes with no guarantees. It probably still randomly fails CI....

Resolves #500

### DIFF
--- a/shakenfist/daemons/main.py
+++ b/shakenfist/daemons/main.py
@@ -124,11 +124,11 @@ def main():
 
         if not util.check_for_interface(subst['physical_bridge']):
             # NOTE(mikal): Adding the physical interface to the physical bridge
-            # is considered outside the scope of the orchestration software as it
-            # will cause the node to lose network connectivity. So instead all we
-            # do is create a bridge if it doesn't exist and the wire everything up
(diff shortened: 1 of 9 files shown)


mikalstill (approved):
LGTM.

PR #502, opened 2020-10-31
Code formatting

jackadamson:
Add python black, to automatically format code. Also sets up pre-commit to provide commit hooks to auto lint.

Would recommend reviewing by commit, as the commit where black was run, which made the most changes, guarantees an identical AST after run.

Commit where black is run: 
https://github.com/shakenfist/shakenfist/commit/d508e90d1f44248d91579d57505314cdafc9e754


mikalstill:
First off, I really like this concept, especially the git commit hooks. However, I am concerned by how big this patch is.

I've been thinking about the size, and I think my big concern is that there is a large number of lines of change for cosmetic things which make it harder for others to land patches, especially backports if we need to make them. I've joked about the string quotes, but the reality is that it is the largest causer of churn in this patch series:

With black and default args:

```
$ git diff --stat
[...]
 38 files changed, 5641 insertions(+), 4601 deletions(-)
```

With ```black --skip-string-normalization --target-version py38``:

```
$ git diff --stat
[...]
 37 files changed, 3758 insertions(+), 2718 deletions(-)
```

I do note that black is also doing something weird with long strings:

```
@@ -91,8 +90,7 @@ class SFConfig(SFConfigBase):
     # Network Options
     FLOATING_NETWORK: str = Field(
         '192.168.20.0/24',
-        description='Network block used for NAT egress from VMs and floating '
-                    'IPs',
+        description='Network block used for NAT egress from VMs and floating ' 'IPs',
```

See how it didn't combine those two strings?

So, how would you feel about us using a different set of flags to black to reduce the size of the code churn? We'd need to manually cleanup those weird strings I suppose.

mandoonandy:
I agree that we should probably skip the changing of the quotes. (Much as I hate to say that!)

If we could limit squash the commits in the code change that would good. Preferably to just one for the actual changing code.

Then we list the commit(s) in a `.git-blame-ignore-revs`

BUT I would like to wait until CI works properly (which is very close).

jackadamson:
That seems reasonable to me. So for the moment, I'll hold off until CI is fully working, then will configure black to ignore quote normalization, fix the multiline strings, squash the commits that apply formatting, and add them to a `.git-blame-ignore-revs` file.  Does that sounds right?

mandoonandy:
Sounds reasonable to me. 

CI *should* be fixed. Up to our BGOMFL to decide how many good CI runs constitutes stability.

PR #505, opened 2020-11-01
Avoid Network deletion whilst loading from the DB

mandoonandy:
Lock Network whist loading the object from the DB

The avoids an Exception caused by the deletion of the Network whilst it is being loaded. The Exception is raised by attempting to load an already deleted IPManager.

Also removed an unnecessary Network DB reload.

Small modification to repr() of QueueTasks.


mikalstill (approved):
LGTM

PR #510, opened 2020-11-02
Retry downloads after checksum failures.

### LINKED ISSUE #509: Image fetch should retry after a BadChecksum exception

mandoonandy:
Previously, after a single download checksum failure, the instance start would be aborted.

The Image lock has moved above Image.get() therefore there is no lock to retry.

Resolves #509


mandoonandy:
CI correctly caught me introducing a bug.

mikalstill (approved):
LGTM

PR #519, opened 2020-11-02
Increase CI instance creation time allowance

mandoonandy:
It can take over two minutes to download a Ubuntu image. This causes failures in CI _await_instance_event().


mikalstill (approved):
LGTM

PR #521, opened 2020-11-05
Initial attempt at glusterfs support.

mikalstill:
This doesn't work except for the fuse mount case. You're welcome.


mikalstill:
I just realised I need to make a new pull request because this one is against master not develop (it predates develop I think).

PR #537, opened 2020-11-12
Require that disk size be an int, even if bad in db.

### LINKED ISSUE #536: Instances failing to start if size is passed as a string (which client-python does)
This command line:

```
sf-client instance create disk-thing 1 1024 \
    -D size=8,base=cirros,bus=ide,type=disk \
    -D size=16,type=disk \
    -D base=http://archive.ubuntu.com/ubuntu/dists/focal/main/installer-amd64/current/legacy-images/netboot/mini.iso
```

Causes this exception:

```
Nov 12 07:00:33 syd-sf-3 ERROR sf-queues-1605164279.3725202-000[1529097] [Exception] Ignored error in sf-queues: %d format: a number is required, not str
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/shakenfist/daemons/queues.py", line 82, in handle
    instance_start(instance_uuid, task.network())
  File "/usr/local/lib/python3.8/dist-packages/shakenfist/daemons/queues.py", line 241, in instance_start
    instance.create(lock=lock)
  File "/usr/local/lib/python3.8/dist-packages/shakenfist/virt.py", line 196, in create
    images.create_cow([lock], hashed_image_path,
  File "/usr/local/lib/python3.8/dist-packages/shakenfist/images.py", line 331, in create_cow
    'qemu-img create -b %s -f qcow2 %s %dG'
TypeError: %d format: a number is required, not str
```

mikalstill:
Fixes #536.

### DIFF
--- a/shakenfist/virt.py
+++ b/shakenfist/virt.py
@@ -61,6 +61,12 @@ def _get_defaulted_disk_type(disk):
     return 'disk'
 
 
+def _safe_int_cast(i):
+    if i:
+        return int(i)
+    return i
+
+
 class Instance(object):
     def __init__(self, db_entry):
         self.db_entry = db_entry
@@ -103,7 +109,7 @@ class Instance(object):
             'devices': [
                 {
                     'type': disk_type,
-                    'size': self.db_entry['disk_spec'][0].get('size'),
+                    'size': _safe_int_cast(self.db_entry['disk_spec'][0].get('size')),
                     'device': root_device,
                     'bus': bus,
                     'path': os.path.join(self.instance_path, root_device),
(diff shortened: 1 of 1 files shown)


shakenfist/virt.py
```diff
@@ -103,7 +109,7 @@ def _populate_block_devices(self):
             'devices': [
                 {
                     'type': disk_type,
-                    'size': self.db_entry['disk_spec'][0].get('size'),
+                    'size': _safe_int_cast(self.db_entry['disk_spec'][0].get('size')),
```
mandoonandy, line 112: If it's not an integer then what is the value?
At this point, the size has to be an integer?
A `.get('size', 0)` could be neater but would push the exception/error further down the road?

mandoonandy: To unblock your CI work lets accept this, and promise this debt will be repaid during the API input clean up.

mikalstill: Yeah, I was thinking about how get() can return None, but you're right that providing a default would work. That said this is still wrong. We use None later as a special case for "detect", so I need to fix this.

mikalstill: No wait its right. We want int or None here.

PR #539, opened 2020-11-13
Fix bug in IP address check

### LINKED ISSUE #538: Client cannot create an instance with a network without specifying the IP address
```
$ sf-client instance create bob 1 1024 -d 13@cirros -n a4fb84b7-be80-44f4-9f22-f6b3f9813caf 
ERROR: Internal Server Error: {"error": "server error: ValueError('None does not appear to be an IPv4 or IPv6 address',)", "status": 500}
```
Caused by change to ensure IP addresses are in range.

mandoonandy:
Resolves #538


mikalstill (approved):
LGTM, thanks.

PR #543, opened 2020-11-15
Change Instance object model

### LINKED ISSUE #484: Move Instance to new object model

mandoonandy:
Move Instance class to similar model used by Image.

Add appropriate test cases.

Also fix daemon/cleaner.py L#64
* Code was not executable. Since not causing exceptions it may not be even be needed.

Resolves #484


mikalstill:
Looks like there is a failure to mock etcd in here somewhere?

mandoonandy:
Reversed the change from the last push. (Lesson re-learnt: Don't rush small changes)

shakenfist/virt.py
```diff
@@ -68,106 +53,280 @@ def _safe_int_cast(i):
...
+        self.power_state = power_state
+        self.power_state_previous = power_state_previous
+        self.power_state_updated = power_state_updated
+        self.requested_placement = requested_placement
+        self.ssh_key = ssh_key
+        self.state = state
+        self.state_updated = state_updated
+        self.user_data = user_data
+        self.vdi_port = vdi_port
+        self.video = video
```
mikalstill, line 87: For these, what would you think about using properties a bit more to hide the setter methods?

```
$ cat accessor.py 
#!/usr/bin/python3

class C(object):
    def __init__(self, x):
        self.__set_x(x)

    def __get_x(self):
        return self._x

    def __set_x(self, x):
        self._x = x

    x = property(__get_x, __set_x)


foo = C(1)
print(foo.x)
foo.x = 2
print(foo.x)
```

Which would return:

```
$ python3 accessor.py 
1
2
```

Notably this would not work with the setters where we take more than one argument, but I am unsure how common that is without an audit.

mikalstill (commented):
I'd be ok with the property thing happening later in this series if we agree its the right path.

mikalstill:
And you can in fact pass multiple arguments to a property setter, although there is debate as to if it is a good idea or not. https://stackoverflow.com/questions/18714262/property-setter-with-multiple-values

shakenfist/virt.py
mandoonandy: I like that idea. Will make it so!

mikalstill:
We've verbally agreed we'll do properties at the end of this patch series, so I am going to merge this one.

PR #544, opened 2020-11-15
Ensure image is re-fetched after checksum failure

### LINKED ISSUE #542: Image checksum error does not cause new download attempt

mandoonandy:
Fix bug in Image fetch. Add extra unit tests for the functions involved.

Resolves #542


mikalstill (approved):
LGTM

Issue #552, opened 2020-11-23
ETCD server error causes locks to fail other tasks


mandoonandy:
CI run failed due to undeleted instance holding network.
https://jenkins.shakenfist.com/blue/organizations/jenkins/shakenfist/detail/PR-551/1/pipeline

I'm not sure what happened here. It is intermittent due randomness from the etcd failure and the timing of the failure.

Etcd failed, it appears during a lock release. My first thought was the TTL didn't expire in time for the instance delete to occur but that is not supported by the time stamps. Lock TTL=120.

```
Nov 23 10:09:06 sfci-ypk8jq INFO sf-queues-1606126137.8121235-000[7446] Added event; duration=4.381286859512329; fqdn=sfci-ypk8jq.c.mikal-269605.internal; instance=7dd52ae8-0811-4aa2-99b9-9f92c4e501de; message=None; operation=instance creation; phase=finish
Nov 23 10:09:06 sfci-ypk8jq INFO sf-queues-1606126137.8121235-000[7446] Finish instance creation; duration=4.381286859512329; instance=7dd52ae8-0811-4aa2-99b9-9f92c4e501de
Nov 23 10:09:06 sfci-ypk8jq ERROR sf-queues-1606126137.8121235-000[7446] [Exception] Ignored error in sf-queues: Internal Server Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/queues.py", line 81, in handle
    instance_start(instance_uuid, task.network())
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/queues.py", line 255, in instance_start
    instance.update_instance_state('created')
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/etcd.py", line 114, in __exit__
    if not self.release():
  File "/usr/local/lib/python3.6/dist-packages/etcd3gw/lock.py", line 91, in release
    result = self.client.transaction(txn)
  File "/usr/local/lib/python3.6/dist-packages/etcd3gw/client.py", line 345, in transaction
    data=json.dumps(txn))
  File "/usr/local/lib/python3.6/dist-packages/etcd3gw/client.py", line 88, in post
    resp.reason
etcd3gw.exceptions.InternalServerError: Internal Server Error
```
Later causing the delete job to be dropped
```
[...]

PR #556, opened 2020-11-24
Network object model

mandoonandy:
First tranche of changes to the Network object

**TODO**
* Further Unit tests
* Move network processing from db to net
* Move IPM settings out of Network object
* Add functions to protect handling of floating network (and avoid direct DB access)
* Comb through code for improved net methods


mikalstill:
CI looks like it failed because etcd was unhealthy. Rerunning now.

mandoonandy:
Yep. Looks like a repeat of #552

mikalstill (approved):
LGTM, thanks.

PR #561, opened 2020-11-27
Provide explicit URI for libvirt.

### LINKED ISSUE #202: libvirt open doesn't use a URI
https://github.com/shakenfist/shakenfist/blob/d599ec1e1720292c924583c757a48ce433709556/shakenfist/virt.py#L436

Currently we rely on libvirt finding the correct driver to use based on it's heuristic probe. We should be explicitly stating the driver. 

An example for local host system KVM: `libvirt.open('qemu:///system')`

mikalstill:
Fixes #202.


mandoonandy (approved):
LGTM

PR #566, opened 2020-11-28
Tweak the name of persist_instance_metadata.

mikalstill:
I feel like the term "metadata" is used in a different way elsewhere
in the codebase, and this method is really persisting the instance,
not data about the instance. Its a nit, but I found it confusing
while reading this code.


mandoonandy (approved):
Yes, I can see that.

LGTM

PR #567, opened 2020-11-28
Non volatile objects

mikalstill:
What if the Instance object only cached values gauranteed to never change, and volatile values were stored in little subrecords in the database which were fetched and updated as required? Would this solve our stale caching problem?


mikalstill:
Not expected to pass unit tests or CI. This is a draft for comment.

mikalstill:
I think failing all checks might be a personal best.

shakenfist/db.py
```diff
@@ -146,13 +161,86 @@ def persist_ipmanager(network_uuid, data):
...
+# Theorem: we never write to the instance object, except for when we
+# create or delete it. Anything which changes value during the lifetime
+# of the instance (state, power_state, etc) is stored in a set of
+# secondary values in etcd that we never cache. This is version 2 of
+# the instance object.
+
+
+def create_instance(instance_uuid, cpus, disk_spec, memory, name, namespace,
+                    requested_placement, ssh_key, user_data, video):
+    metadata = {
```
mandoonandy, line 177: I think it would be better to keep object metadata generation in the class code, like `Network.metadata()`

shakenfist/db.py
```diff
@@ -146,13 +161,86 @@ def persist_ipmanager(network_uuid, data):
...
+        'video': video,
+
+        'version': 2
+    }
+    etcd.put('instance', None, instance_uuid, metadata)
+
 
 def get_instance(instance_uuid):
-    return etcd.get('instance', None, instance_uuid)
+    # TODO(mikal): handle upgrade here
```
mandoonandy, line 196: I was going to suggest a separate upgrade tool but having the ability to auto-upgrade without downtime is pretty cool. 

I don't like the complexity of having upgrade code in the middle of the main code, especially because it will keep growing.

I don't have a good solution if we want to keep the ability to upgrade without downtime.

shakenfist/daemons/queues.py
```diff
@@ -277,9 +277,7 @@ def instance_delete(instance_uuid, new_state, error_msg=None):
         instance = virt.Instance.from_db(instance_uuid)
         if instance:
             instance.delete()
-            instance.update_instance_state(new_state)
-            if error_msg:
-                instance.update_error_message(error_msg)
+            instance.update_instance_state(new_state, error_message=error_msg)
```
mandoonandy, line 280: I left a lot of inefficient function names because I didn't the want the code change to be too massive.

mandoonandy (commented):
I like this idea. It improves our use of the KV store.

How will locking be applied?

shakenfist/db.py
mikalstill: I can do that. I think I might try and get tests passing at all first, but make sure I don't forget.

mikalstill:
> How will locking be applied?

I _think_ we should be ok if we just lock around the attribute changes, which are small and fast. Perhaps a lock per attribute record. Thoughts?

shakenfist/db.py
mikalstill: Its a fair point. I think if we go this route we will _have_ to do some upgrade CI as well to ensure that existing instances aren't clobbered. An upgrade tool is tempting as I agree it would make this code a lot simpler.

mikalstill:
Oh, and I feel a bit like events/instance/<<uuid>> should become instance/<<uuid>>/events as part of this transition. Thoughts?

mandoonandy:
> > How will locking be applied?
> 
> I _think_ we should be ok if we just lock around the attribute changes, which are small and fast. Perhaps a lock per attribute record. Thoughts?

I agree. Keep the locks small as possible. It will have to be case by case anyway. Good opportunity to review our lock use.

shakenfist/db.py
mikalstill: This is done in the next version.

mikalstill:
Unit tests and lint should now pass. Still to do:

* Upgrades (tested with CI)
* Locking
* Move events

mikalstill:
Sigh, needed a rebase.

mikalstill:
> Oh, and I feel a bit like events/instance/uuid should become instance/uuid/events as part of this transition. Thoughts?

Actually, I can't do this, because it would make get_instances() very slow, as it iterates the entire tree.

mikalstill:
I do hope you're all enjoying this at home.

mikalstill:
CI failure is now unrelated (network race). Still to do:

* Upgrades (tested with CI)
* Locking

I've decided moving events is a bad idea for now.

mikalstill:
Locking done, with CI failures being bogus network things. Final step is upgrades and testing of upgrades.

shakenfist/db.py
```diff
@@ -146,47 +162,83 @@ def persist_ipmanager(network_uuid, data):
...
+    # done as part of the upgrade process.
+    i = etcd.get('instance', None, instance_uuid)
+    if i.get('version') != 2:
+        raise exceptions.BadObjectVersion(
+            'Unknown version - Instance: %s', i)
 
-def get_instance(instance_uuid):
-    return etcd.get('instance', None, instance_uuid)
+    # If this is an external view, then mix back in attributes that users expect
+    if external_view:
```
mandoonandy, line 190: These could be retrieved via a property function.

mikalstill: Is that true if we're returning a dict here not an object? I wanted to have these be explicit so we don't rely on this behaviour in internal code, its just here so API callers get the data back in the right format.

mandoonandy: I was hoping to remove system logic from the DB interface layer. All instance logic could (should) be in the Instance class. Leaving DB to the be shim that gathers/persists data.

shakenfist/virt.py
```diff
@@ -53,39 +53,82 @@ def _safe_int_cast(i):
...
-        self.state = state
-        self.state_updated = state_updated
-        self.user_data = user_data
-        self.vdi_port = vdi_port
-        self.video = video
+    def __init__(self, static_values):
+        # This dictionary contains values which will never change for this
+        # instance. It is therefore safe to cache forever. I know this moves
+        # us away from member variables again, but I wanted to make it very
+        # clear that these values are not meant to be assigned to.
```
mandoonandy, line 125: These static values complicate (look non-pretty) the use of the class. We can enforce the read-only nature using properties.

shakenfist/virt.py
```diff
@@ -698,17 +667,19 @@ def _snapshot_device(self, source, destination):
...
     def snapshot(self, all=False):
-        disks = self.block_devices['devices']
+        disks = db.get_instance_attribute(
+            self.static_values['uuid'], 'block_devices')['devices']
         if not all:
             disks = [disks[0]]
 
         snapshot_uuid = str(uuid4())
-        snappath = os.path.join(self.snapshot_path(), snapshot_uuid)
+        snappath = os.path.join(_snapshot_path(
```
mandoonandy, line 676: This reverses `snapshop_path` being a property of an instance?

shakenfist/virt.py
```diff
@@ -616,27 +582,30 @@ def _create_domain_xml(self):
...
         # I hadn't spent _ages_ finding a bug related to it.
+        block_devices = db.get_instance_attribute(
+            self.static_values['uuid'], 'block_devices')
+        ports = db.get_instance_attribute(self.static_values['uuid'], 'ports')
         xml = t.render(
-            uuid=self.uuid,
-            memory=self.memory * 1024,
-            vcpus=self.cpus,
-            disks=self.block_devices['devices'],
+            uuid=self.static_values['uuid'],
```
mandoonandy, line 589: It would be nice to make this code fragment `static_values['uuid']` simpler and less likely to be mistyped.

A possible source of errors is unchecked constants like `'uuid'`. 

(We would never make a typo but others might.)

mikalstill:
That was a painful rebase I am hoping I haven't messed up.

mikalstill:
This has all been merged in bite sized chunks.

PR #574, opened 2020-12-05
Optionally ignore stale nodes, especially in scheduling.

### LINKED ISSUE #573: We should not place instances on nodes which haven't been seen in a while
If a node hasn't checked in in five minutes or so, we should stop trying to schedule to it. We probably shouldn't put the instances into an error state though.

mikalstill:
Fixes #573.


mandoonandy (approved):
LGTM

Issue #577, opened 2020-12-06
Frequent single node CI failures for missing physical interfaces


mikalstill:
We see this a lot in develop:

```
Dec  6 16:30:16 sfci-1pfefh sh[2969]: Traceback (most recent call last):
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/bin/sf-daemon", line 8, in <module>
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     sys.exit(main())
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 169, in main
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     _audit_daemons()
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 162, in _audit_daemons
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     _start_daemon(d)
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 103, in _start_daemon
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     DAEMON_IMPLEMENTATIONS[d].Monitor(d).run()
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 188, in run
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     self._process_network_node_workitems()
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 154, in _process_network_node_workitems
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     n.create()
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/net.py", line 320, in create
Dec  6 16:30:16 sfci-1pfefh sh[2969]:     self.deploy_nat()
Dec  6 16:30:16 sfci-1pfefh sh[2969]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/net.py", line 358, in deploy_nat
[...]

PR #578, opened 2020-12-06
Chase why we're queuing messages for the network node in a single node deployment

mikalstill:
I am debugging #577.


mikalstill:
I don't think I've proven that this is a cause of CI flakeyness, so closing.

PR #583, opened 2020-12-07
Move from a dict to properties for static values in virt.

mikalstill:


shakenfist/virt.py
```diff
@@ -170,47 +178,203 @@ def from_db(uuid):
...
+
+    @uuid.setter
+    def uuid(self, _):
+        raise exceptions.NotImplementedException(
+            'You cannot set a static variable')
+
+    @uuid.deleter
+    def uuid(self):
+        raise exceptions.NotImplementedException(
+            'You cannot delete a static variable')
```
mandoonandy, line 195: I think you could reduce code by not defining the `setter` and `deleter`. If they are called I think you get an `AttributeError` which is probably more pythonic.

mandoonandy (commented):
Looks very pretty to me! (Just one question)

shakenfist/virt.py
mikalstill: Oh interesting. I will test this when I am not in a meeting.

mikalstill: And you're right. Fixed.

mandoonandy (approved):
LGTM

PR #587, opened 2020-12-12
Move to a more OO Instance.

mikalstill:
This doesn't completely remove the old way of using the database, but it's a step in that direction without being an impossibly large amount of code to review.


mikalstill:
Chasing some weirdness in CI.

mikalstill:
The filtering of Instances is causing me some issues...

mikalstill:
This PR fixes #590.

shakenfist/virt.py
```diff
@@ -138,6 +140,14 @@ def __init__(self, static_values):
             LOG.withObj(self).error('Found disk spec empty')
             raise exceptions.InstanceBadDiskSpecification()
 
+    @staticmethod
+    def _db_create_instance(instance_uuid, metadata):
```
mandoonandy, line 144: Should this just be in new()?

shakenfist/virt.py
```diff
@@ -793,3 +862,53 @@ def get_console_data(self, length):
...
+                # the result set.
+                if not f(i):
+                    skip = True
+                    break
+
+            if not skip:
+                yield i
+
+
+def placement_filter(node, inst):
```
mandoonandy, line 889: This does more than filter?

shakenfist/virt.py
```diff
@@ -793,3 +862,53 @@ def get_console_data(self, length):
             'Client requested %d bytes of console log, returning %d bytes'
             % (length, len(d)))
         return d
+
+
+class Instances(object):
```
mandoonandy, line 867: This is not a concept that I find easy to like. If we are going OO maybe this should be a factory class of some sort?

mandoonandy (commented):
Looks good overall. I have a little uneasiness about the Instances class, but I am liking the move towards OO'ness making the code easier to read.

mikalstill:
Replaced by #596 because rebase.

PR #591, opened 2020-12-13
Chasing these missing phy-XXX-i errors.

mikalstill:
- Network node setup is always done via a queue message, even if local.
- DHCP is no longer recreated for every instance on a freshly started
  hypervisor. Instead, we do that when the network node restarts.


mikalstill:
It doesn't get to the destination, but its getting us closer and is getting a bit big. How do you feel about merging this one? Note that the biggest change here is that _all_ network node work is now done via queues, there is no fast path for if you're on the network node already.

mandoonandy (approved):
LGTM

PR #597, opened 2020-12-16
Be less permissive of primitives not existing.

### LINKED ISSUE #595: Be less permissive of manipulated object state in the API
At the moment, if you request an instance on a network which isn't yet created (its being built), we patiently wait and then do that thing. Simple is beautiful. We should simply reject your request and tell you to come back later.

mikalstill:
Don't poll or wait if the network or instance being managed doesn't
exist, just tell the client to come back later. This fixes #595.


mandoonandy (approved):
LGTM

PR #598, opened 2020-12-19
Move remaining attribute calls into the Instance object.

mikalstill:


mikalstill:
I think this one is ready for review now. The remaining failure appears to be flakey networking.

shakenfist/tests/test_virt.py
```diff
@@ -171,93 +171,124 @@ def _make_instance(self, mock_get_instance, mock_create_instance):
...
-                })
+    @mock.patch('shakenfist.virt.Instance.state',
+                new_callable=mock.PropertyMock)
+    @mock.patch('shakenfist.virt.Instance._db_set_attribute')
     @mock.patch('shakenfist.etcd.put')
-    def test_update_instance_state(self, mock_put, mock_get_attribute,
-                                   mock_lock):
+    def test_update_instance_state(
+            self, mock_put, mock_attribute_set, mock_state_get, mock_lock):
+        mock_state_get.return_value = {
```
mandoonandy, line 180: Would it be more consistent to put the return value in the PropertyMock instantiation?

shakenfist/virt.py
```diff
@@ -443,22 +477,18 @@ def hard_delete(self):
     def allocate_instance_ports(self):
         with db.get_lock('attribute/instance', self.uuid, 'ports',
```
mandoonandy, line 478: As an aside, I think it would be better to use a lock construction method to guarantee consistent lock names.

shakenfist/virt.py
```diff
@@ -285,12 +296,43 @@ def video(self):
     def version(self):
         return self.__version
 
+    # Values routed to attributes, writes are via helper methods.
```
mandoonandy, line 299: I feel attribute setters should be defined if needed, rather than using helper methods elsewhere. The code would be safer if get/set are strictly controlled.

mandoonandy (approved):
Just a few comments on style not substance. I want to get CI working so I'm happy to merge and address them later.

I'll leave the merge up to you.

shakenfist/tests/test_virt.py
mikalstill: I actually couldn't get that to work... I ended up doing this because its what the docs told me to do.

shakenfist/virt.py
mikalstill: That's true, except this name doesn't fit with our current naming scheme. I need to fix that.

shakenfist/virt.py
mikalstill: The helpers don't fit the signature of a setter is why they're done this way at the moment. We can work through that though? I think we could pass a tuple to a setter for example.

shakenfist/tests/test_virt.py
mandoonandy: This seems to work. (Ignoring the rubbish spacing)
```
     @mock.patch('shakenfist.db.get_lock')
     @mock.patch('shakenfist.virt.Instance.power_state',
-                new_callable=mock.PropertyMock)
+                new_callable=mock.PropertyMock(return_value={
+                                                'power_state_previous': 'off',
+                                                'power_state': 'on',
+                                                'power_state_updated': 0
+                                            }))
     @mock.patch('shakenfist.virt.Instance._db_set_attribute')
     def test_update_power_state(
             self, mock_attribute_set, mock_power_state_get, mock_lock):
-        mock_power_state_get.return_value = {
-            'power_state_previous': 'off',
-            'power_state': 'on',
-            'power_state_updated': 0
-        }
 
         i = self._make_instance()
         i.update_power_state('off')
```

PR #599, opened 2020-12-19
Add experimental new API using FastAPI

LudvikGalois:
The current external API is written using flask_restful. This is functional, but currently requires separate documentation which may fall out of date.

In theory (at least) moving to FastAPI should allow documentation to be automatically generated it in the OpenAPI format, simplifying the work involved in writing a client as well reducing drift between documentation and implementation.

Currently, the only additions to the API, are a `/token` end point which accepts credentials as form data to support the bare minimum of OAuth 2 required for the docs to be interactive, and `/docs` (automatically added by FastAPI) which serves the the documentation.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,6 +18,10 @@ bcrypt              # apache2
 gunicorn            # mit
 pylogrus            # mit
 pydantic            # mit
+fastapi             # mit
+python-multipart    # apache2
+python-jose
+cryptography
 
 # Is difficult to get install working, use system packages instead. On Ubuntu
 # those are: libvirt-daemon-system libvirt-dev python3-libvirt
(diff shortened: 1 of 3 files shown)


mikalstill:
I was talking to Andy v1 about this today and realised that part of the issue is that the API layer is very "business logicky". I wonder if this would be easy if the business logic was refactored out to somewhere else and then each API implementation just called that one layer instead of repeating the code in two places?

LudvikGalois:
I was thinking about that as well. That seems like a perfectly reasonable idea, even if there was only one API.
But it clashes a little with my plans to provide `db` via dependency injection in the Fast API implementation to simplify testing.

Can I sell you on something like

```python
class SF:
    def __init__(self, db):
        self.db = db
    def create_namespace(some, args, here):
        pass
    # etc
```
where it's an explicit parameter?

LudvikGalois:
Before this can be properly addressed, more substantial changes need to be made, so I'm closing it.

PR #603, opened 2020-12-24
The vxid passed to the ip command must be decimal.

mikalstill:


shakenfist/net.py
```diff
@@ -209,9 +209,12 @@ def update_floating_gateway(self, gateway):
 
     def subst_dict(self):
         # NOTE(mikal): it should be noted that the maximum interface name length
-        # on Linux is 15 user visible characters.
+        # on Linux is 15 user visible characters, we therefore use hex for vxids
+        # where they appear in an interface nate. Note that vx_id does not appear
```
mandoonandy, line 213: nate?

mandoonandy (changes requested):
The spelling nazi rides again.

shakenfist/net.py
mikalstill: Its short for Nathan I believe.

mandoonandy: Maaate

mikalstill:
Self merge after verbal approval from Andy v2.

PR #604, opened 2020-12-24
Turn snapshot requests into task

LudvikGalois:
Right now, snapshot aren't tasks. This is a problem since they are
definitely a long lived operation which we can reasonably expect to
time out.

A new field has been added to snapshots, `status`, to track whether or
not they're creating or ready for use. This is perhaps too ad-hoc and
will probably need to change once there's a more robust story around
snapshots.

The API now returns the a UUID for a snapshot immediately instead of
waiting for the operation to finish, since my (possibly wrong)
assumption is that it's very rare for a user to want to immediately
want to use a snapshot that they've made.

### DIFF
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -23,6 +23,7 @@ from shakenfist.tasks import (QueueTask,
                               InstanceTask,
                               PreflightInstanceTask,
                               StartInstanceTask,
+                              SnapshotInstanceTask,
                               )
 
 
@@ -94,6 +95,17 @@ def handle(jobname, workitem):
                     db.enqueue('%s-metrics' % config.NODE_NAME, {})
                 except Exception as e:
                     util.ignore_exception(daemon.process_name('queues'), e)
+
+            elif isinstance(task, SnapshotInstanceTask):
+                snap_uuid = task.snap_uuid
+                instance = virt.Instance.from_db(instance_uuid)
+                if instance is None or instance.status != 'created':
+                    log_i.error('Instance not ready, unable to snapshot')
+                    return
+                instance.snapshot(all=all, snapshot_uuid=snap_uuid)
+                instance.add_event('api', 'snapshot (all=%s)' % all,
+                                   None, snap_uuid)
+                db.add_event('snapshot', snap_uuid, 'api', 'create', None, None)
 
             else:
                 log_i.withField('task', task).error('Unhandled task - dropped')
--- a/shakenfist/db.py
+++ b/shakenfist/db.py
@@ -235,8 +235,16 @@ def create_snapshot(snapshot_uuid, device, instance_uuid, created):
             'uuid': snapshot_uuid,
             'device': device,
             'instance_uuid': instance_uuid,
-            'created': created
+            'created': created,
+            'status': 'creating'
         })
+
+
+def finalize_snapshot(snapshot_uuid, instance_uuid, created):
+    snapshot = etcd.get('snapshot', instance_uuid, created)
+    snapshot['status'] = 'created'
+    etcd.put(
+        'snapshot', instance_uuid, created, snapshot)
 
 
 def get_instance_snapshots(instance_uuid):
--- a/shakenfist/tasks.py
+++ b/shakenfist/tasks.py
@@ -99,6 +99,14 @@ class ErrorInstanceTask(InstanceTask):
         return self._error_msg
 
 
+class SnapshotInstanceTask(InstanceTask):
+    _name = 'instance_snapshot'
+
+    def __init__(self, instance_uuid, snap_uuid):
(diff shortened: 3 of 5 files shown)


shakenfist/db.py
```diff
@@ -235,10 +235,18 @@ def create_snapshot(snapshot_uuid, device, instance_uuid, created):
             'uuid': snapshot_uuid,
             'device': device,
             'instance_uuid': instance_uuid,
-            'created': created
+            'created': created,
+            'status': 'creating'
         })
```
mikalstill, line 240: I thought this was your error until I expanded the view and realised its mine. Its confusing to me that the path for a snapshot is the time it was created, not the snapshot UUID. That means for every request for a specific snapshot we need to iterate through all the snapshots for that instance. I suspect that's a more common operation than wanting to list them by time? Thoughts?

shakenfist/db.py
```diff
@@ -235,10 +235,18 @@ def create_snapshot(snapshot_uuid, device, instance_uuid, created):
...
 
 
+def finalize_snapshot(snapshot_uuid, instance_uuid, created):
+    snapshot = etcd.get('snapshot', instance_uuid, created)
+    snapshot['status'] = 'ready'
```
mikalstill, line 245: "ready" is a new name for a state. The current ones are: preflight; creating; created; deleting; deleted.

Do we think there is value in having a standard naming convention for states? I suspect it would reduce errors in the future?

shakenfist/tasks.py
```diff
@@ -99,6 +99,25 @@ def error_msg(self):
...
+
+        self._network = []
+        # General checks
+        if not instance_uuid:
+            raise NoInstanceTaskException(
+                'No instance specified for InstanceTask')
+        if not isinstance(instance_uuid, str):
+            raise NoInstanceTaskException('Instance UUID is not a string')
+        if not isinstance(self._network, list):
+            raise NetworkNotListTaskException()
```
mikalstill, line 118: There is no network here?

mikalstill (commented):
This looks generally really good, sorry I took so long to review it. Some comments inline.

shakenfist/db.py
LudvikGalois: That seems perfectly reasonable to me

shakenfist/tasks.py
LudvikGalois: That comes from being copied and pasted. I feel like I've been a bit of a potato about the entire init thing, given that the requisite logic is in the `__init__` for `InstanceTask` which I should already be calling, but instead I'm calling `__init__` for `QueueTask`. Let me unbreak that now

shakenfist/db.py
LudvikGalois: Yes, it does seem a bit strange, but right now listing all the snapshots for an instance is the only way to fetch snapshots. I feel like this should wait until there's a more fleshed out story around snapshots. It also blocks making them an superclass of `baseobject.DatabaseBackedObject` instead of just a dict, since `from_db` can't be reasonably implemented in a sane way.

shakenfist/daemons/queues.py
```diff
@@ -95,6 +96,14 @@ def handle(jobname, workitem):
                 except Exception as e:
                     util.ignore_exception(daemon.process_name('queues'), e)
 
+            elif isinstance(task, SnapshotInstanceTask):
+                snap_uuid = task.snap_uuid
+                instance_from_db = virt.Instance.from_db(instance_uuid)
```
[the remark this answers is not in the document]
LudvikGalois: Changed to `instance`

shakenfist/daemons/queues.py
```diff
@@ -95,6 +96,14 @@ def handle(jobname, workitem):
                 except Exception as e:
                     util.ignore_exception(daemon.process_name('queues'), e)
 
+            elif isinstance(task, SnapshotInstanceTask):
+                snap_uuid = task.snap_uuid
+                instance = virt.Instance.from_db(instance_uuid)
+                instance.snapshot(all=all, snapshot_uuid=snap_uuid)
```
mandoonandy, line 102: Probably should check that the instance is not deleted here? Also, check that the instance exists at all.

mandoonandy (changes requested):
I like this becoming a queued operation.

shakenfist/daemons/queues.py
LudvikGalois: You're right. This is checked when the task is created, but not again when it's time to handle it and things could have changed since them.

shakenfist/db.py
mikalstill: Yeah, I feel like most (all?) things that land in the database should eventually become children of DatabaseBackedObject, but it is fair to say we're going to need to take some time to get there.

Issue #609, opened 2020-12-29
Net daemon crashes after upgrade


mandoonandy:
Upgrade needs to rename interface to new 6 char hex name extension.
```
Dec 29 12:15:57 sf-3 INFO sf-net[9845] Finish create vxlan interface; duration=0.02299785614013672; network=d56db115-8321-4c0a-b70f-821d6b28c22d
Dec 29 12:15:57 sf-3 sh[9194]: Traceback (most recent call last):
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/bin/sf-daemon", line 8, in <module>
Dec 29 12:15:57 sf-3 sh[9194]:     sys.exit(main())
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 182, in main
Dec 29 12:15:57 sf-3 sh[9194]:     _audit_daemons()
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 162, in _audit_daemons
Dec 29 12:15:57 sf-3 sh[9194]:     _start_daemon(d)
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/main.py", line 103, in _start_daemon
Dec 29 12:15:57 sf-3 sh[9194]:     DAEMON_IMPLEMENTATIONS[d].Monitor(d).run()
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 198, in run
Dec 29 12:15:57 sf-3 sh[9194]:     self._maintain_networks()
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/daemons/net.py", line 95, in _maintain_networks
Dec 29 12:15:57 sf-3 sh[9194]:     n.create_on_hypervisor()
Dec 29 12:15:57 sf-3 sh[9194]:   File "/usr/local/lib/python3.6/dist-packages/shakenfist/net.py", line 309, in create_on_hypervisor
Dec 29 12:15:57 sf-3 sh[9194]:     self._create_common()
[...]

PR #610, opened 2020-12-29
Minor things found whilst fiddling

mandoonandy:
* Avoid crashing daemon when VXLAN already exists. This should occur until Issue #609 is fixed.
* Minor instance property methods
* Start moving lock names to one location
* Avoid client errors return floating_gateway in net. This will return None initially since it has not been assigned but
subsequent requests give correct data. PR will be raised to request floating_gateway be assigned on instantiation, rather than network creation on network node.


shakenfist/virt.py
```diff
@@ -290,9 +291,11 @@ def block_devices(self):
         return self._db_get_attribute('block_devices')
 
     # Implementation
+    def get_lock(self, name, op):
```
mikalstill, line 294: Isn't this confusing? Its locking an attribute not the entire object?

mikalstill (commented):
One small question, and I've re-run CI which I suspect will pass now that ansible galaxy works again...

shakenfist/virt.py
mandoonandy: True. Let me modify that.

mikalstill (approved):
LGTM

shakenfist/virt.py
```diff
@@ -290,9 +291,11 @@ def block_devices(self):
         return self._db_get_attribute('block_devices')
 
     # Implementation
+    def get_lock_attr(self, name, op):
```
mikalstill, line 294: Sorry, brain fart. Shouldn't this be in baseobject.DatabaseBackedObject?

mikalstill (commented):
Sorry, thought of something.

shakenfist/virt.py
mandoonandy: Good point. The PR auto closed whilst I was messing with my personal develop branch. Will reopen.

PR #615, opened 2021-01-02
Minor things found whilst fiddling

mandoonandy:
* Avoid crashing daemon when VXLAN already exists. This should occur until Issue #609 is fixed.
* Minor instance property methods
* Start moving lock names to one location
* Avoid client errors return floating_gateway in net. This will return None initially since it has not been assigned but 
    subsequent requests give correct data. PR will be raised to request floating_gateway be assigned on instantiation, rather than network creation on network node.

The removal of hard coded attribute lock names uncovered a subtle bug at shakenfist/net.py#213

### DIFF
--- a/shakenfist/baseobject.py
+++ b/shakenfist/baseobject.py
@@ -28,14 +28,14 @@ class DatabaseBackedObject(object):
         return self.__version
 
     def __str__(self):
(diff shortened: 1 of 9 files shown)


mikalstill (approved):
LGTM

PR #627, opened 2021-01-02
State checking for object state updates

### LINKED ISSUE #608: Model and enforce valid state transitions in baseobject.py
This would be nice because it would be self documenting and catch errors when we introduce them.

mandoonandy:
Resolves #608

### DIFF
--- a/shakenfist/images.py
+++ b/shakenfist/images.py
@@ -31,6 +31,14 @@ resolvers = {
 class Image(baseobject.DatabaseBackedObject):
     object_type = 'image'
     current_version = 2
+    state_targets = {
+        None: ('initial', 'error'),
+        'initial': ('creating', 'error'),
+        'creating': ('created', 'deleted', 'error'),
+        'created': ('creating', 'deleted', 'error'),
+        'error': ('deleted', 'error'),
+        'deleted': ('error'),
+    }
 
     def __init__(self, static_values):
         # NOTE(mikal): we call the unique_ref the "uuid" for the rest of this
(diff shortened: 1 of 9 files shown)


shakenfist/images.py
```diff
@@ -68,7 +76,7 @@ def new(cls, url, checksum=None):
             'version': cls.current_version
         })
         i = Image.from_db(uuid)
-        i._db_set_attribute('state', {'state': 'initial'})
+        i._db_set_attribute('state', {'state': None})
         i.update_state('initial')
```
mikalstill, line 80: Why do this in two db writes? Do you see an advantage I'm missing?

mikalstill (commented):
Just one quick question.

shakenfist/images.py
mandoonandy: Minimal touching of the code. Just keeping this change to what was required. The change to that line was only the initial value.

I do agree that it is unnecessary. The related functions look like they will handle the missing DB entry.

mikalstill: Ok, I will merge this and then revisit.

mikalstill:
Take a look at the CI results. There is definitely something up with this change.

mandoonandy:
This has exposed a few issues... I will have to raise individual issues.

mikalstill:
Should this one be a draft until the "other issues" are resolved? Or should I re-run CI?

mandoonandy:
This did pass CI before my last comment. It should pass again...

mandoonandy:
Failed due Internal Server Error in etcd during CI run.

mikalstill (approved):
LGTM

PR #640, opened 2021-01-07
Store object state in State class

mandoonandy:


mandoonandy:
I don't like `instance.state.state` but will think of something better during the next refactor.

shakenfist/baseobject.py
```diff
@@ -78,31 +78,29 @@ def get_lock_attr(self, name, op):
...
     @property
     def state(self):
-        return self._db_get_attribute('state')
+        db_data = self._db_get_attribute('state')
+        if not db_data:
+            return State(None, 0)
+        return State(**db_data)
 
     def update_state(self, state, error_message=None):
+        LOG.withField('state', state).error('update_state()')
```
mikalstill, line 87: ?

shakenfist/daemons/net.py
```diff
@@ -66,8 +66,7 @@ def _maintain_networks(self):
                     if not inst:
                         stray = True
                     else:
-                        dbstate = inst.state
-                        if dbstate.get('state') in ['deleted', 'error', 'unknown']:
+                        if inst.state.state in ['deleted', 'error', 'unknown']:
```
mikalstill, line 69: You're right, inst.state.state is a bit weird.

mikalstill (commented):
Should the state member variable be called something like value so we write inst.state.value?

shakenfist/baseobject.py
mandoonandy: who put that there!?

shakenfist/daemons/net.py
mandoonandy: Weirdness reduced

shakenfist/baseobject.py
```diff
@@ -78,31 +78,28 @@ def get_lock_attr(self, name, op):
...
 
-            dbstate['state'] = state
-            dbstate['state_updated'] = time.time()
-
-            if error_message:
-                dbstate['error_message'] = error_message
-
-            self._db_set_attribute('state', dbstate)
-            self.add_event('state changed', '%s -> %s' % (orig_state, state))
+            new = State(state, time.time(), error_message)
```
mikalstill, line 96: I feel like new isn't the best name for a variable ever, even if its not a reserved word in python.

mikalstill (changes requested):
Shouldn't there be some value renaming in external_view() implementations so that we're not chaing the JSON schema for the API? I can't see that in here.

mandoonandy:
> Shouldn't there be some value renaming in external_view() implementations so that we're not chaing the JSON schema for the API? I can't see that in here.

This has been done with `state` moved from the `attrname` loop in each object.

I think external_view() could be improved. Probably move some items to the baseobject.

shakenfist/baseobject.py
mandoonandy: Improved.

shakenfist/baseobject.py
```diff
@@ -111,11 +109,45 @@ def state_filter(states, o):
...
+
+    @property
+    def update_time(self):
+        return self.__update_time
+
+    @property
+    def error_msg(self):
+        return self.__error_msg
+
+    def json_dump(self):
```
mikalstill, line 148: Sorry to be _that_guy_, but this isn't JSON right? Its a dictionary you then turn into JSON. I think this should probably be `__dict__`, but I am not 100% sure if that's the intended use of that call or not... Its not completely clear from the exactly one web page I read about it.

mikalstill (changes requested):
I see the exernal_view fix now, thanks.

shakenfist/baseobject.py
mandoonandy: lol. that_guy.

(I've never tried redefining __dict__() but I fear it will not end well.)

mandoonandy: Renamed all json_dump()

mikalstill (approved):
LGTM

PR #641, opened 2021-01-07
Do not process queue workitems for dead networks

mandoonandy:


shakenfist/daemons/net.py
```diff
@@ -148,16 +148,31 @@ def _process_network_node_workitems(self):
...
+
+            # Tasks valid for a network in any state
+            if isinstance(workitem, RemoveDHCPNetworkTask):
+                n.remove_dhcp()
+                db.add_event('network', workitem.network_uuid(),
+                             'network node', 'remove dhcp', None, None)
+                return
+
+            # Tasks that should not operate on a dead network
+            if n.is_dead():
```
mikalstill, line 170: Shouldn't we use the deleted and error states here? What is the added value of using is_dead() here?

mandoonandy: `is_dead()` includes the `deleting` state.

 That was the solution from a month or so ago to capture the in-flight delete case.

mikalstill: Oh wait. I see now. I feel like is_ok(), is_created(), and is_dead() should probably move down to baseobject.py. There are lots of other places in the code which could use those calls (I'm mostly thinking instances here).

mikalstill (approved):
LGTM.

shakenfist/daemons/net.py
mandoonandy: I agree. I would like to see all the state comparisons to strings moved to `is_*` functions.

PR #642, opened 2021-01-07
Refactor State and error message

mandoonandy:


mandoonandy:
Check my changes to shakenfist/tests/test_daemon_cleaner.py. I think the original unit test was wrong, but make sure I haven't done anything crazy.

mandoonandy:
One of the causes of CI flakiness for this PR is late reporting of instance error state. Raising an issue.

shakenfist/baseobject.py
```diff
@@ -98,6 +99,21 @@ def update_state(self, new_value, error_message=None):
...
+            return None
+        return db_data
+
+    @error_msg.setter
+    def error_msg(self, msg):
+        if self.state.state != 'error':
+            raise exceptions.InvalidStateException(
+                'Object not in error state (state=%s, object=%s)' % (
+                    self.state, self.object_type))
+        self._db_set_attribute('error_msg', msg)
```
mikalstill, line 117: I'm not sure how I feel about saving the characters in the database schema, I feel a bit like we'd be better off calling this "error_message", or just "error". It is a bit nit picky though.

shakenfist/daemons/queues.py
```diff
@@ -265,7 +265,9 @@ def instance_delete(instance_uuid, new_state, error_msg=None):
         instance = virt.Instance.from_db(instance_uuid)
         if instance:
             instance.delete()
-            instance.update_state(new_state, error_message=error_msg)
+            instance.update_state(new_state)
+            if error_msg:
+                instance.error_msg = error_msg
```
mikalstill: If update_state() set the error message for error states, this would not need to change. In fact we don't do it now, but should entering a state other than error ensure that the error message in the database is blank?

shakenfist/baseobject.py
```diff
@@ -93,11 +95,27 @@ def update_state(self, new_value, error_message=None):
...
+        return db_data
+
+    @error_msg.setter
+    def error_msg(self, msg):
+        s = self.state
+        if s.value != 'error':
+            raise exceptions.InvalidStateException(
+                'Object not in error state (state=%s, object=%s)' % (
+                    s, self.object_type))
+        self._db_set_attribute('error_msg', msg)
```
mikalstill, line 117: Additionally (this was more obvious once I read the upgrade), this would be the only example of storing just a string at a key. Should we use a dict here just in case we want to add other fields later?

mikalstill (commented):
Questions...

shakenfist/baseobject.py
mandoonandy: It felt wasteful to store a dict, but happy to change it to be consistent with other attributes.

shakenfist/baseobject.py
mandoonandy: Sure.

mikalstill (approved):
LGTM

PR #651, opened 2021-01-13
Set error state on instance immediately

### LINKED ISSUE #650: Late reporting of instance error state causes CI flakiness
When an instance experiences an error (eg. image URL is invalid), an error is queued. Due to the queue task system, the instance does not immediately go into the error state. The CI pipline and the sf-client is sometimes too quick and finds the instance not in error.

The error state and message for the instance is set when the queue task is executed. In fact, the error state (and msg) is set after all the instance delete actions (including network deletion).

Worse. When a bad image URL causes the error, the error msg is set to the original HTTP exception. Then the image queue task sets a different (prettier) message.

The API return value can be one of at least three different states. The sf-client also is able to show the same problem.

mandoonandy:
* Set error state on instance immediately
* Send correct error code when network doesn't exist

Resolves #650

### DIFF
--- a/shakenfist/client/upgrade.py
+++ b/shakenfist/client/upgrade.py
@@ -5,9 +5,10 @@ import ipaddress
 import json
 import time
 
+from shakenfist import baseobject
 from shakenfist import db
 from shakenfist import util
-from shakenfist import baseobject
+from shakenfist import virt
 
 # Very simple data upgrader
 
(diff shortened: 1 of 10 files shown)


mandoonandy:
This PR is branched from PR #642. I'll rebase on `develop` once PR #642 is merged.

mikalstill (approved):
LGTM

PR #654, opened 2021-01-13
Lock networks to prevent deletion during creation

### LINKED ISSUE #647: Networks can be deleted during creation
CI regularly has random failures due to an API network `DELETE` call before the creation of the same network is completed.

```
shakenfist.exceptions.InvalidStateException: ('Invalid state change from %s to %s for object=%s uuid=%s', 'deleted', 'created', 'network', 'ae1d7760-c64d-47fe-a55f-6a7bf69f9588')
```
There are no locks around the creation or the deletion of networks. Therefore the API can delete the network whilst the `net daemon` is creating the network.

It also appears that vxlan bridge and interface are only deleted on the node that receives the API call.

mandoonandy:
* Lock networks to prevent deletion during creation
* Only restore networks that are not dead
* Move get_object_lock into DatabaseBackedObject

Resolves #647

### DIFF
--- a/shakenfist/baseobject.py
+++ b/shakenfist/baseobject.py
@@ -72,6 +72,12 @@ class DatabaseBackedObject(object):
     def _db_set_attribute(self, attribute, value):
         etcd.put('attribute/%s' % self.object_type, self.__uuid, attribute, value)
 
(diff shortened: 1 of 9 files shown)


mikalstill (approved):
LGTM

PR #668, opened 2021-01-25
Remove stray persist in floating network creation.

### LINKED ISSUE #667: Stray persist() call in floating network setup in released version 0.4.0
This is a bit sad. I think we should push a release to correct this.

mikalstill:
Fixes #667.


mandoonandy (approved):
LGTM

PR #672, opened 2021-01-28
Convert nodes to real objects

mikalstill:


shakenfist/net.py
```diff
@@ -572,42 +574,44 @@ def ensure_mesh(self):
             # as an overlay cloud...
             node_ips = [config.NETWORK_NODE_IP]
             for fqdn in node_fqdns:
-                ip = db.get_node(fqdn)['ip']
-                if ip not in node_ips:
-                    node_ips.append(ip)
+                n = Node.from_db(fqdn)
+                if n:
+                    ip = n.ip
+                    if ip not in node_ips:
```
mandoonandy, line 580: Maybe use a set here?

shakenfist/node.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+LOG, _ = logutil.setup(__name__)
+
+
+class Node(baseobject.DatabaseBackedObject):
+    object_type = 'node'
+    current_version = 2
+    state_targets = {
+        None: ('created', 'error', 'missing'),
```
mandoonandy, line 17: state_targets should use the defined constants

shakenfist/node.py
```diff
@@ -0,0 +1,112 @@
...
+    # Values routed to attributes, writes are via helper methods.
+    @property
+    def last_seen(self):
+        return self._db_get_attribute('observed')['at']
+
+    @property
+    def installed_version(self):
+        return self._db_get_attribute('observed')['release']
+
+    def observe(self):
```
mandoonandy, line 89: What do you think about making this a static method. It could call new() itself.

This would remove the need for the "upsert" explanations in other parts of the code.

Maybe also change the method name to past tense?  `observed`

mandoonandy (changes requested):

shakenfist/node.py
mikalstill: Good catch, done.

shakenfist/node.py
mikalstill: So, I tweaked this along those lines, but I've ended up calling it observe_this_node() because it will only work for the node the code is running on.

shakenfist/net.py
mikalstill: Done.

mandoonandy (approved):
LGTM

PR #680, opened 2021-02-03
Avoid races in floating IPs test.

mikalstill:


deploy/shakenfist_ci/tests/test_floating_ips.py
```diff
@@ -63,6 +63,17 @@ def test_simple(self):
...
+        attempts = 0
+        for _ in range(6):
+            try:
+                r = requests.request(
+                    'GET', 'http://%s/' % ifaces[0]['floating'])
+                self.assertEqual(200, r.status_code)
+                self.assertEqual('Floating IPs work!\n', r.text)
+
+            except Exception as e:
+                if attempts < 5:
```
mandoonandy, line 76: probably should increment `attempts`

mikalstill: Y U so hater? Also yes.

mandoonandy: Sorry to be a pain, but why it does it test 6 times regardless of the outcome?

mikalstill: Because I am bad at computers.

mandoonandy (approved):
LGTM

PR #694, opened 2021-02-12
Fix GCP-XL, but not quite perfectly.

mikalstill:
This is mostly lifted from https://github.com/shakenfist/shakenfist/pull/424
(thanks Jack!), but is just the bits I need to get GCP-XL working
again for the nightly CI pipeline.


mandoonandy (approved):
LGTM

Issue #213, opened 2020-07-24
Unit Test Coverage - Core


mandoonandy:
Presumably the next commit will only partially cover the list below. Create a new issue for that commit and update this issue.
```
Name                                  Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------
shakenfist/config.py                     25      5      4      1    79%
shakenfist/daemons/cleaner.py            93     42     28      2    55%
shakenfist/images.py                    135     30     54      6    74%
shakenfist/ipmanager.py                  57     12     18      3    72%
shakenfist/net.py                       222    171     70      0    17%
shakenfist/scheduler.py                 151    128     88      0    10%
shakenfist/util.py                       97     35     32      0    64%
shakenfist/virt.py                      292    187     72      5    32%
```

mikalstill:
Noting that one issue with these reports is that if there is _zero_ coverage of a file, its not even listed in the report...

mikalstill:
```
Name                                  Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------
shakenfist/config.py                     50      0      0      0   100%
shakenfist/daemons/cleaner.py           129     70     50      5    44%
shakenfist/images.py                    280     47    105     16    79%
shakenfist/ipmanager.py                  70     25     18      3    59%
shakenfist/net.py                       396    268    124      3    27%
shakenfist/scheduler.py                 188      5    104     13    94%
shakenfist/util.py                      182     62     70      5    64%
shakenfist/virt.py                      490    270    126      7    40%
-----------------------------------------------------------------------
TOTAL                                  3968   1591   1279    135    55%
```

mikalstill:
```
Name                                  Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------
shakenfist/config.py                     50      0      0      0   100%
shakenfist/daemons/cleaner.py           129     70     50      5    44%
shakenfist/images.py                    280     47    105     16    79%
shakenfist/ipmanager.py                  72      1     18      0    99%
shakenfist/net.py                       396    268    124      3    27%
shakenfist/scheduler.py                 188      5    104     13    94%
shakenfist/util.py                      182     62     70      5    64%
shakenfist/virt.py                      490    270    126      7    40%
-----------------------------------------------------------------------
TOTAL                                  3972   1566   1279    132    56%
```

mikalstill:
[...]

PR #695, opened 2021-02-12
Improve unit test coverage of core modules.

mikalstill:
Improves #213 without actually fixing it.


mandoonandy:
Does this mean that address allocation was broken? In particular, `broadcast_address` and `network_address` could be allocated randomly since they were stored as IPV4Address but addresses frrom `get_random_address()` are strings (and stored as strings).

mandoonandy (approved):
LGTM

PR #702, opened 2021-02-18
A new installer.

mikalstill:
I've realised while doing the glusterfs work that the current installation
system is broken. Specifically, we currently use the configuration files and
ansible plays from the most recent version for all installs, which means that
when someone installs an older more stable version they can end up out of
alignment. This is especially obvious with the gluster work because we need
to tweak the libvirt template.

So instead, we now package the configuration samples and installer into the
python package, and use those to do the install. This means we need to
install the pip packages _before_ we install, instead of as part of the
install like before.


mandoonandy (approved):
LGTM

PR #709, opened 2021-02-20
Fix nightly ci

mikalstill:


mikalstill:
I've reworked this into https://github.com/shakenfist/shakenfist/pull/719, which is cleaner.

PR #710, opened 2021-02-28
Fix test versions

mikalstill:


deploy/shakenfist_ci/tests/test_floating_ips.py
```diff
@@ -65,20 +60,27 @@ def test_simple(self):
...
-                self.assertEqual(200, r.status_code)
-                self.assertEqual('Floating IPs work!\n', r.text)
-                return
-
-            except Exception as e:
-                if attempts < 5:
-                    pass
-                else:
-                    raise e
+                success = 200 == r.status_code
```
mandoonandy, line 74: So you left this here to see if I'm reviewing?

mandoonandy (changes requested):
One tiny python issue

deploy/shakenfist_ci/tests/test_floating_ips.py
mikalstill: No, that's deliberate. It stores the value of the comparison of 200 and r.status_code in a variable called success, which is checked in the next line. Or am I missing something?

mandoonandy: Would it be simpler to write:
```
if r.status_code == 200 and r.text.find('Floating IPs work!') != -1:
    return
```

mikalstill: He's a witch!

mikalstill: Also yes.

mandoonandy (approved):
LGTM

PR #712, opened 2021-03-01
Tweak floating IPs CI test logic.

mikalstill:
I found another weird bit of logic.


mandoonandy (approved):
LOL. That's the one I really should have found.

PR #738, opened 2021-03-27
Serve snapshots, don't crash when restarted with a disk with no base image

mikalstill:


mikalstill:
This one needed develop rebased...

PR #760, opened 2021-04-13
Provide usage statistics for virtual networks

mikalstill:
This is building up to an API which provides usage statistics for virtual networks. This is interesting to me as I try to work out what a CI run costs me in terms of bytes transferred off-net, and how to optimize that.


mikalstill:
Oh, wrong target branch.

PR #766, opened 2021-04-15
Add an image resolver for Debian.

### LINKED ISSUE #740: Add an image resolver for debian
Just like ubuntu.

mikalstill:
Fixes #740.

### DIFF
--- a/shakenfist/image_resolver_cirros.py
+++ b/shakenfist/image_resolver_cirros.py
@@ -1,3 +1,4 @@
+import os
 import re
 import requests
 
@@ -13,6 +14,7 @@ LOG, _ = logutil.setup(__name__)
 
 
 def resolve(name):
+    # Name is assumed to be in the form cirros or cirros:0.4.0
     resp = requests.get(CIRROS_URL,
                         headers={'User-Agent': util.get_user_agent()})
     if resp.status_code != 200:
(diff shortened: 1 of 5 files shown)


mikalstill:
CI still doesn't work, so ignore that failure please.

shakenfist/image_resolver_cirros.py
```diff
@@ -31,36 +33,16 @@ def resolve(name):
...
-            break
-    if not checksum_url:
-        log.warning('Did not find checksum')
-
-    log.with_field('checksum', checksum).info('Checksum retrieval')
-
+    checksums = util.fetch_remote_checksum(checksum_url)
+    checksum = checksums.get(os.path.basename(url))
+    LOG.with_field('name', name).with_field('url', url).with_field(
+        'checksum', checksum).info('Image resolved')
```
mandoonandy, line 47: `with_fields()` would reduce verbosity here.

shakenfist/image_resolver_debian.py
```diff
@@ -0,0 +1,54 @@
...
+        checksum_url = CHECKSUM_URL % {'version': vernum}
+        resp = requests.head(url,
+                             headers={'User-Agent': util.get_user_agent()})
+        if resp.status_code != 200:
+            raise exceptions.HTTPError(
+                'Failed to fetch %s, status code %d' % (url, resp.status_code))
+
+    else:
+        found_any = False
+        for vernum in range(9, 20):
```
mandoonandy, line 36: Would there be less lines of code if you searched backwards and stopped when a valid URL was found?

mandoonandy (changes requested):
One question (probably misunderstanding the code) and one suggestion.

shakenfist/image_resolver_debian.py
mikalstill: So, Debian only has 9 and 10 online at the moment. 10 is the current stable. So this is more future proofing than anything else. I guess I could load the index page and parse, but I was being lazy.

shakenfist/image_resolver_cirros.py
mikalstill: Done.

mandoonandy (approved):
LGTM

PR #769, opened 2021-04-15
Correct interface order, and include all networks in metadata.

mikalstill:


mikalstill:
With this patch I end up with a correctly configured multi-nic centos.

mandoonandy (approved):
LGTM

PR #773, opened 2021-04-19
Changes to support CI for client-python

mikalstill:


mikalstill:
Self merge as its infrastructural.

PR #781, opened 2021-05-02
Refactor image resolvers, and minor fixes found while adding the CentOS resolver

mikalstill:


shakenfist/baseobject.py
```diff
@@ -100,7 +99,18 @@ def state(self, new_value):
...
+                return
+
             # Ensure state change is valid
+            if not self.state_targets:
+                raise exceptions.NoStateTransitionsDefined(
+                    self.object_type)
+
+            LOG.with_field('original state', orig).with_field(
+                'new state', new_value).with_field(
+                    'targets', self.state_targets).debug('Verifying state transition')
```
mandoonandy, line 112: This could be less verbose using `with_fields()`?

shakenfist/image_resolver/debian.py
```diff
@@ -44,14 +41,18 @@ def resolve(name):
...
         'name': name,
         'url': url,
         'checksum': checksum
     }).info('Image resolved')
-    return (url, checksum)
+
+    if checksum:
+        return (url, checksum)
+    else:
+        return (url, None)
```
mandoonandy, line 58: Would this be better here?
```
return (url, checksum if checksum else None)
```

shakenfist/image_resolver/ubuntu.py
```diff
@@ -56,12 +52,16 @@ def resolve(name):
...
         'name': name,
         'url': url,
         'checksum': checksum
     }).info('Image resolved')
-    return (url, checksum)
+
+    if checksum:
+        return (url, checksum)
+    else:
+        return (url, None)
```
mandoonandy, line 67: as above, using `if`

shakenfist/images.py
```diff
@@ -63,8 +56,8 @@ def __init__(self, static_values):
 
     @classmethod
     def new(cls, url, checksum=None):
-        # Handle URL shortcut with built-in resolvers
-        url, resolver_checksum = Image._resolve(url)
+        # Handle URL shortcut with built-in , '
```
mandoonandy, line 59: missed enter, hit `'`?

mandoonandy (changes requested):
Couple of tiny comments to prove I looked at the code.

shakenfist/image_resolver/debian.py
mikalstill: Done.

shakenfist/images.py
mikalstill: Great moments in search and replace.

mandoonandy (approved):
LGTM

PR #793, opened 2021-05-10
Convert shared states to single constant.

mikalstill:


mandoonandy (approved):
LGTM

Love constants!

PR #795, opened 2021-05-13
Pin werkzeug and flask as the 2.0.0 releases have broken us.

mikalstill:


mikalstill:
CI is broken because of another issue, so we get to fly blind here for a little.

PR #800, opened 2021-05-31
Attempt to even out CPU load by stepping up to the max CPU overcommit.

mikalstill:


shakenfist/scheduler.py
```diff
@@ -239,11 +247,33 @@ def place_instance(self, instance, network, candidates=None):
...
+            actual_cluster_cpu = 0
+            for node in self.metrics:
+                current_cluster_cpu += self.metrics[node].get(
+                    'cpu_total_instance_vcpus', 0)
+                actual_cluster_cpu += self.metrics[node].get(
+                    'cpu_available', 0)
+            cpu_ratio = min(config.get('CPU_OVERCOMMIT_RATIO'),
+                            math.ceil(current_cluster_cpu / actual_cluster_cpu))
+            log_ctx.info('Current cluster CPU overcommit ratio is %d, configured maximum %d'
+                         % (cpu_ratio, config.get('CPU_OVERCOMMIT_RATIO')))
```
mandoonandy, line 266: It's better to leave the string interpolation to the logging function.

mandoonandy (changes requested):
Found a non-bug to comment on since I don't understand the substantive code.

mikalstill:
Rage merging because CI is unreliable with an unrelated failure.

PR #826, opened 2021-07-16
Image downloads are blobs.

mikalstill:
A step in the direction of replacing the current image implementation with artifacts and blobs.

Defunct tests removed. I will add new tests when the churn has settled down. There are still extensive CI tests.


mandoonandy (approved):
LGTM

PR #841, opened 2021-08-03
Remove flat disks.

### LINKED ISSUE #483: QCOW and Flat disk options
DISCUSSION
@mikal 
> I wonder if we need the qcow_flat and flat options at all any more?

I was hoping Mikal would ask this question.

Removing the two disk options further simplifies image management. We will not have any resized image layer files. This fits well with the opinionated nature of SF. This opinion is not without purpose. It removes code that will be rarely used thereby improving maintainability.

Are there use cases for these disk options that are common enough to keep them?

### LINKED ISSUE #836: Flat image variants complicate the code and are untested. Should we remove them?

mikalstill:
This isn't tested enough, adds complexity, and violates our manifesto
of being "opinionated".

Fixes #836. Fixes #483.


mandoonandy (approved):
LGTM

PR #845, opened 2021-08-06
Extra logging to help chase down start failures.

mikalstill:


mikalstill:
Merging this to get the logging for failures, even though I haven't figured out what is going wrong yet.

PR #847, opened 2021-08-07
Check MTU of correct interface.

mikalstill:


mikalstill:
Relevant part of CI passed.

PR #861, opened 2021-08-13
Fix duplicate packets on the mesh network.

### LINKED ISSUE #859: Duplicate packets observed when pinging between VMs on a single hypervisor
It seems that this only happens when there is another VM on the network on a remote hypervisor.

mikalstill:
Fixes #859.


mikalstill:
This change passed manual testing in a CI cluster and doesn't have a covering CI test right now. Merging, and I'll add a CI test ASAP.

PR #873, opened 2021-08-17
Include SF version in process title.

mikalstill:
This change includes the currently running version string in the process title of the "main" process so that operators can quickly tell what's running.


mikalstill:
This is a real error -- `raise Exception('Code Error: Bad process name: %s' % name)` -- I wonder what a valid process name is?

mandoonandy (approved):
LGTM

PR #875, opened 2021-08-18
Reduce number of DB calls for instance listing

mandoonandy:


mikalstill:
Superceded by https://github.com/shakenfist/shakenfist/pull/880

PR #878, opened 2021-08-19
Refresh locks during blob transfers.

mikalstill:
Otherwise large blob transfers result in the lock expiring and then a stack trace when we try to release it.

### DIFF
--- a/shakenfist/images.py
+++ b/shakenfist/images.py
@@ -4,12 +4,14 @@ import random
 import re
 import requests
 import shutil
+import time
 import uuid
 
 from shakenfist.artifact import Artifact, BLOB_URL
 from shakenfist import blob
 from shakenfist.blob import Blob
 from shakenfist.config import config
+from shakenfist import db
 from shakenfist import exceptions
 from shakenfist import image_resolver
 from shakenfist import logutil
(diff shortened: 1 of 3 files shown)


shakenfist/images.py
```diff
@@ -29,7 +31,7 @@ def __init__(self, inst, url):
             {'url': self.url, 'artifact': self.__artifact.uuid})
 
     def get_image(self):
-        with self.__artifact.get_lock() as lock:
+        with self.__artifact.get_lock(ttl=300) as lock:
```
mandoonandy, line 34: Magic number alert. Maybe define as a constant at the top of the file?

shakenfist/images.py
```diff
@@ -141,9 +143,15 @@ def _blob_get(self, url):
                                               'User-Agent': util.get_user_agent()})
 
                 with open(blob_path + '.partial', 'wb') as f:
+                    last_refresh = 0
+
                     for chunk in r.iter_content(chunk_size=8192):
                         f.write(chunk)
 
+                        if time.time() - last_refresh > 5:
```
mandoonandy, line 151: Magic number alert.

mandoonandy (changes requested):

shakenfist/images.py
mikalstill: This is just a lock refresh, the timeout is copied from elsewhere. I'll fix it.

shakenfist/images.py
mikalstill: Done, sort of.

mandoonandy (approved):
LGTM

Thanks for the changes. The extra variable name is effectively an extra bit of documentation.

PR #880, opened 2021-08-19
Store network interfaces in instances.

mikalstill:
A counter proposal to https://github.com/shakenfist/shakenfist/pull/875 -- we know what interfaces an instance has because we create them before we create the instance, so let's just store them in the instance as well.


mikalstill:
CI failures are unrelated to this code, they're my flakey internet.

mandoonandy (approved):
I'll trust you.

PR #885, opened 2021-08-22
DRAFT: Wait for clean network before network deletion

### LINKED ISSUE #877: sf-client would appreciate a single endpoint to clean a namespace
Network deletion has to wait until all instances are deleted, also should only delete networks in namespace at command start not when instance delete finishes.
Fix could be sending a net del command to queue with some sort of delayed recheck thingy.

mandoonandy:
Closes #877 

No tests yet - draft for comment


shakenfist/daemons/queues.py
```diff
@@ -121,6 +123,15 @@ def handle(jobname, workitem):
                 snapshot(inst, task.disk(),
                          task.artifact_uuid(), task.blob_uuid())
 
+            elif isinstance(task, DeleteNetworkWhenClean):
+                remaining_inst = instance.instances_in_namespace()
```
mikalstill, line 127: I think this is probably not what we meant?

The list of instances to be deleted should be populated when the API call is made, and then reduced as things are deleted. When that list is empty the network is deleted.

If the network still has instances in it at that time, then that's because they were created after the namespace delete was requested, and the delete should in fact fail.

Thoughts?

mikalstill (changes requested):
I like the general thrust, one implementation tweak requested inline.

shakenfist/daemons/queues.py
mandoonandy: I agree with your description but since the network blocks further attachments the code should have the same effect. Of course, a bug could leave the task hanging forever. So plan A (as you describe) is probably the safer course.

Now that I look at the code I see that it is borked. It does not specify nor have access to the namespace name.

I will try harder.

mandoonandy: This has been harder than the initial concept indicated...

No tests yet, but want to get ahead of any major code changes to interfaces...

PR #887, opened 2021-08-23
Fix vscode deploy.yml display issue

mandoonandy:
Found solution in https://github.com/redhat-developer/vscode-yaml/issues/397


mikalstill:
Its a Christmas miracle!

PR #901, opened 2021-08-25
Add etcd_master as configurable in topology.

mikalstill:
This is needed so we can have clusters where not every node is an etcd master.


mikalstill:
I've hand verified that this sets up on sf-[123] as etcd masters, and [45] as proxies.

PR #923, opened 2021-08-29
Do not gather facts to speed up install.

mikalstill:


mikalstill:
Yeah, this doesn't work how I hoped it would.

PR #932, opened 2021-09-01
Ensure we don't attempt to create a dead network.

mikalstill:
I wouldn't think this would be all that common, but I can see at least
ten examples a day of us deleting a network while it is being created
and thefore attempting to transition from DELETE_WAITING to CREATED
state. Add a final check before the state change to catch this more
gracefully.


mandoonandy (approved):
LGTM

PR #938, opened 2021-09-04
Use create_interface() for floating IP veths.

mikalstill:
This means they'll pick up MTUs like other interfaces we manage.


mikalstill:
CI failure is unrelated.

PR #956, opened 2021-09-08
Convert the old image APIs into artifact APIs.

mikalstill:


mikalstill:
CI failure is unrelated.

PR #957, opened 2021-09-10
Disks should not be allowed to get smaller.

mikalstill:


mikalstill:
CI failure is unrelated.

PR #959, opened 2021-09-10
Pause before retrying network deletes.

### LINKED ISSUE #958: Delete_wait networks hard retry constantly
I see 260,000 retries in a cluster I am watching at the moment. When we queue a retry, we should add a delay.

mikalstill:
Add the concept of deferring a queued job, and then use that to add
a delay between retries for deleting a network which is still in
use. Fixes #958.

### DIFF
--- a/shakenfist/etcd.py
+++ b/shakenfist/etcd.py
@@ -286,10 +286,10 @@ def delete_all(objecttype, subtype):
     WrappedEtcdClient().delete_prefix(path)
 
 
-def enqueue(queuename, workitem):
+def enqueue(queuename, workitem, delay=0):
     with get_lock('queue', None, queuename, op='Enqueue'):
         i = 0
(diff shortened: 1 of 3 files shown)


shakenfist/etcd.py
```diff
@@ -286,10 +286,10 @@ def delete_all(objecttype, subtype):
     WrappedEtcdClient().delete_prefix(path)
 
 
-def enqueue(queuename, workitem):
+def enqueue(queuename, workitem, delay=0):
     with get_lock('queue', None, queuename, op='Enqueue'):
         i = 0
-        entry_time = time.time()
+        entry_time = time.time() + 0
```
mikalstill, line 292: Shit. This is wrong. It should be "+ delay". I blame tired.

mandoonandy (approved):
LGTM

PR #961, opened 2021-09-10
Treat delete_wait as an active state.

### LINKED ISSUE #960: Networks in the new delete_wait state are not returned by the API
They should be, as they still exist.

mikalstill:
Fixes #960.


mandoonandy:
This is probably a saner approach.

Issue #655, opened 2021-01-13
We should have a get_image() API call


mikalstill:
Which returns the details of a single image. What nodes its on. Who is using it. etc etc. I think this should probably wait until we've done the image rewrite though.

mikalstill:
I think artifact show now meets this usecase.

PR #963, opened 2021-09-11
Add an /artifacts/<uuid> API endpoint.

mikalstill:
This fixes the server side of #655. There is client side work required
as well.

### DIFF
--- a/shakenfist/daemons/queues.py
+++ b/shakenfist/daemons/queues.py
@@ -203,12 +203,13 @@ def handle(jobname, workitem):
 
 
 def image_fetch(url, inst):
+    a = Artifact.from_url(Artifact.TYPE_IMAGE, url)
     try:
         # TODO(andy): Wait up to 15 mins for another queue process to download
         # the required image. This will be changed to queue on a
         # "waiting_image_fetch" queue but this works now.
         images.ImageFetchHelper(inst, url).get_image()
-        db.add_event('image', url, 'fetch', None, None, 'success')
+        a.add_event('fetch', None, None, 'success')
 
     except (exceptions.HTTPError, requests.exceptions.RequestException) as e:
         LOG.with_field('image', url).info('Failed to fetch image')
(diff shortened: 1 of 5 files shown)


shakenfist/daemons/queues.py
```diff
@@ -381,8 +382,7 @@ def instance_delete(inst):
 def snapshot(inst, disk, artifact_uuid, blob_uuid):
     blob.snapshot_disk(disk, blob_uuid)
     a = Artifact.from_db(artifact_uuid)
-    if a.state == dbo.STATE_INITIAL:
-        a.state = dbo.STATE_CREATED
+    a.state = dbo.STATE_CREATED
```
mandoonandy, line 385: Would it be better to reference the class of the object? ie. not `dbo` but `Artifact.STATE_CREATED`.

It seems nit picky but when I've been searching for state assignments the code currently has both usages. One should be chosen and I would argue for attribute of the actual class.

mandoonandy (changes requested):
Just one controversial request. Controversial because it will flow on to touch lots of lines of code.

shakenfist/daemons/queues.py
mikalstill: I think most places use the base object version. They do that because it felt more natural to use the generic form of the generic states. Given these are actually just strings in the code I don't think there's a way to enforce which you use.

What's the advantage of using the derived version?

mandoonandy: The advantage is that assignments of state for Artefact can easily be found. They are can be visually verified as correct.

Enforcement would be difficult. The solutions I can think of would require extra code when converting to/from JSON. We have that already for other cases but not ideal for such a minor benefit.

(Should we swap to golang?)

mandoonandy (approved):
LGTM

shakenfist/daemons/queues.py
mikalstill: I've filed https://github.com/shakenfist/shakenfist/issues/974 to track this.

PR #967, opened 2021-09-12
Disable logrotate in CI so we don't get truncated logs while debugging.

### LINKED ISSUE #966: Disable logrotate in CI, it causes truncated logs while trying to debug.
https://sfcbr.shakenfist.com/jenkins/job/shakenfist_pr_ubuntu_2004/view/change-requests/job/PR-965/2/execution/node/44/log/?consoleFull is an example of a truncated log. I've never seen this before, I guess I'm just lucky.

mikalstill:
Fixes #966.

### DIFF
--- a/deploy/ansible/ci.yml
+++ b/deploy/ansible/ci.yml
@@ -237,6 +237,18 @@
     source_path: "/home/jenkins/src/shakenfist/"
 
   tasks:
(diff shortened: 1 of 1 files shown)


mikalstill:
Relevant parts of CI have passed, and the run will ultimately fail without #965.

PR #972, opened 2021-09-13
Add network object state documentation

mandoonandy:
Initial pass. More details would nice, but when I have more time.


mikalstill:
Thanks.

PR #1004, opened 2021-09-26
Add a single node CI job as well.

mikalstill:


mikalstill:
The voting on this PR is confusing because of the change in job configuration. I'm going to close / reopen to make that clearer.

PR #1020, opened 2021-10-03
Fix boot from NVMe.

mikalstill:


mikalstill:
CI failures are unrelated.

PR #1036, opened 2021-10-09
Lay the groundwork for nextgen events.

mikalstill:
I think we should be moving events to a scheme where they're queued through etcd, but not stored there. This would massively reduce our key count in etcd. Additionally, I see events and logs as increasingly overlapping. We should handle an event as a special case of a log message instead of a completely separate construct.


shakenfist/logutil.py
```diff
@@ -141,8 +161,44 @@ def process(self, msg, kwargs):
...
+
+            encoded = json.dumps(
+                EventLogMessageTask(time.time(), msg, extra),
+                indent=4, sort_keys=True, cls=JSONEncoderTasks)
+            Etcd3Client().put(
+                '/sf/queues/eventlog/%s/%s-%s'
+                % (config.NODE_NAME, time.time(), util_random.random_id),
+                encoded, lease=None)
+        except Exception:
+            # Don't let event errors break logging
```
mandoonandy, line 199: Silent exceptions lead to not so silent pain. Any exceptions here need to surface somehow.

mandoonandy (commented):
What's the overall intention here? Maybe a short design description is in order?

shakenfist/logutil.py
mikalstill: My issue is that an exception here can cause logging to stop working entirely, which would stop you from seeing other exceptions...

mikalstill:
Oh, this is against the wrong target branch of course.

PR #1038, opened 2021-10-10
Use random ids for queue items, instead of sequential counters.

mikalstill:
This should reduce etcd load when adding a queue item.


mandoonandy (approved):
LGTM

PR #1046, opened 2021-10-13
Handle graceful shutdowns and allow persistent config overrides.

mikalstill:
Allow deployers to provide persistent configuration overrides per node, and provide a mechanism to gracefully reload Shaken Fist when the configuration changes.


shakenfist/daemons/daemon.py
```diff
@@ -48,6 +49,13 @@ def __init__(self, name):
         log, _ = logutil.setup(name)
         set_log_level(log, name)
 
+        self.running = True
+        signal.signal(signal.SIGTERM, self.exit_gracefully)
+
+    def exit_gracefully(self):
```
mandoonandy, line 55: `The handler is called with two arguments: the signal number and the current stack frame `

mikalstill: Oh cool. Thanks!

mikalstill:
Doh. Wrong target branch yet again.

PR #1047, opened 2021-10-14
Reference count blobs and allow artifact deletion

mandoonandy:
Request for comments. Test to be added once design settled.

### DIFF
--- a/deploy/shakenfist_ci/tests/test_artifacts.py
+++ b/deploy/shakenfist_ci/tests/test_artifacts.py
@@ -83,3 +83,76 @@ class TestImages(base.BaseNamespacedTestCase):
             inst = self.test_client.get_instance(inst['uuid'])
 
         self.assertTrue(inst['state'] in ['creating-error', 'error'])
+
+    def test_artifact_ref_count_label(self):
+        url = ('https://sfcbr.shakenfist.com/gw-basic/gwbasic.qcow2')
+
+        img = self.system_client.cache_artifact(url)
+
+        # Get all artifacts once to make sure we get added to the list
+        image_urls = []
+        for image in self.system_client.get_artifacts():
+            image_urls.append(image['source_url'])
+        self.assertIn(url, image_urls)
+
+        # And then just lookup the single artifact
+        start_time = time.time()
+        found = False
+        while time.time() - start_time < 7 * 60:
+            img = self.system_client.get_artifact(img['uuid'])
+            if img['state'] == 'created':
+                found = True
+            time.sleep(5)
+
+        if not found:
+            self.fail('Image was not downloaded after seven minutes: %s' % (
+                img['uuid']))
+
+        self.assertIn('blobs', img)
+        self.assertEqual(1, len(img['blobs']))
+        self.assertIn('1', img['blobs'])
+        self.assertIn('reference_count', img['blobs']['1'])
+        self.assertEqual(1, img['blobs']['1']['reference_count'])
+
+        self.assertIn('blob_uuid', img)
+        blob_uuid = img['blob_uuid']
+
+        # Create a label artifact pointing at the blob
+        label_name1 = 'test_label_01'
+        lbl = self.test_client.update_label(label_name1, blob_uuid)
+        self.assertIn('blobs', lbl)
+        self.assertEqual(1, len(lbl['blobs']))
+        self.assertIn('1', lbl['blobs'])
+        self.assertIn('reference_count', lbl['blobs']['1'])
+        self.assertEqual(2, lbl['blobs']['1']['reference_count'])
+
+        # Create second label also pointing at the blob
+        label_name2 = 'test_label_02'
+        lbl2 = self.test_client.update_label(label_name2, blob_uuid)
+        self.assertIn('blobs', lbl2)
+        self.assertEqual(3, lbl['blobs']['1']['reference_count'])
+
+        # Delete the first label
+        self.assertIn('uuid', lbl)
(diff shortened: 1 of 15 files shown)


shakenfist/blob.py
```diff
@@ -97,10 +98,22 @@ def locations(self):
...
+    def locations(self, new_locations):
+        self._db_set_attribute('locations', {'locations': new_locations})
+
     @property
     def info(self):
         return self._db_get_attribute('info')
 
+    @property
+    def ref_count(self):
+        '''Counts artifact references to the blob'''
```
mikalstill, line 111: Just betweent the two of us, PEP8 says this should be triple double quotes, so:

`"""Counts the things."""`

shakenfist/blob.py
```diff
@@ -137,6 +166,32 @@ def observe(self):
...
+                raise BlobDeleted
+            new_count = self.ref_count + 1
+            self._db_set_attribute('ref_count', {'ref_count': new_count})
+            return new_count
+
+    def ref_count_dec(self):
+        with self.get_lock_attr('ref_count', 'Increase ref count'):
+            new_count = self.ref_count - 1
+            if new_count < 0:
+                raise BlobRefCountDecrementBelowZero
```
mikalstill, line 181: I'm actually not sure if we should fail here. I guess we should so we notice a bug, but its weird that my instance delete would fail entirely because of a reference counting error.

shakenfist/external_api/artifact.py
```diff
@@ -48,6 +48,41 @@ class ArtifactEndpoint(api_base.Resource):
...
+
+        Artifacts can only be deleted from the system if they are not in use.
+        The actual deletion of the on-disk files is left to the cleaner daemon.
+
+        It is acknowledged that there is a potential race condition between the
+        check that an artifact is not in use and the marking of the artifact as
+        deleted. This is only caused by a user simultaneously deleting an
+        artifact and attempting to start a VM using it. It is recommended that
+        the user does not do that.
+        '''
```
mikalstill, line 63: The quote thing again.

mikalstill: Also, nice passive aggresive comment. That's a keeper.

shakenfist/external_api/instance.py
```diff
@@ -135,19 +136,35 @@ def post(self, name=None, cpus=None, memory=None, network=None, disk=None,
...
+                    return api_base.error(404, 'label %s is not ready (state=%s)'
+                                          % (label, a.state.value))
                 blob_uuid = a.most_recent_index.get('blob_uuid')
                 if not blob_uuid:
                     return api_base.error(404, 'label %s not found (no versions)' % label)
+                b = Blob.from_db(blob_uuid)
+                if not b:
+                    return api_base.error(404, 'artifact references non-existent blob (%s)' % blob_uuid)
+                if b.state == Blob.STATE_DELETED:
+                    return api_base.error(404, 'artifact references deleted blob (%s)' % blob_uuid)
```
mikalstill, line 149: In general elsewhere we don't tell them _why_ the object isn't found. Just that its not found. I guess I was thinking that we shouldn't confirm that the object ever existed? Thoughts? Maybe my concern applies more to the "incorrect permissions" case.

mikalstill (commented):
Generally pretty sane. Some small nits.

shakenfist/blob.py
mandoonandy: The exception would be alerting that a major bug has occurred.

The only way to get a negative value is to start at zero. If the counter is at zero then the state will (should) be DELETED. If not state deleted, yet another bug has been uncovered.

shakenfist/external_api/artifact.py
mandoonandy: The alternative was a major code change in instance create to get and hold locks during instance creation.

shakenfist/external_api/instance.py
mandoonandy: I agree in the security case. In this case, I'm thinking that the user might have a blob UUID and attempt to use it after deletion. This msg explains why a previously valid blob UUID does not work.

Being UUID's, it's unlikely they will guess a foreign UUID. It also does not reveal any information about the blob.

When Blobs are namespace'd, the security check should be before this message.

shakenfist/artifact.py
```diff
@@ -141,6 +144,7 @@ def external_view(self):
             blobs[blob_index['index']] = {
                 'instances': b.instances,
                 'size': b.size,
+                'ref_count': b.ref_count,
```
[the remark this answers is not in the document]
mandoonandy: Changed. (I felt `reference` is a little verbose in the context.)

deploy/shakenfist_ci/tests/test_artifacts.py
```diff
@@ -83,3 +83,76 @@ def test_resize_image_to_small(self):
...
+
+        if not found:
+            self.fail('Image was not downloaded after seven minutes: %s' % (
+                img['uuid']))
+
+        self.assertIn('blobs', img)
+        self.assertEqual(len(img['blobs'], 1))
+        self.assertIn('1', img['blobs'])
+        self.assertIn('reference_count', img['blobs']['1'])
+        self.assertEqual(img['blobs']['1']['reference_count'], 1)
```
mikalstill, line 115: Totally a nit, but in general asserts are written as (expected value, actual value) so they match the error message when they fail. That is, the "1" here should be the first argument.

shakenfist/blob.py
mikalstill: Fair.

PR #1054, opened 2021-10-15
Add missing sata bus support.

mikalstill:


mikalstill:
Doh, wrong target branch.

PR #1058, opened 2021-10-15
Delete versions within snapshots

mandoonandy:
Delete versions within snapshots

Second commit that only standardises artifact delete endpoint.

Depends on shakenfist/client-python#116

### DIFF
--- a/shakenfist/artifact.py
+++ b/shakenfist/artifact.py
@@ -123,6 +123,9 @@ class Artifact(dbo):
             self.log.with_fields(entry).info('Added index to artifact')
             return entry
 
+    def del_index(self, index):
+        self._db_delete_attribute('index_%012d' % index)
+
     def external_view_without_index(self):
         return {
             'uuid': self.uuid,
--- a/deploy/shakenfist_ci/tests/test_artifacts.py
+++ b/deploy/shakenfist_ci/tests/test_artifacts.py
@@ -130,7 +130,7 @@ class TestImages(base.BaseNamespacedTestCase):
         label_name2 = 'test_label_02'
         lbl2 = self.test_client.update_label(label_name2, blob_uuid)
         self.assertIn('blobs', lbl2)
-        self.assertEqual(3, lbl['blobs']['1']['reference_count'])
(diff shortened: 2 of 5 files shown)


shakenfist/artifact.py
```diff
@@ -123,6 +123,9 @@ def add_index(self, blob_uuid):
             self.log.with_fields(entry).info('Added index to artifact')
             return entry
 
+    def del_index(self, index):
```
mikalstill, line 126: I hadn't thought this through. How do we handle not reusing an index? Or does it not matter if we reuse an index? It would have to be the highest index for it to be reused. Should we instead somehow mark indexes as deleted but leave them in the database? That will bloat the database over time. Discuss.

mikalstill (commented):
A minorish question.

shakenfist/artifact.py
mikalstill: Yeah, that seems fair. I think a "highest index" as an attribute of the artifact would also save the potentially expensive index scan we do here too.

mikalstill: https://github.com/shakenfist/shakenfist/issues/1060

mikalstill:
There's something wrong with Jenkins here. These failed test runs actually succeeded. I will debug tomorrow.

mikalstill:
This failure looks real:

```
shakenfist_ci.tests.test_snapshots.TestSnapshots.test_single_disk_snapshots
---------------------------------------------------------------------------

Captured traceback:
~~~~~~~~~~~~~~~~~~~
    Traceback (most recent call last):

      File "/home/debian/shakenfist/deploy/shakenfist_ci/tests/test_snapshots.py", line 125, in test_single_disk_snapshots
    self.test_client.delete_artifact_version(snapshot_uuid, 2)

      File "/usr/local/lib/python3.7/dist-packages/shakenfist_client/apiclient.py", line 484, in delete_artifact_version
    '/versions/' + version_id)

    TypeError: can only concatenate str (not "int") to str
```

I'll do a client patch now for it.

mikalstill:
CI failures are unrelated.

PR #1084, opened 2021-10-29
Limit number of versions kept of an artifact

mandoonandy:


mikalstill (approved):
LGTM, will merge when CI passes.

mikalstill:
This is good enough for now. I shouldn't punish Andy for flakey CI.

PR #1093, opened 2021-11-01
Retrieve artifact metadata with non-existent blob

mandoonandy:
When a new blob is added to an artifact ie. a
snapshot, the blob UUID is stored in the artifact
before the blob is created in the DB. This causes
an error when retrieving non-existent metadata for
an artifact.

A future fix should resolve this by only storing UUID's
in the DB that refer to objects already in existence.


mikalstill:
I think even with only storing uuids for objects which already exist we'd want this as defence in depth against some sort of DB inconsistency issue. LGTM, thanks.

mikalstill:
CI failures are all unrelated (my slow internet), so merging.

PR #1097, opened 2021-11-04
Handle graceful shutdowns and allow persistent config overrides.

mikalstill:


shakenfist/daemons/main.py
```diff
@@ -208,7 +210,7 @@ def _audit_daemons():
             if d not in running_daemons:
                 _start_daemon(d)
 
-        for d in DAEMON_PIDS:
+        for d in copy.copy(DAEMON_PIDS):
```
mandoonandy, line 213: to avoid importing copy, this line can use the built in dict copy  `DAEMON_PIDS.copy()`

To avoid making a full copy of the dict, just copy the keys `list(DAEMON_PIDS)`?

mandoonandy (approved):
LGTM (with a something nothing suggesion)

shakenfist/daemons/main.py
mikalstill: Huh. I didn't know that there was a copy() method on dicts. I think this is a good suggestion, but I think I'll handle it as an audit of the code base looking for all the examples that could be cleaned up. Thanks!

mikalstill: Actually, I'll do it now.

PR #1100, opened 2021-11-05
Cache instances to improve speed of artifact listing.

mikalstill:


mandoonandy (approved):
LGTM

 I'm seeing a halving on the artifact list time

mikalstill:
If we decide we like https://github.com/shakenfist/shakenfist/pull/1101, then we should not merge this.

mikalstill:
#1101 is going to win here as the more generic and useful solution.

PR #1103, opened 2021-11-06
Do not persist blob reference images.

mikalstill:
At the moment we create artifacts of type image when we need to
refer to a specific blob -- for example starting an instance. This
is actually pretty annoying as a user because you end up seeing a
lot of artifacts like this in the artifact list:

   sf://blob/6c3aa525-5af5-4c9e-8371-3549e88e2de8

On the other hand its nice as a developer, because it means I can
treat a blob as just another artifact when starting an instance.

So instead of making someone unhappy, let's just have the concept
of an artifact which isn't persisted to the database. I get my nice
abstraction, and the database stays cleaner.

### DIFF
--- a/shakenfist/baseobject.py
+++ b/shakenfist/baseobject.py
@@ -1,5 +1,6 @@
 
 from functools import partial
+import json
 import time
 
 from shakenfist import constants
@@ -24,9 +25,13 @@ class DatabaseBackedObject(object):
     STATE_DELETE_WAIT = 'delete_wait'
     STATE_ERROR = 'error'
 
-    def __init__(self, object_uuid, version=None):
+    def __init__(self, object_uuid, version=None, in_memory_only=False):
         self.__uuid = object_uuid
         self.__version = version
+
+        self.__in_memory_only = in_memory_only
+        if self.__in_memory_only:
+            self.__in_memory_values = {}
 
         self.log = LOG.with_object(self)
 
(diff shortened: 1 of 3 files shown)


shakenfist/baseobject.py
```diff
@@ -71,23 +77,40 @@ def _db_get(cls, object_uuid):
...
+
         for key, data in etcd.get_all('attribute/%s' % self.object_type,
                                       self.__uuid, prefix=attribute_prefix):
             yield key, data
 
     def _db_set_attribute(self, attribute, value):
+        if self.__in_memory_only:
+            self.__in_memory_values[attribute] = json.dumps(
+                value, indent=4, sort_keys=True, cls=etcd.JSONEncoderCustomTypes)
+
```
mandoonandy, line 106: if there an `else` missing here?

shakenfist/baseobject.py
```diff
@@ -71,23 +77,40 @@ def _db_get(cls, object_uuid):
...
         etcd.put('attribute/%s' % self.object_type,
                  self.__uuid, attribute, value)
 
     def _db_delete_attribute(self, attribute):
+        if self.__in_memory_only and attribute in self.__in_memory_values:
+            del self.__in_memory_values[attribute]
+
```
mandoonandy, line 113: if there an else missing here?

shakenfist/baseobject.py
mikalstill: Oh yes, good catch. Thanks!

shakenfist/baseobject.py
mikalstill: And this one too. Thanks!

PR #1105, opened 2021-11-08
Reduce logging for etcd caching, and start collecting simple statistics per thread.

mikalstill:


shakenfist/etcd.py
```diff
@@ -98,7 +113,12 @@ def get(self, key):
             raise exceptions.PrefixNotInCache(
                 'The prefix %s is not in the cache' % key)
 
-        return self.cache.get(key)
+        val = self.cache.get(key)
+        if val:
+            _increment_statistics('cache_miss')
```
mandoonandy, line 118: I think this might be backwards? You have a "hit" when the value is found.

mandoonandy:
This etcd cache will raise an exception if it's queried for a prefix it hasn't cached. So this PR can't count cache misses, but it will be counting actual DB absence. Is that intended?

mikalstill:
Yeah, what I mean by "miss" here is you cached instances, and then asked for an instance which didn't exist. It should be rare I suppose.

mikalstill:
This got a big re-write based on Andy's and my chat about #1106.

mandoonandy (approved):
LGTM

PR #1106, opened 2021-11-08
Automatically determine prefixes when caching etcd

mandoonandy:


shakenfist/etcd.py
```diff
@@ -64,54 +65,53 @@ def read_only_cache():
...
-                'The prefix %s is not in the cache' % key)
+    def _find_prefix(self, key):
+        uuid_regex = re.compile('.{8}-.{4}-.{4}-.{4}-.{12}')
 
-        return self.cache.get(key)
+        keys = key.split('/')
+        while keys:
+            if uuid_regex.match(keys.pop()):
+                return '/'.join(keys)
+        raise ValueError('Attempt to cache etcd key without a UUID: %s' % key)
```
mikalstill, line 94: So this wouldn't work with say, node metrics, right?

mandoonandy:
I've left the excessive logging in place in order to easily verify correct operation. It should be removed before merging and probably collides with PR #1105.

shakenfist/etcd.py
```diff
@@ -64,54 +65,53 @@ def read_only_cache():
...
+                 'and cached %d keys from %s' % (
+                     time.time() - start_time, len(self.cache), prefix))
+        self.prefixes.append(prefix)
 
-        if not whitelisted:
-            raise exceptions.PrefixNotInCache(
-                'The prefix %s is not in the cache' % prefix)
+    def get(self, key):
+        if not self._cached(key):
+            self._cache_prefix(self._find_prefix(key))
```
mikalstill, line 108: Is this a weird performance choice? What if I genuinely only want one instance for example? Additionally, should this also cache the associate attributes?

mikalstill (commented):
I do like the magic here, but I'm struggling to tell what the performance cost of turning single instance lookups into prefix caches is...

mikalstill:
This PR would also invalidate most of the cache statistics in #1105 wouldn't it?

shakenfist/etcd.py
mandoonandy: If you want only one instance then don't wrap with the context?

shakenfist/etcd.py
mandoonandy: Correct. The scheduler effectively builds a cache of decision data so I judged it a fair trade off to keep the code simple.

I see this context as something only deployed as necessary, not a layer placed everywhere?

mandoonandy:
Is there a performance trade-off? Within the cache context you can only access the keys that have been pre-specified, not single keys.

This is just lazy loading of the prefix instead of raising an exception.

mikalstill:
Isn't the performance trade off if I only want one instance this will now load all instances? Maybe I'm working on one instance but lots of networks or something? That said, prefix scans seem quite cheap in etcd... I think we think they're expensive because they're the things which log, not the things which are nessesarily broken.

mikalstill:
Thanks for the phone chat. This sounds good to me now that you've made me smarter.

PR #1112, opened 2021-11-11
Reduce API timeout time.

mikalstill:
Now that all long lived tasks are queue jobs, we should reduce the
maximum amount of time an API request can take. This might help us
debug API lockups as well.

### DIFF
--- a/shakenfist/config.py
+++ b/shakenfist/config.py
@@ -22,11 +22,11 @@ class SFConfig(BaseSettings):
     # API Options
     API_PORT: int = Field(13000, description='Port for the REST API')
     API_TIMEOUT: int = Field(
-        900,
+        30,
         description='How long gunicorn processes can use for a single request'
     )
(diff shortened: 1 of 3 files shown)


mikalstill:
Actually, I don't think this is safe right now, because movement of blobs between nodes in the cluster is done with this API server as well, and those will definitely take longer than 30 seconds.

mikalstill:
I've misunderstood API timeouts in gunicorn it seems. This is apparently how long an _idle_ worker lives before being reaped, not how long a HTTP transaction which is not idle survives. That is, back in the day we'd get a timeout because we'd stop talking to the client while we did a slow thing. That's gone, but now we sometimes talk actively at a client for a long time while we transfer a big thing (a blob). Those long transfers should not be reaped because of a lower timeout.

Refer to https://docs.gunicorn.org/en/stable/settings.html#timeout for details.

PR #1122, opened 2021-11-13
Manage blob replication levels

mikalstill:
Some light refactoring, and then adding two queue tasks. One replicates a blob to the target, and the other removes the blob from a target. Use the first to replicate blobs whenever they are created to config.BLOB_REPLICATION_FACTOR nodes. The deletion task isn't used yet.

### DIFF
--- a/shakenfist/blob.py
+++ b/shakenfist/blob.py
@@ -6,15 +6,16 @@ import os
 import random
 import requests
 import time
+import urllib3
 
 from shakenfist.baseobject import (
     DatabaseBackedObject as dbo,
     DatabaseBackedObjectIterator as dbo_iter)
 from shakenfist.config import config
-from shakenfist.constants import LOCK_REFRESH_SECONDS, GiB
+from shakenfist.constants import ETCD_ATTEMPT_TIMEOUT, LOCK_REFRESH_SECONDS, GiB
 from shakenfist import db
 from shakenfist import etcd
-from shakenfist.exceptions import BlobDeleted
+from shakenfist.exceptions import BlobDeleted, BlobFetchFailed
 from shakenfist import instance
 from shakenfist import logutil
 from shakenfist.node import nodes_by_free_disk_descending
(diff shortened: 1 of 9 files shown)


shakenfist/blob.py
```diff
@@ -193,6 +203,58 @@ def ref_count_dec(self):
...
+        os.rename(blob_path + '.partial', blob_path)
+        self.observe()
+
+    def request_replication(self):
+        targets = config.BLOB_REPLICATION_FACTOR - len(self.locations)
+        if targets > 0:
+            by_disk = {}
+            for n in Nodes([node_active_states_filter]):
+                metrics = db.get_metrics(n.fqdn)
+                disk_free_gb = int(metrics.get('disk_free', '0')) / GIGABYTE
```
mandoonandy, line 245: Python3 does true division. Wrap the int around the whole equation.

shakenfist/blob.py
```diff
@@ -193,6 +203,58 @@ def ref_count_dec(self):
...
+                by_disk.setdefault(disk_free_gb, [])
+                by_disk[disk_free_gb].append(n.fqdn)
```
mandoonandy, line 247: Should this check for minimum space available? Or better, check for minimum space after the blob has been replicated?

shakenfist/constants.py
```diff
@@ -1,3 +1,8 @@
+# Note the most exciting constants ever
+KILOBYTE = 1024
+MEGABYTE = 1024 * 1024
+GIGABYTE = 1024 * 1024 * 1024
```
LudvikGalois, line 4: I don't suppose I can convince you to throw style to the wind, and use `KiB`, `MiB` and `GiB` instead? `MEBIBYTE` looks stupid, but a "megabyte" could be one of 3 distinct values (a million bytes, 1000KiB or 1024KiB). This forces a new dev to check this file, making the named constant less useful than the magic number.

shakenfist/blob.py
mikalstill: Done.

shakenfist/blob.py
mikalstill: I've had a go at a tweak to do this.

mikalstill: I've tweaked the matching place in the scheduler too.

shakenfist/constants.py
mikalstill: Seems reasonable. Done.

mikalstill:
One thing which worries me about managing blob replicatation down is that we don't current have _any_ locking on blobs anywhere in the code base. I worry about a blob which has been fetched to a node to start an instance, but the instance has not yet started, and then the cleaner removes the blob causing the instance start to fail. I think we need to do a bit of a bench audit to ensure we have the right locks around blobs.

mikalstill:
Although maybe that's not true because an instance in state INITIAL is counted as a user of the blob in the Blob.instances() code.

shakenfist/blob.py
```diff
@@ -198,6 +208,67 @@ def ref_count_dec(self):
...
+        self.observe()
+
+    def request_replication(self):
+        replica_count = len(self.locations)
+        targets = config.BLOB_REPLICATION_FACTOR - replica_count
+        self.log.info('Desired replicate count is %d, we have %d'
+                      % (config.BLOB_REPLICATION_FACTOR, replica_count))
+        if targets > 0:
+            blob_size_gb = int(self.size / GiB)
+            by_disk = {}
```
mandoonandy, line 251: If you use `by_disk = defaultdict(list)` you can remove line 261 `by_disk.setdefault(disk_free_gb, [])`

(Adding the import `from collections import defaultdict`)

mikalstill:
I am running this branch on the CI cluster to shake out any weirdness, but it seems to mostly work without smoke coming out.

shakenfist/blob.py
mikalstill: Done.

shakenfist/daemons/cluster.py
```diff
@@ -37,37 +42,175 @@ class Monitor(daemon.Daemon):
...
+                        'blob': b,
+                        'ignored_locations': ignored_locations
+                    }).info('Ignored some blob locations as nodes are absent')
+
+                delta = len(locations) - config.BLOB_REPLICATION_FACTOR
+                if delta > 0:
+                    excess_locations = b.locations
+                    in_use_locations = []
+
+                    for instance_uuid in b.instances:
```
mandoonandy, line 155: This line iterates every disk of every instance, and is repeated for every blob. Would be faster to build a dict of instances using each blob. I realise that the data is cached in memory.

Maybe leave that optimisation till later?

mandoonandy (approved):
LGTM but a bit large to fully grok.

shakenfist/daemons/cluster.py
mikalstill: Yeah, I feel like we both agree this is a bit big in terms of patch size now. Let's merge this and I'll do the optimization and the "allow excess" thing in another change.

mikalstill:
I'm going to convert this one to a draft and try and split out logical steps now that it works into separate PRs.

mikalstill:
This is now the remainder of that series after the other things had been teased out.

mandoonandy (approved):
Lots of logic in there. Lets take it for spin around the block.

PR #1134, opened 2021-11-18
Add instance tags and CPU affinity

mandoonandy:
Still needs tests. Request comments before that job.

Depends on shakenfist/client-python#134

### DIFF
--- a/shakenfist/external_api/instance.py
+++ b/shakenfist/external_api/instance.py
@@ -92,7 +92,8 @@ class InstancesEndpoint(api_base.Resource):
     @api_base.requires_namespace_exist
     def post(self, name=None, cpus=None, memory=None, network=None, disk=None,
              ssh_key=None, user_data=None, placed_on=None, namespace=None,
-             video=None, uefi=False, configdrive=None):
+             video=None, uefi=False, configdrive=None, affinity=None,
+             tags=None):
         global SCHEDULER
 
         if not namespace:
@@ -244,8 +245,38 @@ class InstancesEndpoint(api_base.Resource):
             if 'memory' not in video:
                 return api_base.error(400, 'video specification requires "memory"')
 
+        # Verify tags
+        if tags:
+            if not isinstance(tags, list):
+                return api_base.error(400, 'tags should be a list of strings')
+            for t in tags:
+                if not isinstance(t, str):
+                    return api_base.error(400, 'tag (%s) is not a string' % t)
+
+        # Verify affinity settings
+        if affinity:
+            if not isinstance(affinity, dict):
+                return api_base.error(400, 'affinity should be a dictionary')
+
+            for key_type, aff in affinity.items():
+                # Check the validity of each affinity setting
+                if key_type not in ('cpu', 'disk', 'instance'):
+                    return api_base.error(
+                        400, 'can only set affinity for cpu, disk or instance')
+                if not isinstance(aff, dict):
+                    return api_base.error(
+                        400, 'an affinity must be a dictionary (%s)' % aff)
+                for tag, val in aff.items():
+                    try:
+                        affinity[key_type][tag] = int(val)
+                    except ValueError:
+                        return api_base.error(
+                            400, 'affinity (%s) contains non integer value (%s)' % (
+                                key_type, val))
+
         # Create instance object
         inst = instance.Instance.new(
+            affinity=affinity,
             name=name,
             disk_spec=disk,
             memory=memory,
@@ -256,7 +287,8 @@ class InstancesEndpoint(api_base.Resource):
             video=video,
             uefi=uefi,
             configdrive=configdrive,
-            requested_placement=placed_on
+            requested_placement=placed_on,
+            tags=tags,
         )
 
         # Initialise metadata
--- a/shakenfist/external_api/instance.py
+++ b/shakenfist/external_api/instance.py
@@ -92,8 +92,7 @@ class InstancesEndpoint(api_base.Resource):
     @api_base.requires_namespace_exist
     def post(self, name=None, cpus=None, memory=None, network=None, disk=None,
              ssh_key=None, user_data=None, placed_on=None, namespace=None,
-             video=None, uefi=False, configdrive=None, affinity=None,
-             tags=None):
+             video=None, uefi=False, configdrive=None, metadata=None):
         global SCHEDULER
 
         if not namespace:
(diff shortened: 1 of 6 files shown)


shakenfist/external_api/instance.py
```diff
@@ -91,7 +91,8 @@ def get(self, all=False):
     @api_base.requires_namespace_exist
     def post(self, name=None, cpus=None, memory=None, network=None, disk=None,
              ssh_key=None, user_data=None, placed_on=None, namespace=None,
-             video=None, uefi=False, configdrive=None):
+             video=None, uefi=False, configdrive=None, affinity=None,
+             tags=None):
```
mikalstill, line 95: What is the difference between a tag and instance metadata? Should this instead be a more generic way of setting metadata on instance creation, that then appears via the metadata APIs?

shakenfist/external_api/instance.py
```diff
@@ -243,8 +244,38 @@ def post(self, name=None, cpus=None, memory=None, network=None, disk=None,
...
+                    return api_base.error(400, 'tag (%s) is not a string' % t)
+
+        # Verify affinity settings
+        if affinity:
+            if not isinstance(affinity, dict):
+                return api_base.error(400, 'affinity should be a dictionary')
+
+            for key_type, aff in affinity.items():
+                # Check the validity of each affinity setting
+                if key_type not in ('cpu', 'disk', 'instance'):
```
mikalstill, line 262: Can I specify an affinity with an instance in a different namespace?

shakenfist/scheduler.py
```diff
@@ -239,73 +240,89 @@ def place_instance(self, instance, network, candidates=None):
...
+            if cpu_affinity:
+                for i in instance.Instances([instance.healthy_states_filter]):
+                    if not i.tags:
+                        continue
+                    for tag, val in cpu_affinity.items():
+                        if tag in i.tags:
+                            # Use get() to allow for unplaced healthy instances
+                            pseudo_load[i.placement.get('node')] += val
+
+            inst.add_event('schedule', 'Pseudo load due inst affinity',
```
mikalstill, line 309: I feel like "pseudo load due inst affinity" needs to be an english sentence which would make sense to users given its in a user facing event message.

shakenfist/scheduler.py
```diff
@@ -239,73 +240,89 @@ def place_instance(self, instance, network, candidates=None):
...
+                           None, str(dict(pseudo_load)))
+
             # Order candidates by current CPU load
-            by_load = {}
+            by_load = defaultdict(list)
             for n in copy.copy(candidates):
                 load = math.floor(self.metrics[n].get('cpu_load_1', 0))
-                by_load.setdefault(load, [])
+                load -= pseudo_load.get(n, 0)
```
mikalstill, line 316: Why do we substract the pseudo load here? I would of thought we added it?

mikalstill (changes requested):
We should chat about this. I overall like the idea, but I've asked a few questions...

shakenfist/scheduler.py
mandoonandy: The value supplied is affinity. The node with the lowest load is chosen, therefore if an instance has affinity with another instance it has to reduce the CPU load on a node in order to be placed on that node.

It follows that an instance expresses anti-affinity with a negative number. `--cpuaffinity nastyvm=-99`

(I've written very little doc since I'm just floating the idea and seeing if passes CI.)

shakenfist/scheduler.py
mandoonandy: good point.

shakenfist/external_api/instance.py
mandoonandy: Maybe [he says coyly]. This would be very useful but depends on whether all users are trusted. Maybe make that a installation configuration option.

This is probably a bigger picture decision, whether the whole installation should have this an installation option?

shakenfist/external_api/instance.py
mandoonandy: Tags are more controlled in there format ie. a list of strings. Unless we want to uses keys in metadata as tags? That seemed a bit too overloaded but I could be convinced.

mikalstill: I was thinking we could have a metadata key for the instance called "tags", which requires that there be a list of values as its value? I guess I was also thinking that I should be able to set _any_ metadata key on the instance create call, not just tags?

shakenfist/external_api/instance.py
mikalstill: We've talked before about a heirarchy of namespaces, and I think this falls into that future looking category to be honest. I'd have a namespace for DB servers for example, and then I could say "I don't like sharing with that namespace" somewhere else?

shakenfist/scheduler.py
mikalstill: I don't know how I feel about using CPU load here. Doesn't that mean an affinity might result in you landing on a very overloaded node? I feel like maybe this should be a separate filter?

mandoonandy: Remember that this filter is after the `_has_sufficient_cpu` filter, so complete overcommit is not possible.

If this is a separate filter, then the user loses the ability to say "I want this instance away from that other instance, only if the non-optimal CPU load imbalance is less than this value".

I would suggest, implement this `CPU affinity` (and `disk space affinity`) and also implement a pure affinity that is actually a binary decision to share a node with a particular tag.

This is only "affinity" not outright placement "control". The affinity value should be low to express "I desire to be close to another instance". To express "I demand" then it can be a value higher than the max possible CPU load.

The user might know that the load will reduce after instance startup, so is happy with initial higher load. This is about letting the user utilise their knowledge and communicate what is best, without the user demanding a placement that would be detrimental.

shakenfist/external_api/instance.py
mandoonandy: I think there is two suggestions here? Move `tags` to metadata, and secondly add metadata to the `instance create` API?

The reasoning behind making tags a separate parameter, is that tags should be static after instance start. If the tags are changed whilst the instance is running, should the instance move nodes?

mikalstill: Oh I agree that tags / affinity are static values. I hadn't thought about the implications of that if we smash them into the existing metadata. It just feels weird to me to have two parallel metadata systems. I guess there are other forms of metadata which aren't updated -- for example if I implemented a default ssh key in namespace metadata, then you'd get the value at the time of instance create and it wouldn't be updated.

I guess that's the other thing here, if it was metadata then we could do metadata inheritance -- set these on the namespace if you want them to apply to all your instances for example.

shakenfist/scheduler.py
mikalstill: Does that imply that for me to specify "I demand", I need to know what the max CPU load for a given node is, or do I just use a really big number like 100?

mandoonandy: Correct. Use a really big number. Turns the system in a control system rather than affinity - allowing for overcommit limits.

In reality, forcing an instance on or off a node when the CPU load advantage is greater than an entire physical CPU seems counter-productive.

shakenfist/external_api/instance.py
mandoonandy: That's a good point. If we implement the hierarchy of namespaces, we could specify which namespaces can leak information between each other.

So, we currently stop interaction between namespaces for other purposes. This PR should (for now) be the same. Concur?

shakenfist/external_api/instance.py
mandoonandy: I think SSH keys are a prime target for an inheritance system. So there will be a number of parameters that could be inherited in such a system: `ssh_key`, `metadata`, `tags`. Tags don't need to be in metadata for that.

I think the most powerful argument to move `tags` to `metadata` is: what if we hate affinity using tags for scheduling and want to remove it? It would be better to not have `tags` in the API.

[moves to sitting on fence]

shakenfist/external_api/instance.py
mikalstill: I agree for now. I see the leakage between namespaces being expressed by the heirarchy -- if you have a common owner other than system, you can share.

mikalstill:
So where are we with this? It seems like the sticking point is if we add tags to the API and then hate it they're hard to remove later, whereas if tags are a metadata value, then we're a bit less wedged. Thoughts?

mandoonandy:
As per my comments, I'm on the fence with the tag locations (since we may hate tags one day). To avoid any wedgies I will sway towards your position.

So, I'll move tags to metadata and enforce namespace separation unless the caller is system.

mandoonandy:
So I'm going to punt on the namespace isolation till we brainstorm that one.

If tags are namespaced and the `system` user is allowed to cross namespaces, the instance scheduler does not know that the system user is the owner when the instance is placed in another namespace. This is easily fixed with another DB value but maybe there is a better solution?

mandoonandy:
Reminder: Once we like the concept, I will add tests before merge.

mikalstill (approved):
FWIW this looks good to me. I know you've promised more tests. Its a bit hard to tell how it will perform in the wild, but there's one way to find out.

mandoonandy:
Rebased to develop and a few unittests added.

mikalstill (approved):

PR #1146, opened 2021-11-25
Report slowest CI tests.

mikalstill:


mikalstill:
Sigh, we already do this of course.

PR #1157, opened 2021-11-27
Lay the groundwork for nextgen events.

mikalstill:


mikalstill:
Was previously https://github.com/shakenfist/shakenfist/pull/1036, but that was against master not develop.

mandoonandy (commented):
I'm wondering if events should even pass though etcd. They do not need to be persisted. Given the load and issues that etcd experiences, maybe events should just bypass etcd. That would also remove a code path that is itself subject to potential failure.

Events could directly call the eventlog_node API. No slower than calling the etcd API and waiting for a confirmed write.

mikalstill:
Well, they do need to be persisted, they just don't need to be persisted in etcd. Overall I think I agree, but I am sneaking up on it. I think queueing them via etcd is a nice transition step, and I suspect our etcd problems are total key count, not traffic related.

mikalstill:
This really should be a draft given there's nothing here to actually read the emitted event logs.

shakenfist/logutil.py
```diff
@@ -132,14 +155,47 @@ def with_prefix(self, prefix=None):
...
     def process(self, msg, kwargs):
-        msg = '%s[%s] %s' % (setproctitle.getproctitle(), os.getpid(), msg)
-        kwargs["extra"] = self.extra
-
         if config.LOG_METHOD_TRACE:
             self._extra['method'] = util_callstack.get_caller(-5)
 
+        # Emit events: set hidden=True in fields to hide this log message from end users
+        if self.extra.get('is_event', False):
+            self._emit_event(msg, self._extra)
```
mikalstill, line 163: @mandoonandy this is the code we were talking about the other day which decides if a log message is also an event or not. It seems like I changed it to opt in already, but then forgot because this code is so old. The comment is also wrong of course.

mikalstill:
Replaced by a simpler approach in #1226.

PR #1164, opened 2021-12-07
First cut of removing excess blob replicas.

mikalstill:


mikalstill:
@mandoonandy, I think this is the one where you're going to want to start pushing back.

shakenfist/daemons/cluster.py
```diff
@@ -53,21 +56,116 @@ def _await_election(self):
...
+
+                    if len(in_use_locations) >= config.BLOB_REPLICATION_FACTOR:
+                        # We have enough locations in use to meet the replication
+                        # factor, remove everything else
+                        overreplicated[b.uuid] = excess_locations
+                    else:
+                        # We need to keep some idle locations to maintain our
+                        # replication factor. Prefer to remove the fullest first.
+                        overreplicated[b.uuid] = []
+                        target = config.BLOB_REPLICATION_FACTOR - \
```
mandoonandy, line 126: PEP8 prefers brackets for line continuation

shakenfist/daemons/queues.py
```diff
@@ -163,9 +163,11 @@ def handle(jobname, workitem):
                 n.delete_on_hypervisor()
 
             elif isinstance(task, FetchBlobTask):
+                log.with_fields({
```
mandoonandy, line 166: `with_field()`

shakenfist/node.py
```diff
@@ -130,3 +132,27 @@ def __iter__(self):
...
+
+
+def _sort_by_key(d):
+    for k in sorted(d):
+        for v in d[k]:
+            yield v
+
+
+def nodes_by_free_disk_descending(minimum=0, maximum=-1):
+    by_disk = {}
```
mandoonandy, line 144: another place to use `defaultdict(list)`

mandoonandy (changes requested):
Only one "real" change.

I feel I'm sniping from the sidelines here. I'm sure there a bigger issues I'm not picking up on.

shakenfist/blob.py
```diff
@@ -265,30 +262,24 @@ def ensure_local(self, locks):
...
+                minimum=blob_size_gb + config.MINIMUM_FREE_DISK)
+            if config.NODE_NAME in nodes:
+                nodes.remove(config.NODE_NAME)
 
-                by_disk.setdefault(disk_free_gb, [])
-                by_disk[disk_free_gb].append(n.fqdn)
-
-            for node in list(_sort_by_key(by_disk))[:targets]:
-                etcd.enqueue(node, {
+            for n in nodes[:targets]:
```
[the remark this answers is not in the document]
mikalstill: Python handles that for us -- we'll just get all `nodes` if `targets` is longer than `nodes`:

```
>>> a = [1, 2, 3]
>>> print(a[:100])
[1, 2, 3]
```

shakenfist/blob.py
```diff
@@ -265,30 +262,24 @@ def ensure_local(self, locks):
...
+                etcd.enqueue(n, {
                     'tasks': [FetchBlobTask(self.uuid)]
                 })
-                LOG.with_fields({
-                    'blob': self.uuid,
-                    'node': node
+                self.log.with_fields({
```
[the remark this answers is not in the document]
mikalstill: I have kind of fallen into a rut where I always use `with_fields()`, to the extent that I have started thinking about removing `with_field()` and the related helpers such as `with_instance()`. I think it would be good to have a conversation about if these extra helpers add much, and if the simplicity is worth having fewer ways of expressing the same thing.

shakenfist/daemons/cluster.py
```diff
@@ -53,21 +56,116 @@ def _await_election(self):
...
             a.delete_old_versions()
 
+        # Prune over replicated blobs
+        overreplicated = {}
+        underreplicated = []
+        nodes_by_disk_ascending = sorted(
+            nodes_by_free_disk_descending(), reverse=True)
+        low_disk_nodes = nodes_by_free_disk_descending(
+            minimum=0, maximum=config.MINIMUM_FREE_DISK)
+        absent_nodes = [n.fqdn for n in Nodes([node_inactive_states_filter])]
```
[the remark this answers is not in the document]
mikalstill: So the scheduler maintains a cache of metrics already, but it does it by itself. We could push that down into `ThreadLocalReadOnlyCache`, but is there likely to be much benefit?

shakenfist/daemons/cluster.py
mikalstill: As do I. I think this was my editor. I will fix.

shakenfist/node.py
mikalstill: Good catch.

shakenfist/daemons/cluster.py
```diff
@@ -53,21 +56,116 @@ def _await_election(self):
...
+                            break
+
+        for blob_uuid in overreplicated:
+            b = blob.Blob.from_db(blob_uuid)
+            for node in overreplicated[blob_uuid]:
+                LOG.with_fields({
+                    'blob': b,
+                    'node': node
+                }).info('Blob over replicated, removing from node with no users')
+                b.drop_node_location(node)
```
mikalstill, line 158: So you missed the bit I thought would trigger you -- this block keeps blob replication counts down to as close to the specified number as possible. Now, it does remove from the fullest nodes first, but if you're drowning in free disk it still cleans up. You were arguing a couple of weeks ago that we should only cleanup if we're disk constrained, and otherwise over replication wasn't a problem. Discuss.

shakenfist/blob.py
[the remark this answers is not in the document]
mandoonandy: Fair enough. I tend to prefer more explicit code.

shakenfist/blob.py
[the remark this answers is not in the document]
mandoonandy: `with_object()` automatically uses the correct object label. That is important for parameterised logging.

`with_instance()` et al could probably be removed in favour of `with_object()`

shakenfist/daemons/cluster.py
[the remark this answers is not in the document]
mandoonandy: `nodes_by_free_disk_descending` is not cached here.

The benefit I see is twofold. Firstly avoids coding surprises (that metrics can't be cached in the context), second the calculations done in this code snippet are completed over unchanging data.

shakenfist/daemons/cluster.py
mandoonandy: I was thinking this was the drip-feed of a larger branch?

Yes, it would be good to specify an upper and lower disk space bound before moving blobs.

mikalstill: It is the drip feed of a bigger series, but we can definitely change things along the way if we need to. That's the point of making it more reviewable.

mikalstill:
This is as reliable as I can get this one without some of the later reviews in this series (from the cluster-maintenance branch). I say we merge this while its a smaller reviewable unit and then I'll immediately rebase cluster-maintenance on top.

shakenfist/baseobject.py
```diff
@@ -58,9 +58,12 @@ def add_event(self, operation, phase, duration=None, msg=None):
...
         etcd.create(cls.object_type, None, object_uuid, metadata)
-        LOG.with_fields({
-            'object_type': cls.object_type
-        }).with_fields(metadata).debug('Object created')
+
+        # Mangle things a little so it matches other log messages for this object
+        log_metadata = metadata.copy()
+        log_metadata[cls.object_type] = log_metadata['uuid']
+        del log_metadata['uuid']
+        LOG.with_fields(metadata).debug('Object created')
```
mandoonandy, line 66: Could I suggest replacing 5 lines with 1 line before the `uuid` is added:
```
LOG.with_fields(metadata).with_field('uuid', object_uuid).debug('Creating object')
```

mandoonandy (changes requested):

shakenfist/baseobject.py
mikalstill: But we want to remove ("uuid", uuid) and replace it with ("instance", uuid). Right?

mandoonandy: Whoops. Yep.
```
LOG.with_fields(metadata).with_instance(object_uuid).debug('Creating object')
```

mikalstill: But that doesn't rename the uuid field to instance? Or am I being confused here?

mandoonandy: No I am being confused. I forgot this was a class method. So no log method defined.
```
LOG.with_fields(metadata).with_instance(object_uuid).debug('Creating object')
```

mikalstill: Oh, the above should read:

```
LOG.with_fields(log_metadata).debug('Object created')
```

Does that clear up the confusion?

mikalstill: Thanks for the call. Let us never speak of these things again.

mandoonandy (approved):
LGTM

PR #1165, opened 2021-12-14
Mock etcd DB and scheduler tests use case example

mandoonandy:
Unit tests have become terribly complicated requiring too much system knowledge. This is the start of making unit tests simple to write and trying to return them to something closer to unit tests (or maybe functional tests since mocking every call is too painful).

Only the Instance object is currently available. The scheduler tests have been modified to use the mock_etcd class, reducing the 811 line file to 348 lines.

Comments on the approach?

### DIFF
--- a/shakenfist/etcd.py
+++ b/shakenfist/etcd.py
@@ -45,14 +45,10 @@ class WrappedEtcdClient(Etcd3Client):
             api_path=api_path)
 
 
-# This read only cache is thread local, a bit like Flask's request object.
-# The basic idea is that you can use this cache a bit like playing billiards --
-# you declare what data you need, it gets cached, and then you can refer to it
-# as much as you like. If you request data that is not in the cache then that's
-# a foul and an exception is raised. Given this is a read only cache, once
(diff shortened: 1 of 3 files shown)


shakenfist/etcd.py
```diff
@@ -350,7 +346,7 @@ def put(objecttype, subtype, name, data, ttl=None):
 
 
 @retry_etcd_forever
-def create(objecttype, subtype, name, data, ttle=None):
+def create(objecttype, subtype, name, data, ttl=None):
```
mikalstill, line 349: LOL. Nice catch.

mikalstill (commented):
I think this looks like a nice approach, I haven't liked the current unit tests for a while which is why I've ended up preferring to write CI tests than unit tests. What do you see yourself needing to do to turn this into something ready for review?

mandoonandy:
Actually that's ready to land. We can add more comments and other objects over time. I would expect the default nodes might also be tweaked.

mikalstill:
CI failures are environmental.

PR #1183, opened 2021-12-24
Setup the namespace on an etcd master, not a hypervisor.

mikalstill:


mikalstill:
This does not work for reasons I don't really understand. Let's assume I'm wrong somehow really subtly.

PR #1200, opened 2021-12-30
Do not merge: debugging lost structured values from log messages.

mikalstill:


mikalstill:
Ok, I got that figured out at least.

PR #1201, opened 2022-01-01
Fix autocompletion for sf-client installation

mandoonandy:
Python Click made changes requiring changes to autocomplete setup.
https://github.com/pallets/click/pull/1622
First released in v8.0.0  2021-05-11


mikalstill:
LGTM, thanks for chasing.

mandoonandy:
I should note that this may/will break auto-completion on systems running old version of the click library. It's easily fixed, so maybe we can afford to be a little opinionated on what is supported?

PR #1204, opened 2022-01-02
Check metadata before instance creation

mandoonandy:


mikalstill (approved):
Looks reasonable. Let's wait for CI.

PR #1205, opened 2022-01-02
Reduce daemon shutdown time by avoiding sleeps

mandoonandy:
Using time.sleep() results in the cleaner only checking 
shutdown requests once per minute.
Waiting on a threading.Event results in an immediate
response.


mikalstill (approved):
Clever. I'm happy with this once CI passes.

PR #1207, opened 2022-01-02
Specify instance or network by name in API calls

mandoonandy:
Specify instance or network by name in API calls
Test instance and netowrk API using names
Add Network object to MockEtcd

Note that the tests are not wholly moved to the new etcd mock (too worn out)
Upside is that the testing is more complete.

### DIFF
--- a/shakenfist/external_api/base.py
+++ b/shakenfist/external_api/base.py
@@ -5,21 +5,19 @@ from flask_jwt_extended.exceptions import (
     RevokedTokenError, FreshTokenRequired, CSRFError
 )
 from flask_jwt_extended import decode_token, get_jwt_identity
-from functools import partial
 import json
 from jwt.exceptions import DecodeError, PyJWTError
 import requests
 import sys
 import traceback
-import uuid
 
-from shakenfist import baseobject
 from shakenfist.baseobject import DatabaseBackedObject as dbo
 from shakenfist.config import config
 from shakenfist.daemons import daemon
 from shakenfist import db
+from shakenfist import exceptions
 from shakenfist.etcd import ThreadLocalReadOnlyCache
-from shakenfist import instance
+from shakenfist.instance import Instance
 from shakenfist import logutil
 from shakenfist import net
 from shakenfist.upload import Upload
@@ -72,37 +70,22 @@ def caller_is_admin(func):
     return wrapper
 
 
-def valid_uuid4(uuid_string):
-    try:
-        uuid.UUID(uuid_string, version=4)
-    except ValueError:
-        return False
-    return True
-
-
 def arg_is_instance_uuid(func):
     # Method uses the instance from the db
     def wrapper(*args, **kwargs):
         with ThreadLocalReadOnlyCache():
-            inst_uuid = kwargs.get('instance_uuid')
-            if inst_uuid and not valid_uuid4(inst_uuid):
-                # Check if uuid is a valid name of an instance
-                filters = [
-                    partial(baseobject.namespace_filter, get_jwt_identity()[0]),
-                    instance.active_states_filter,
-                ]
-                for i in instance.Instances(filters):
-                    if i.name == inst_uuid:
-                        kwargs['instance_from_db'] = i
-                        break
-            else:
-                kwargs['instance_from_db'] = instance.Instance.from_db(inst_uuid)
+            try:
+                inst = Instance.from_db_by_ref(kwargs.get('instance_uuid'),
+                                               get_jwt_identity()[0])
+            except exceptions.MultipleObjects as e:
+                return error(400, str(e))
 
-            if not kwargs.get('instance_from_db'):
+            if not inst:
                 LOG.with_field('instance', kwargs.get('instance_uuid')).info(
                     'Instance not found, missing or deleted')
                 return error(404, 'instance not found')
 
+        kwargs['instance_from_db'] = inst
         return func(*args, **kwargs)
     return wrapper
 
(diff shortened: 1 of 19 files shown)


shakenfist/external_api/base.py
```diff
@@ -68,16 +72,37 @@ def wrapper(*args, **kwargs):
...
-            return error(404, 'instance not found')
+        with ThreadLocalReadOnlyCache():
+            inst_uuid = kwargs.get('instance_uuid')
+            if inst_uuid and not valid_uuid4(inst_uuid):
+                # Check if uuid is a valid name of an instance
+                filters = [
+                    partial(baseobject.namespace_filter, get_jwt_identity()[0]),
+                    instance.active_states_filter,
+                ]
+                for i in instance.Instances(filters):
```
mikalstill, line 94: This feels inefficient to me -- you have an instance object, and then you store its uuid and look it up again below. That said, I not its inside a cache block.

Do you think its worth re-writing this to have `instance=None` on line 87, and then only do an instance lookup if instance is still None on like 98?

shakenfist/external_api/base.py
```diff
@@ -156,16 +181,36 @@ def wrapper(*args, **kwargs):
...
     def wrapper(*args, **kwargs):
-        if 'network_uuid' in kwargs:
-            kwargs['network_from_db'] = Network.from_db(
-                kwargs['network_uuid'])
-        if not kwargs.get('network_from_db'):
-            LOG.with_field('network', kwargs['network_uuid']).info(
-                'Network not found, missing or deleted')
-            return error(404, 'network not found')
+        with ThreadLocalReadOnlyCache():
+            net_uuid = get_network_uuid(kwargs.get('network_uuid'))
```
mikalstill, line 205: Why are network lookups a method, and instance lookups are inline? You could probably also write a more generic "object lookup" by providing an object type to lookup method mapping if that makes sense.

shakenfist/tests/test_external_api.py
```diff
@@ -744,39 +724,59 @@ def test_delete_instance_metadata_bad_key(self, mock_get_lock,
...
+                                      uuid='12345678-1234-4321-1234-123456789012',
+                                      metadata={'foo': 'bar', 'real': 'smart'})
+        resp = self.client.delete(
+            '/networks/12345678-1234-4321-1234-123456789012/metadata/wrong',
+            headers={'Authorization': self.auth_header})
         self.assertEqual({'error': 'key not found', 'status': 404},
                          resp.get_json())
         self.assertEqual(404, resp.status_code)
 
+    def test_get_instance_by_name_1(self):
```
mikalstill, line 750: Why `_1`? I don't see a `_2`?

mikalstill (changes requested):
Generally good, some small questions.

shakenfist/external_api/base.py
mandoonandy: Good point. Will sort out.

shakenfist/external_api/base.py
mandoonandy: The network lookup is also used at https://github.com/shakenfist/shakenfist/blob/647f797b6f4f564db4b89e14af28cdeb21d3c042/shakenfist/external_api/instance.py#L270

It's a good idea to make it more generic, since the concept might stretch to artifacts.

shakenfist/tests/test_external_api.py
mandoonandy: I ran out of interest

mandoonandy:
Changes made for all of the above, plus changes to tests. I haven't changed all tests in order to keep this PR only to relevant changes. They still need to be done.

I quite like being able to `sf-client instance create test2 1 1024 -d8@cirros -f testnet`

To keep this PR smaller and more reliable, I have not moved any of the DBO Iterator classes to use the iterator in DBO. Only the new reference-by-name code uses the DBO iterator.

shakenfist/external_api/base.py
```diff
@@ -93,14 +93,13 @@ def wrapper(*args, **kwargs):
                 ]
                 for i in instance.Instances(filters):
                     if i.name == inst_uuid:
-                        inst_uuid = i.uuid
+                        kwargs['instance_from_db'] = i
```
mikalstill: Reusing kwargs is clever.

shakenfist/external_api/instance.py
```diff
@@ -501,114 +501,114 @@ def get(self, instance_uuid=None, instance_from_db=None):
...
-    @api_base.arg_is_instance_uuid
+    @api_base.arg_is_instance_ref
     @api_base.requires_instance_ownership
     @api_base.redirect_instance_request
     @api_base.requires_instance_active
-    def post(self, instance_uuid=None, instance_from_db=None):
+    def post(self, instance_ref=None, instance_from_db=None):
         with db.get_lock(
-                'instance', None, instance_uuid, ttl=120, timeout=120,
+                'instance', None, instance_from_db.uuid, ttl=120, timeout=120,
```
mikalstill, line 518: All of these `db.get_lock()` calls are a bit old school now. We should instead be using `instance_from_db.get_lock()`.

mikalstill:
This looks good to me, however the failure at https://sfcbr.shakenfist.com/jenkins/job/shakenfist_pr_debian_10_slim_primary/job/PR-1207/ looks real to me?

shakenfist/external_api/instance.py
mandoonandy: To stay focused and minimise risk, I'll leave that change to the next PR. (We want to release this immediately.)

mikalstill:
CI failures look real?

mandoonandy:
> CI failures look real?

That's me thinking "it's only a CI test, I can do a quick cut-paste" and not actually read the code properly.

mikalstill:
LGTM.

PR #1215, opened 2022-01-06
Make snapshot image compression configurable and other minor tweaks

mandoonandy:


shakenfist/config.py
```diff
@@ -166,6 +166,10 @@ class SFConfig(BaseSettings):
         1800, description='How long to wait for an image transfer to occur before giving up'
     )
 
+    COMPRESS_SNAPSHOTS: bool = Field(
+        False, description='Compress snapshots taken of instances'
```
mikalstill, line 170: I don't agree that the default should be false. I think this should default to true and then you can override it in your config file. Compression is the right choice, especially in terms of disk IO performance for new instances in most cases.

mikalstill (changes requested):
Just one small tweak please.

shakenfist/config.py
mandoonandy: I knew you wouldn't like that. I do posit that it's very slow and thus makes the user experience horrible.

mikalstill (approved):
Thanks.

PR #1220, opened 2022-01-10
Optionally load config from etcd.

mikalstill:


shakenfist/config.py
```diff
@@ -14,6 +17,22 @@ def get_node_name():
...
+    try:
+        value = Etcd3Client(
+            host='localhost', port=2379, protocol='http', api_path='/v3beta/').get(
+                '/sf/config', metadata=True)
+        if value is None or len(value) == 0:
+            return {}
+        return json.loads(value[0][0])
+
+    except ConnectionFailedError:
+        # NOTE(mikal): I'm not sure this is the right approach, as it might cause
```
mandoonandy, line 30: Probably good to log this for visibility.

shakenfist/config.py
```diff
@@ -14,6 +17,22 @@ def get_node_name():
     return socket.getfqdn()
 
 
+def etcd_settings(_settings):
+    try:
+        value = Etcd3Client(
+            host='localhost', port=2379, protocol='http', api_path='/v3beta/').get(
+                '/sf/config', metadata=True)
+        if value is None or len(value) == 0:
+            return {}
```
mandoonandy, line 26: Good to log that DB has no config.

mandoonandy (changes requested):
Just requesting some extra logging

mikalstill:
So the issue here is that pydantic runs this code very very early. Logging isn't setup yet, and in fact when I tried to use print statements here it didn't work in unit tests for $REASONS. So I agree logging would be good, but I haven't managed to come up with a feasible way to actually do that thing...

mikalstill:
That said, we already log every config options value on start. So it should hopefully be apparent if config isn't loading once someone starts looking?

shakenfist/config.py
mandoonandy: Fair enough. Until logging is sorted, the lack of config values should be a useful proxy.

mandoonandy (approved):
LGTM

PR #1227, opened 2022-01-29
Have a common implementation of hard delete.

mikalstill:


shakenfist/baseobject.py
```diff
@@ -237,6 +237,14 @@ def error(self, msg):
...
                     % (s, self.object_type))
         self._db_set_attribute('error', {'message': msg})
+        self.add_event('set error', 'complete', None, msg)
+
+    def hard_delete(self):
+        etcd.delete(self.object_type, None, self.uuid)
+        etcd.delete_all('attribute/%s' % self.object_type, self.uuid)
+        etcd.delete_all('event/%s' % self.object_type, self.uuid)
+        db.delete_metadata(self.object_type, self.uuid)
+        self.add_event('hard delete', 'complete', None, 'Hard deleted object')
```
mandoonandy, line 247: Should events be added to a hard deleted object? They will never be garbage collected.

mikalstill: So the new event code means events will never be garbage collected as they're not on etcd. I had figured this was good for billing etc.

That said, I think leaving a single vestial message saying why the data is missing is probably a good idea?

mandoonandy: So this is still in transition to the new sqlite event storage system? 

So the `delete_all('event/'...)` will become redundant and the `add_event()` code will be changed to write to the sqlite db (via etcd)?

mikalstill: Yes, that's a good call out. I should modify #1226 to remove this delete line if this merges first.

mikalstill: To be honest, I think this should merge first and then I'll tweak #1226 to remove the delete line when I rebase. I wanted to pull this one out separately as its actually pretty clear what's happening and it can be back ported to master if we want, whereas we're unlikely to backport the entire event system change.

mandoonandy (approved):
LGTM

PR #1277, opened 2022-03-01
Fix NetworkInterfacesEndpoint API bug

mandoonandy:


mikalstill:
LGTM, and I've retried that failed CI job.

PR #1314, opened 2022-04-13
Record blob existence earlier to avoid races.

mikalstill:


mikalstill:
The one remaining part of this PR has been pulled out into https://github.com/shakenfist/shakenfist/pull/1600.

PR #1316, opened 2022-04-13
Cleanup events.

mikalstill:


mikalstill:
Merging this because I think it needs #1314 in order to be more reliable.

PR #1317, opened 2022-04-17
Pin the sf-agent version.

mikalstill:


mikalstill:
Actually, I don't love this approach any more.

PR #1320, opened 2022-04-21
Move queue task consumption to using etcd watches.

mikalstill:
Also provide a nice sf-ctl command which will show you the status
of tasks in a given queue over time.


mikalstill:
I think this is a dead idea for now... Instead, we'll move to a queuing system outside etcd.

PR #1323, opened 2022-04-30
Use etcd watch for events.

mikalstill:


mikalstill:
I think this is a dead idea for now... Instead, we'll move to a queuing system outside etcd.

PR #1324, opened 2022-05-01
Move to OS packages for etcd.

mikalstill:


mikalstill:
This is a long dead though bubble.

PR #1338, opened 2022-06-20
Correct incomplete conversion to getsf.

mikalstill:


mikalstill:
Replaced by a simpler backport.

PR #1339, opened 2022-06-20
fix: network topology returning invalid IP NIC due to default route showing full string of iface

veenarm:
Fix as per: https://askubuntu.com/questions/1085789/default-route-interface-name


mikalstill:
Jenkins CI will fail because of unrelated configuration issues.

PR #1340, opened 2022-06-20
fix: get nic interface only, not whole string.

veenarm:
provided by: https://askubuntu.com/questions/1085789/default-route-interface-name


mikalstill:
CI failures are unrelated and already being worked on in #13137.

PR #1345, opened 2022-06-27
Sleep for a bit if etcd is angry.

mikalstill:


mikalstill:
This doesn't actually help, because the fix for this issue appears to be to delete keys from etcd, which is what the eventlog daemon is trying to do. We really need to get events out of etcd.

PR #1358, opened 2022-07-05
Fix broken artifact upgrades.

mikalstill:


mikalstill:
CI is broken until this fix lands.

PR #1387, opened 2022-08-09
Don't delete shared artifacts when deleting a namespace.

### LINKED ISSUE #1384: Deleting all artifacts in a namespace erroneously deletes all shared artifacts as well
Specifically this code is wrong in this context:

```
def namespace_filter(namespace, o):
    if namespace == 'system':
        return True
    if o.namespace == Artifact.SHARED_WITH_ALL:
        return True
    return o.namespace == namespace
```

This means all shared artifacts will appear to be _in_ the namespace being deleted and be reaped.

mikalstill:
Fixes #1384.


mikalstill:
Flying blind here because these bugs broke CI.

PR #1391, opened 2022-08-09
Use a system package for fixtures.

mikalstill:
This works around https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003252


mikalstill:
CI is still mashed.

PR #1405, opened 2022-08-28
Use a /32 netmask for floating IPs.

mikalstill:
This is more exact, but it also simplifies some of the subnet code
when that lands.


mikalstill:
This is apparently a bad idea, for reasons I am not sure I currently understand.

PR #1428, opened 2022-10-02
Add scheduled tests for develop.

mikalstill:


mikalstill:
CI tests will fail because Debian 10 testing is currently broken for environmental reasons that this PR is helping chase.

PR #1432, opened 2022-10-02
Don't count deferred tasks as waiting for execution.

mikalstill:


mikalstill:
Debian 10 is broken, which is being chased in another PR.

PR #1813, opened 2023-03-24
Configure reasonable defaults for SPICE channel compression.

mikalstill:
The underlying qemu defaults are reasonable, but not particularly obvious. Let's me more explicit. Also force on streaming mode for all "rapidly changing windows".


mikalstill:
Debian 10 test failure is for a bug fixed in #1814.

mikalstill:
Ditto Debian 11.

PR #1844, opened 2023-04-03
Fix discovered_refs tracebacks.

mikalstill:


mikalstill:
Upgrade failure is because the IPs of the backed up and the test instance don't match. I need to rethink upgrade testing.

PR #2207, opened 2024-01-01
Remove apt caches from CI images.

mikalstill:


mikalstill:
CI wont work until this is merged, so merging blind.

Issue #2242, opened 2024-01-26
Lock expiry in Instance._configure_block_devices()


mikalstill:
I keep seeing CI failures in `instance._configure_block_devices()`. I need to chase that down.

PR #2261, opened 2024-01-31
Drop the old etcd3 client.

mikalstill:
This means we wont be pinned to such an old grpc implementation. It does imply that we do compactions a little bit differently, but its not too bad.

At the same time, stop fetching metadata we're not using from etcd.

All of this is because it looks like #2242 might be related to etcd load...


mikalstill:
This was done in a different PR.

PR #2354, opened 2024-04-07
Cleanup all instances on node delete, not just the healthy ones.

mikalstill:


mikalstill:
This is clearly not a good idea.

PR #2701, opened 2024-12-08
Fixes for errors seen in CI.

mikalstill:


mikalstill:
Covered by the bisection series now.

PR #2714, opened 2024-12-17
Reuse etcd channels and retry failed requests.

mikalstill:


mikalstill:
At this point I think I just need to be all in on threads to make this reliable.

PR #2878, opened 2025-12-22
Finish off implementing states in mariadb.

mikalstill:


mikalstill:
> 1. **Missing `is_configured()` check** (shakenfist/mariadb.py:125-158, 161-198, 201-227, 230-260):
>    The `is_configured()` checks were removed from `get_state()`, `set_state()`, `delete_state()`, and `get_objects_by_state()`. This means if MariaDB is not configured, these functions will now raise a `RuntimeError` from `_get_connection_url()` instead of gracefully returning `None`/`False`/`[]`. This is probably intentional since MariaDB is now required, but it's a breaking change for any code that might call these functions before MariaDB is set up.

This review comment made me realise two things:

* services which depend on mariadb should abort on startup with a descriptive message if mariadb is not configured. Mariadb is now a hard requirement, not optional.

* secondly, in a perfect world only the database service should need the mariadb configuration -- I know at the moment its more like the eventlog, the sentinels, and the database service but we can fix that over time. Other services should _not_ have access to the mariadb configuration, and should be forced to go through the database service to access data. This allows us to enforce greater security later, for example a Node only being able to look up details of instances it is running.

So, I think we need to address those things.

mikalstill:
> 2\. **Migration command object type list** (shakenfist/client/ctl.py:163-176):
>     The `OBJECT_TYPES_WITH_STATE` list doesn't include all operation types (e.g., `artifact_fetch_op`, `imgcache_op`, `net_op`, etc.). These operation types are created via `base_mutations()` which now writes state to MariaDB. If an existing deployment has operations in etcd, they won't be migrated. Consider whether this list should be more comprehensive or if operation state is handled differently.

This seems real too. I am not too concerned about the case of queued operations that happen at upgrade time. I am able to ensure that 100% of my 1 users don't let this occur during the transition. However, the object type list should be complete given its a schema and all.

mikalstill:
> 3\. **Atomicity concern in operations** (shakenfist/schema/operations/util.py:45-47):
>     The `mariadb.set_state()` call happens before the etcd transaction in `base_mutations()`. If the subsequent etcd transaction fails, the state has already been written to MariaDB. This could leave orphaned state entries. Previously, state was part of the atomic etcd transaction.

I think this one is just part of the cost of transitioning data stores. Once we're fully on mariadb we can write everything in a SQL transaction and life will be good.

mikalstill:
> 4\. **Iterator efficiency** (shakenfist/baseobject.py:625-640):
>     The `DatabaseBackedObjectIterator.get_iterator()` now does two database calls: one to MariaDB for UUIDs, then iterates all objects from etcd and filters by the UUID set. For large clusters, iterating all objects from etcd could be expensive. Consider whether you could fetch only the matching objects from etcd using the UUID list directly.

This one is hard with etcd. It would either be N reads of exact keys, or this. It will improve as things move to mariadb.

mikalstill:
> 5\. **Index creation** (shakenfist/mariadb.py:97-100):
>     The index `idx_object_states_type_state` is created via `sa.Index()` but appears to be outside the table definition (after the closing parenthesis). This may not actually create the index. The pattern should either use `sa.Index(..., table.c.column)` with the table reference, or the index should be inside the `Table()` constructor.

This one seems real.

mikalstill:
> 6\. **Ansible MariaDB password** (deploy.py and deploy.yml):
>     The MariaDB password is generated but may not be persisted properly for repeat runs of the deployment. Ensure the password is stored/retrieved consistently.

I should definitely double check this.

mikalstill:
> > 6. **Ansible MariaDB password** (deploy.py and deploy.yml):
> > The MariaDB password is generated but may not be persisted properly for repeat runs of the deployment. Ensure the password is stored/retrieved consistently.
> 
> I should definitely double check this.

Claude assures me:

> The automated review concern appears to be a false positive. The MariaDB password is being persisted properly in /etc/sf/deploy-vars.json, just like admin_password and auth_secret. The difference from GETSF_ADMIN_PASSWORD in getsf is that getsf persists to /root/.getsfrc which is read before deploy.py runs, while deploy.py persists to /etc/sf/deploy-vars.json. Both mechanisms work correctly for repeat deployments. Conclusion: The MariaDB password persistence is working correctly. No changes are needed.

PR #2880, opened 2025-12-25
Mariadb for IPAM

mikalstill:


mikalstill:
> ### Type inconsistency in `reserve()` method (shakenfist/ipam.py:158)
> The `reservation_type` parameter is typed as `ReservationType` but when creating the `IPAMReservation`, it's passed directly. However, `IPAMReservation.reservation_type` is typed as `str`. This works due to `ReservationType(str, Enum)`, but it's worth verifying that Pydantic serializes the enum value correctly.

This seems fair.

mikalstill:
> ### Migration script key parsing edge case (shakenfist/client/ctl.py:344-350)
> ```python
> parts = key.split('/')
> if len(parts) < 5:
>     click.echo(f'  Skipping invalid key: {key}')
> ```
> 
> If an IP address contains a `/` (unlikely for IPv4 but possible in edge cases), this parsing could fail. Consider using `rsplit('/', 1)` to get the address from the end, or validate that `address` is a valid IP.

This one is not a thing. IP addresses never have a slash. That said, this column should be INET4 in the mariadb schema.

mikalstill:
> The new file lacks the standard copyright header that other files have.

Honestly, we're not that consistent about that already and I am not too worried. The git commit history makes it clear who the authors have been.

mikalstill:
> The method returns `Optional[Union[IPAMReservation, dict[str, Any]]]` which is inconsistent - for in-memory mode it returns a dict, for MariaDB mode it returns `IPAMReservation`. This complicates caller code. Consider always returning `IPAMReservation` (or always dict via `to_legacy_dict()`).

Fair.

mikalstill:
> ### Hardcoded string in halo release (shakenfist/mariadb.py:1097)
> ```python
> table.c.reservation_type == 'deletion-halo',
> ```
> 
> This should use `ReservationType.DELETION_HALO.value` for consistency with the rest of the codebase.

This one is already addressed.

mikalstill:
This comment will allow the automated exception fixer to run.

PR #2881, opened 2025-12-27
Add exception tracking to record unhandled exceptions.

mikalstill:
Add a new util/exceptions module that records unhandled exceptions to /srv/shakenfist/exceptions/ as JSON files. Each unique traceback is hashed and stored separately with occurrence counts and timestamps.

The module provides:
- install_exception_tracking(): installs hooks for sys.excepthook and threading.excepthook to capture unhandled exceptions
- record_exception(): records an exception to disk
- ignore_exception(): moved from util/general, now also records to disk

All daemon entrypoints now call install_exception_tracking() at startup. The sf-api Flask app also installs tracking at module load time.

Prompt: Add exception tracking to record unhandled exceptions, update all callers of ignore_exception to use the new location, and install exception tracking in all daemon entrypoints.

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -537,7 +537,7 @@ jobs:
     name: "Automated exception fixer"
     runs-on: [self-hosted, claude-code]
     needs: functional_matrix_pr
-    if: ${{ always() && contains(needs.functional_matrix_pr.outputs.check_logs_failed, 'true') }}
+    if: ${{ !cancelled() && contains(needs.functional_matrix_pr.outputs.check_logs_failed, 'true') }}
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-exception-fixer
       cancel-in-progress: true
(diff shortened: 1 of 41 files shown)


mikalstill:
> Incorrect usage of `traceback.format_exc()`

Sad panda. Will fix.

mikalstill:
> String encoding issue with `hashlib.sha256()`

Also a good catch.

mikalstill:
> The JSON data is written to the file but the data is bytes

True that.

mikalstill:
> Resource leak in error handling

Fixed.

mikalstill:
> **shakenfist/util/exceptions.py:53** - File truncation issue:
> When updating an existing file with a shorter JSON string, the old content will remain at the end. After `os.lseek(fd, 0, os.SEEK_SET)` and `os.write()`, you should call `os.ftruncate(fd, ...)` to truncate the file to the new length.

Luckily this code will always make the file _longer_...

mikalstill:
> **shakenfist/util/exceptions.py:1** - Missing copyright header per project style guidelines.

I don't think we even have a project style guide? Certainly we're not consistent about copyright statements in existing code.

mikalstill:
> The condition uses `always()` which means this job will run even if the workflow is cancelled

Fixed.

mikalstill:
> The `sf-api` entry point was removed

Yes, this was intentional.

PR #2885, opened 2025-12-29
Continue to chase mariadb changes for IPAM

mikalstill:


mikalstill:
> **backingStore XML generation may have formatting issues** (`instance.py:1072-1077`)
> 
> The f-string conversion for the backing store XML removes the newlines between elements that were present in the original `%` format string. The original had:
> 
> ```python
> '%(indent)s<backingStore type=\'file\'>'
> '%(indent)s  <format type=\'qcow2\'/>'
> ```
> 
> The new version concatenates these without newlines:
> 
> ```python
> f'{indent}<backingStore type="file">'
> f'{indent}  <format type="qcow2"/>'
> ```
> 
> This produces:
> 
> ```
>       <backingStore type="file">        <format type="qcow2"/>...
> ```
> 
> Instead of (original behavior):
> 
> ```
>       <backingStore type='file'>
>         <format type='qcow2'/>
> ...
> ```
> 
> While this may work (XML doesn't require newlines), it changes the formatting which could affect debugging/readability of the generated XML. Consider adding `\n` at the end of each line if preserving the original format is desired.

This is a good catch.

mikalstill:
> **No unit test coverage for the new XML validation**
> 
> The new `InvalidDomainXML` exception and the validation logic in `_create_domain_xml()` don't appear to have test coverage. Consider adding a test that verifies malformed XML triggers the expected exception and calls `enqueue_delete_due_error`.

This is covered by extensive functional testing however.

mikalstill:
> * The PR title mentions "mariadb changes for IPAM" but the changes are primarily f-string conversions and XML validation. The title may be misleading or this might be part of a larger series of changes.

Yes well. This change was mostly sent at this point to kick off the merge CI so I could chase mariadb failures there.

mikalstill:
@shakenfist-bot please re-review.

mikalstill:
> 1. **protos/_generate_enums.py:53-57** - The script exits with `sys.exit(1)` on parse failure but doesn't specify which file failed to parse. Consider adding the filename to the error message for easier debugging:
>    ```python
>    except SyntaxError as e:
>        print(f'Failed to parse {source_file}: {e}', file=sys.stderr)
>        sys.exit(1)
>    ```

Fair.

mikalstill:
> 2\. **shakenfist/schema/object_types.py:28-38** - The `__str__` method now returns `self.value.string`, but the docstring still references the old behavior. The docstring should be updated to reflect that `self.value` is now a `NamedTuple` with `string` and `proto_id` attributes.

Also fair.

mikalstill:
> 3\. **shakenfist/schema/object_types.py:50-51** - The comment "# Core objects" is now misplaced since it appears after `__str__` rather than before the enum members. Consider moving comments to maintain readability.

Claude disagrees:

> Looking at the current file, the comments are actually in the right place - the # Core objects comment is at line 107, right before the enum members start at line 108. The reviewer's comment references lines 50-51, which would have been the old line numbers before we added the docstring clarification. The current layout is correct:
> 
> * Lines 55-105: Methods (__new__, __str__, proto_id, from_proto_id)
> * Line 107: # Core objects comment
> * Lines 108-118: Core object enum members
> * Line 120: # Operation objects comment
> * Lines 121-136: Operation object enum members
> * etc.
> 
> The reviewer's comment appears to be based on an outdated view of the file. The comments are correctly placed before their respective enum member groups. No changes needed here.

mikalstill:
> 4\. **shakenfist/mariadb.py:373-377** - The `user_type` conversion uses `ObjectType.from_proto_id()` which could raise `ValueError` if an unknown proto_id is received. While this should theoretically never happen in a consistent system, consider whether this edge case needs handling (logging? returning None?).

Fair.

PR #2890, opened 2026-01-01
Move uploads to mariadb storage.

mikalstill:


mikalstill:
> ### Bug: Misplaced function call in cleaner daemon
> **shakenfist/daemons/cleaner/main.py:198** - The first call to `remove_stale_uploads_for_this_node()` is inside a block labeled "update power states" but has nothing to do with power states. Looking at the original code (line 142-155 in the diff context), `_remove_stale_uploads` was originally a method that was being called without parentheses on line 197, which would have been a no-op bug. The fix correctly calls the new function, but it's now in the wrong context block.
> 
> Consider either:
> 
> * Moving this call to a dedicated block, or
> * Removing this first call entirely since the same function is called again every 300 seconds (lines 212-216)

This was a real thing, but also surfaced a half implementation of the new style scheduled tasks in the cleaner daemon which needed fixing. I've fixed those.

mikalstill:
> ### Minor: Missing `hard_delete()` override in Upload class
> The `Upload` class overrides `_db_create()` and `from_db()` to use MariaDB, but `hard_delete()` is inherited from `DatabaseBackedObject` which still calls `etcd.delete()`. Since uploads no longer live in etcd, this could leave orphaned state records in MariaDB or cause issues. I'd recommend adding:
> 
> ```python
> def hard_delete(self):
>     mariadb.delete_upload(self.uuid)
>     mariadb.delete_state(self.object_type, str(self.uuid))
>     self.add_event(EVENT_TYPE_AUDIT, 'hard deleted object')
> ```
> 
> Or verify that `remove_abandoned_uploads()` correctly cleans up via `hard_delete()` - looking at line 191 it does call `upload.hard_delete()`, but this inherits the base implementation which tries to delete from etcd. This seems like a bug.

This is 75% true. The base class handles hard delete of both etcd and mariadb attributes, but it does not clean up this new upload table and that is required.

PR #2904, opened 2026-01-04
Convert blobs to mariadb

mikalstill:


mikalstill:
I am pretty tired of this UUID-as-string thing, but historic mistakes like using the node FQDN as a "uuid" make this hard to cleanup right now. I guess I'll accept this for now because mariadb is such a huge improvement over etcd, and then clean this up as soon as I can.

mikalstill:
> #### Potential Race Condition in `_direct_record_relationship` (mariadb.py:2207-2231)
> The upsert uses `INSERT ... ON DUPLICATE KEY UPDATE` which is good for idempotency, but the comment says "if the row already exists, update last_active". However, this will **always** update `last_active` even on first insert, which is the intended behavior. But consider: if two nodes simultaneously call `record_relationship` for the same relationship, one will insert and one will update - both will succeed, but with potentially different `last_active` values. This is probably fine since timestamps will be close, but worth documenting.

I don't even think this is worth documenting. We use this timestamp to determine if a reference is stale and should be deleted, where the timescale is in tens of minutes not nano-seconds.

mikalstill:
> #### Missing Validation in `observe_local_blobs` (blob.py:1000-1040)
> ```python
> def observe_local_blobs() -> int:
>     import pathlib  # Local import to avoid circular dependency
> ```
> 
> The local import comment mentions "circular dependency" but doesn't explain what the circular dependency is. If there's a real circular import issue, it would be better to restructure the imports at module level or document why this specific pattern is needed.

I had wondered about this one, but will admit I was being lazy. I've moved it to the top of the file because I can't see how this would ever be a circular import.

mikalstill:
> #### Inconsistent Error Handling in Migration Command (client/ctl.py:690-939)
> In `migrate_references_to_mariadb`, when parsing UUIDs fails for nodes (which use FQDNs as identifiers), the code correctly handles this for BLOB_LOCATION but there's a subtle issue:
> 
> ```python
> # Line 919-921
> # Node UUIDs are node names (strings like "sf-1"), not UUIDs
> success = mariadb.record_relationship(
>     ObjectType.NODE, node_name,  # node_name is a string, not UUID
> ```
> 
> But the function signature for `record_relationship` expects `UUID`:
> 
> ```python
> def record_relationship(
>     source_type: ObjectType,
>     source_uuid: UUID,  # Type hint says UUID
> ```
> 
> This type mismatch works at runtime because the code converts to string anyway, but it's misleading. Consider updating the type hints to `Union[UUID, str]` or documenting this exception.

This raises a couple of things for me. The type hint coverage for `ctl.py` was incomplete, as well as the issue raised here, although I will note this is all part of the uuids-as-strings pain we are working through right now. I will correct those.

mikalstill:
> #### Schema Version Migration Gap (mariadb.py:678-698)
> The `OBJECT_REFERENCES_VERSION = 2` and the migration from v1→v2 changes UUID columns to VARCHAR(255). This is a good change for Node FQDNs, but:
> 
> * What happens to existing UUID data during this migration? MariaDB should auto-convert, but this should be tested.
> * The migration runs `ALTER TABLE MODIFY COLUMN` which can be slow on large tables.

These are not issues because the previous table version never merged.

mikalstill:
> #### Cluster Daemon Change May Affect Cleanup Timing (daemons/cluster/main.py:218-231)
> ```python
> # If the blob's reference count is zero, we can reap it. With the
> # MariaDB-based object_references table, ref_count is computed
> # dynamically from actual relationships, so there's no drift.
> if b.ref_count < 1:
> ```
> 
> The old code waited for `ref_count_with_age['update_time']` to be > 300 seconds old before cleanup. The new code deletes immediately when `ref_count < 1`. This could cause issues if:
> 
> * A relationship is being created while the cleanup runs
> * The database query returns stale data
> 
> Consider adding back a grace period or using a transaction-safe approach.

We definitely still want the grace period to handle races.

PR #2930, opened 2026-01-10
Move blob hashes to mariadb.

mikalstill:


mikalstill:
> 1. **shakenfist/mariadb.py:3333-3337** - The `_direct_find_blob_by_hash()` query doesn't filter by `verification_status = 'valid'`. This could return a blob with an invalid hash. The `external_api/blob.py:348-351` endpoint does filter by status, but the underlying function might be used elsewhere without that protection. Consider adding the filter in the base function or documenting this expectation clearly.

Fair. I personally would prefer that where possible filtration is pushed down to the SQL layer where it can benefit from indices. I've strengthened CLAUDE.md to say that, as well as having this specific item resolved.

mikalstill:
> 3\. **shakenfist/mariadb.py:1293** - In `_migrate_etcd_blob_hashes()`, when `nodes` dict is empty, it defaults to `{'unknown': time.time()}`. The node name 'unknown' might cause confusion or issues if a real node has this name. Perhaps use a more distinctive sentinel like `'__migration_unknown__'` or document this behavior.

Explicit is always better I think.

PR #2931, opened 2026-01-11
Migrate blob transfers to mariadb.

mikalstill:


mikalstill:
> 1. **shakenfist/blob.py:654-655** - Missing transfer cleanup on failure path:
>    When transfer setup fails (lines 654-661), the transfer record created at line 640 is not deleted. The server side will eventually clean it up via stale transfer deletion, but this could leave orphaned records for 10 minutes. Consider adding `mariadb.delete_blob_transfer(locations[0], transfer_name)` before raising the exception.

This is a nice cleanup suggestion.

mikalstill:
> 2\. **shakenfist/daemons/transfers/main.py:59** - Token receive buffer size changed from 16 to 64 bytes:
>     The `recv(64)` should match what the client sends. Looking at `blob.py:670`, the client sends `token.encode('utf-8')`. The token is generated by `sf_random.random_id()` - verify this always produces tokens ≤64 bytes. The original code used `recv(16)` which seems too small if random_id() produces longer strings.

I think this one is actually not a problem because the `recv(64)` is merely specifying the _maximum_ amount to be read, it wont block.

mikalstill:
> 3\. **shakenfist/blob.py:621** - Late import inside method:
>     The `from shakenfist.schema.blob_transfer import BlobTransfer` import is inside the method. While this works, it's inconsistent with the import at the top of `mariadb.py`. Consider moving to module-level imports for consistency, or add a comment explaining why the late import is needed (e.g., circular import avoidance).

Yeah, I don't know why Claude has fallen in love with imports in the middle of methods, and I should tell it to stop. I missed this one during human review because I am a bad robot.

mikalstill:
> 4\. **shakenfist/blob.py:283-286** - Return type change is a breaking change for callers:
>     `incomplete_locations` property now returns `list[dict[str, Any]]` instead of `list[str]`. The diff shows `cluster/main.py:229` was updated, but verify no other callers depend on the old string format. The `external_view()` method was also updated, which is good.

I am pretty sure all the callers were changed, but can double check.

PR #2933, opened 2026-01-15
Migrate blob static values from etcd to MariaDB.

mikalstill:
This commit completes Phase 1 of the blob migration, moving static blob values (uuid, modified, fetched_at, version) to a new MariaDB `blobs` table. Key changes:

- Add BlobData Pydantic schema for static blob values
- Add MariaDB blobs table with CRUD operations and schema versioning
- Update Protocol Buffers with BlobData message and RPCs
- Add gRPC handlers in database daemon for blob operations
- Refactor Blob class to use MariaDB via _db_create/_db_get/from_db
- Replace Blobs iterator with get_active_blob_uuids() function
- Update all callers (cleaner, cluster, queues, external_api)
- Add comprehensive unit tests (15 tests)
- Update ARCHITECTURE.md to reflect blobs in MariaDB

The depends_on property now queries object_references instead of storing the value in static data, following the pattern established for blob locations and reference counts.

Prompt: Migrate blob static values from etcd to MariaDB following the Upload object pattern.


mikalstill:
> ### Type inconsistency in `hard_delete()` - shakenfist/blob.py:935
> ```python
> def hard_delete(self) -> None:
>     mariadb.delete_blob_hashes(str(self.uuid))
>     mariadb.delete_blob_transfers_for_blob(str(self.uuid))
>     mariadb.delete_blob(self.uuid)  # <-- self.uuid is already uuid.UUID, no str() wrapper
>     super().hard_delete()
> ```
> 
> The `delete_blob()` function expects a `UUID` type (as defined in mariadb.py:3131), and `self.uuid` from `DatabaseBackedObject` returns `uuid.UUID`. This should work correctly, but the inconsistency with the other two calls (which use `str(self.uuid)`) is worth noting for consistency. The other delete functions take `str`, while `delete_blob` takes `UUID`. This is fine but could be more uniform.

This is an issue with historical use of non-uuids in uuid fields. We're working on fixing it.

mikalstill:
> ### Missing `depends_on` handling in `Blob.new()` - shakenfist/blob.py:233-262
> The `new()` method receives `depends_on` as a parameter and includes it in the metadata dict for the event log, but it no longer stores it anywhere:
> 
> ```python
> def new(..., depends_on: Optional[str] = None) -> 'Blob':
>     metadata: dict[str, Any] = {
>         ...
>         'depends_on': depends_on,  # Included in metadata for logging
>         ...
>     }
>     Blob._db_create(blob_uuid, metadata)  # But _db_create doesn't use depends_on
> ```
> 
> The docstring for `BlobData` mentions that `depends_on` is stored in the `object_references` table, but I don't see where `Blob.new()` creates that relationship. Is there a separate call to `mariadb.record_relationship()` that should happen here, or is it handled elsewhere in the blob creation flow?

This seems like a thing to me. We'll fix this one.

mikalstill:
> ### `get_active_blob_uuids()` bypasses gRPC routing - shakenfist/mariadb.py:3166-3177
> ```python
> def get_active_blob_uuids() -> list[str]:
>     active_states = ['initial', 'created']
>     return _direct_get_objects_by_state(ObjectType.BLOB, active_states)
> ```
> 
> This function calls `_direct_get_objects_by_state()` directly, bypassing the `_use_database_service()` check. If this is intentional (because object state is always accessed directly?), a comment explaining why would be helpful.

This is because mariadb.py _is_ the database service. Its a bit unclear to me why this one specifically is being called out?

mikalstill:
> ### Unused import - shakenfist/blob.py
> The `eventlog` module is imported twice:
> 
> ```python
> from shakenfist import eventlog
> from shakenfist.eventlog import add_event_multi
> ```
> 
> The first import is only used for `eventlog.add_event()` in `_db_create()` and `from_db()`. Consider using just `from shakenfist.eventlog import add_event, add_event_multi` for consistency.

Minor, but worth cleaning up.

mikalstill:
@shakenfist-bot please re-review.

PR #2957, opened 2026-01-31
Report issues on docs

mikalstill:


mikalstill:
> **GitHub Actions artifact path issue** (`.github/workflows/docs-tests.yml:45`) The `path` value uses `${GITHUB_WORKSPACE}` which won't be expanded in YAML context. Environment variables in GitHub Actions `with` blocks are not automatically expanded. This should use the `${{ github.workspace }}` syntax instead:
> 
> ```yaml
> path: ${{ github.workspace }}/artifacts/site.zip
> ```

This was true. Good job little robot. The other comments are minor / already addressed.

PR #2961, opened 2026-02-02
Migrate blob attributes from etcd to MariaDB.

mikalstill:
This completes Phase 2 of the blob migration by moving mutable blob attributes (size, info, last_used, expires_at) from etcd to a dedicated blob_attributes table in MariaDB.

Key changes:
- Add BlobAttributesData Pydantic schema for type-safe attribute storage
- Add blob_attributes table with indexes on last_used and expires_at
- Add gRPC RPCs for blob attributes CRUD operations
- Update Blob class with lazy attribute loading from MariaDB
- Add migration function to move existing attributes from etcd
- Update cluster daemon to use database-level filtering for expired blobs and stale transcodes (get_expired_blob_uuids and get_stale_transcoded_blob_uuids)
- Update migration skill with lessons learned (Optional typing, version numbering)

The database-level filtering eliminates the need to load every blob object just to check one attribute, significantly improving performance of periodic maintenance tasks.

Prompt: Implement blob attributes migration from etcd to MariaDB per PLAN-blob-attributes-migration.md.


mikalstill:
> 1. **Missing Unit Tests** (moderate concern): There are no new unit tests for `BlobAttributesData` schema or the mariadb blob attributes functions. The existing test patterns in `test_mariadb_blob_hashes.py` and `test_mariadb_blob_transfers.py` could serve as templates. While functional tests may cover this, unit tests would provide faster feedback.

Fair call.

mikalstill:
@shakenfist-bot please retest

mikalstill:

mikalstill:
> ### 1. Late import of BlobAttributesData in database daemon (medium)
> **`shakenfist/daemons/database/main.py`** (new lines ~1663 and ~1722 in the diff)
> 
> `BlobAttributesData` is imported inside both `CreateBlobAttributes` and `UpdateBlobAttributes` handler methods. Since this schema is used across multiple handlers, consider importing it at the top of the file alongside the other schema imports (line 41-42). The existing late imports in the file (e.g., `State`, `IPAMReservation`) each appear in only one or two methods, but `BlobAttributesData` is used in at least two. It would be more consistent and efficient as a top-level import.

Fair.

mikalstill:
> ### 6. Missing state filter in expired/stale queries (medium)
> **`shakenfist/mariadb.py`** - `_direct_get_expired_blob_uuids` and `_direct_get_stale_transcoded_blob_uuids`
> 
> The original code in `cluster/main.py` iterated over `get_active_blob_uuids()` which returns only blobs in active states. The new queries operate on `blob_attributes` without joining to `object_states` to filter by state. This means the queries could return UUIDs for blobs that are already in `deleted` or `error` state. The caller (`_cluster_wide_cleanup`) does call `Blob.from_db(blob_uuid)` which will filter out deleted blobs, but the query could return more results than necessary. Consider adding a state filter join, or document that the caller handles this.

Also fair.

PR #3040, opened 2026-02-28
Remove pre-MariaDB upgrade steps from Blob.

mikalstill:
Bump initial_version from 2 to 8 and delete the six upgrade steps (2→3 through 7→8) that predate the MariaDB migration. These steps wrote to etcd, which is no longer read for blob data. This eliminates the last etcd
dependency from blob.py.

Prompt: I see two remaining users of etcd in shakenfist/blob.py. Let's come up with a plan to eliminate them.


mikalstill:

PR #3065, opened 2026-03-13
Migrate AgentOperation from etcd to MariaDB (Phase 12).

mikalstill:
Add dual-write support for AgentOperation static values and mutable attributes (results). Static values are stored in the agent_operations table, mutable results in agent_operation_attributes. Both use JSON columns for complex fields (commands list, results dict). MariaDB is tried first on reads with etcd fallback for unmigrated objects. Data migrations run automatically on database daemon startup.

Prompt: Let's implement this plan (referring to
PLAN-agentoperation-mariadb.md for Phase 12 migration).


mikalstill:

Issue #730, opened 2021-03-27
Health check dnsmasq and restart it when it stops serving DHCP


mikalstill:
I've seen this once in my home cluster. A restart of dnsmasq fixed it. We need to work out how to make sure dnsmasq is healthy all the time.

PR #3284, opened 2026-06-15
Health checks

mikalstill:


mikalstill:

mikalstill:
Thanks for the review — all seven items addressed. Details below, by item.

Addressed in two commits: `0a88d3e5` (items 1–2) and `ea33eac4` (items 3, 4, 6).

### 1 & 2 — 🐛 watchdog keepalive bypassed by wait/defer paths (FIX)

Real bug, good catch. Phase 3 armed `WatchdogSec=60s` but only petted the heavy *work* loops; the shared *wait/defer* paths that loop or `continue` before reaching `idle()` were missed. Auditing all eight armed daemons (as suggested), it was broader than the two flagged:

- The `while not health_check_nodelock(): time.sleep(1)` busy-wait (never pets) was in **cleaner, network, resources, transfers, sidechannel**. Replaced with a new shared `Daemon.wait_for_nodelock()` that waits via `idle(1)` — so it pets the watchdog and observes the abort path.
- The `if not cluster_stable(): … continue` that jumps over the bottom `idle()` was in **cleaner** and **cluster**; both now `idle()` on the defer path (pets + stops the busy-spin).
- **queues** and **database** have no such gap. **network**'s `cluster_stable` `continue` is a *for-loop* continue that still falls through to the bottom `idle(5)`, so it was left as-is (only network's nodelock wait needed fixing).
- The identical-looking nodelock waits in each daemon's `main()` run before `READY=1` (watchdog not yet armed) and are unchanged.

Added unit tests for `wait_for_nodelock` (pets while waiting; exits promptly on abort).

### 3 — drain / `--graceful-timeout` overlap (CONSIDER)

Correct — the windows overlap, they don't stack. The drain delays the worker's call to gunicorn's graceful shutdown by `API_DRAIN_GRACE` (25s), but the master's `--graceful-timeout` countdown starts at SIGTERM (t=0), so `--graceful-timeout 30` left only ~5s for in-flight requests after the drain. Set **`--graceful-timeout 55`** (`API_DRAIN_GRACE` + a 30s in-flight budget; `TimeoutStopSec=70` still exceeds it) and corrected the `sf-api.service` comment and `upgrades.md` to describe the overlap rather than a sequential model.

### 4 — watchdog tracks main loop, not worker threads (INFO)

Noted explicitly in `ARCHITECTURE.md`'s watchdog section: a wedged worker/gRPC thread under a healthy main loop keeps petting, so `WATCHDOG` covers the supervisor loop only; deeper per-worker liveness (e.g. #730) is future work and the signal shouldn't be over-trusted.

### 5 — `check_reachable` per-tick connect (CONSIDER)

Left as-is. At ~0.1/s the connect/auth churn is negligible, and the per-call `NullPool` engine is the deliberate way to get a bounded `connect_timeout` without disturbing the pooled engine (documented in the helper). Recorded as a possible future optimisation only.

### 6 — `/readyz` still passes through `resolve_node_uuid` (CONSIDER)

Fixed. `resolve_node_uuid` now short-circuits for health-probe paths, so the "a `/readyz` burst makes zero DB calls" guarantee holds **unconditionally** — previously it held only because `NODE_UUID` is pre-set on a configured node. Added a test asserting no `Node.from_db` call for `/readyz` with `NODE_UUID` unset.

### 7 — unauthenticated `/readyz` reachability signal (INFO)

No change — intentional (LBs must probe without auth), the bodies are minimal (strictly less than the existing unauthenticated `/`), and the firewall-to-LB-subnet posture is documented.

PR #3425, opened 2026-07-16
Restart libvirt/apparmor only on config change.

mikalstill:
The hypervisor role restarted apparmor and libvirtd with two unconditional `state: restarted` tasks that ran on every deploy, even when nothing changed. On a no-op redeploy this needlessly bounced libvirtd (briefly disrupting hypervisor management) and apparmor -- exactly the churn that makes folding sfcbr.yml into a daily rotation unattractive.

Convert them to handlers (roles/hypervisor/handlers/main.yml) and notify them only from the tasks that actually change the relevant files:

* the two /etc/libvirt/qemu.conf SPICE-TLS tasks notify Restart libvirt;
* every apparmor rule task (usr.sbin.libvirtd capabilities and the libvirt-qemu abstraction rules) notifies Restart apparmor and Restart libvirt.

The handlers are flushed with `meta: flush_handlers` at the exact point the unconditional restarts used to sit -- before `virsh net-destroy default` -- because a libvirtd restart after net-destroy would bring the autostart default network back up. On a genuine change the services still restart in the same order as before; on a no-op redeploy nothing is notified and nothing restarts. ansible-lint passes with the collection's config.

Prompt: proving the sfcbr no-op redeploy clean showed the deploy still bounces libvirt/apparmor every run because these restarts were unconditional. Michael asked to gate them so a repeated (and eventually scheduled) deploy leaves a quiet cluster untouched. Scoped to this one logical change; the remaining cosmetic always-changed command tasks are a separate follow-up.

Assisted-By: Claude Code

### DIFF
--- a/shakenfist/deploy/collection/roles/hypervisor/handlers/main.yml
+++ b/shakenfist/deploy/collection/roles/hypervisor/handlers/main.yml
@@ -10,10 +10,11 @@
 # restart tasks used to sit, so ordering relative to the later
 # `virsh net-destroy default` is preserved.
 
+# Boot-time enablement of apparmor is asserted by an always-run task in the
+# role body; this handler only performs the change-gated restart.
 - name: Restart apparmor
   ansible.builtin.service:
     name: apparmor
-    enabled: true
     state: restarted
 
 - name: Restart libvirt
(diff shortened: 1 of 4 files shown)


mikalstill:

PR #3557, opened 2026-07-29
Widen MariaDB ENUM columns as Python enums grow.

mikalstill:
MariaDB freezes an ENUM column's value list at CREATE TABLE time, so adding a member to a Python enum works on fresh installs but breaks existing databases: inserts of the new value fail with "Data truncated for column" (error 1265). This is how ObjectType.NAMESPACE_KEY broke namespace key state writes on upgraded clusters on 2026-07-28 -- every API request touching namespace keys returned 500, while greenfield CI kept passing.

ensure_schema() now ends with a reconciliation pass that discovers every native sa.Enum column from the SQLAlchemy metadata and widens stale columns in place, so future enum members (and future enum columns) need no hand-written migration. The previous helpers for this were dead code and quoted enum values where the DDL stores member names. Coverage: mock unit tests including a named regression test for the outage; live tests that reproduce the failing insert against a real MariaDB and prove recovery; and a new "Schema ENUM widening" CI job on an ephemeral VM runner gating the merge queue.

Prompt: After diagnosing why CI jobs were queueing (the sfcbr cluster API was returning 500s following the auth-federation deploy), Mikal asked for the shakenfist bug to be fixed on a new worktree/branch, with unit and CI test coverage to ensure this class of failure cannot recur.

Assisted-By: Claude Code

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -569,6 +569,43 @@ jobs:
           if-no-files-found: error
           path: /srv/github/artifacts/bundle.zip
 
+  # MariaDB ENUM columns freeze their permitted values at CREATE TABLE
+  # time, so a Python enum gaining a member silently breaks existing
+  # databases while greenfield CI deploys keep passing -- which is how
+  # ObjectType.NAMESPACE_KEY shipped without a widening migration and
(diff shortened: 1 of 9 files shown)


mikalstill:
Smash merging this one because sfcbr is borked and so CI can't run.

PR #3614, opened 2026-08-03
Scheduler reservations phase 2: capacity tables and reconciler

mikalstill:
Phase 2 of the scheduler reservations work, following phase 1 (#3578).
The plan is in
`docs/plans/PLAN-scheduler-reservations-phase-02-capacity-tables.md`;
the design decisions it implements (D1-D19) are in the phase 0
decisions document.

## What this does

Adds the three materialised-counter tables from decision D2 and the
reconciler from decision D5:

* `scheduler_node_capacity` -- per-node limits, used counters and
  `expected_demand`.
* `namespace_claims` -- per-namespace claims, with `state` and a
  server-side `expires_at`.
* `cluster_capacity` -- a singleton row of cluster totals, claimed and
  unclaimed usage.

The reconciler runs every five minutes on the cluster daemon's elected
leader, via a new `ReconcileSchedulerCapacity` RPC on `sf-database`.
It derives limits from the typed `node_metrics` columns added in phase
1, recomputes usage from `object_references` joined to `instances`
(summing `disk_spec` with `JSON_TABLE`), applies the D13 feedforward
demand decay, and publishes `scheduler_capacity_*` gauges and counters
on the cluster daemon's metrics port.

## What this deliberately does not do

Nothing consumes these tables for admission decisions. The reconciler
is the only writer, and the scheduler is untouched. This is the
observable-but-inert phase: it can be deployed and soaked while phase
3 (the claim primitive and consumption at `place_instance()`) is
written, and the gauges are what tell us whether the derived limits
and usage match reality before anything depends on them.

## Also here

`MIN_MARIADB_VERSION` is raised from 10.6.0 to 10.11.0. Validating
this phase against a real 10.6 server showed that floor had silently
stopped being real: `ipam_reservations.address` uses the native INET4
type, which only exists from MariaDB 10.10, so schema creation on 10.6
fails well after `verify_mariadb_compat()` has passed it. 10.11 is the
oldest in-support LTS above that, is what the debian-12 cluster CI
functional tests actually exercise, and ships with Debian 12/13 and
Ubuntu 24.04. There is a regression test that a 10.6 server is now
refused at compat-check time.

## Validation

Beyond the unit tests, the reconciler was run against a real MariaDB
with seeded nodes, instances and claims: 36 hand-computed expectations
checked, a second pass confirmed idempotent, and a 32-node /
1,205-instance scale set reconciled in 42ms -- comfortably inside the
watchdog budget.

## Deployment note

This adds schema, so `sf-ctl ensure-mariadb-schema` must run before
the daemons roll. Deployments using the collection's `site.yml` get
that automatically (play 5, before node registration).

*(Assisted by Claude Code)*


mikalstill:

mikalstill:

mikalstill:

mikalstill:

mikalstill:

Since the last review: rebased onto develop (the dedicated capacity-reconciler CI job is gone — develop's generalised ENUM-widening job now runs every live MariaDB module, so it was an exact duplicate). Both fix items are addressed: the usage query now binds each enum column's actual storage convention (names for the native-enum object_states column, values for the string object_references columns) and the live suite runs under utf8mb4_bin so that class of mismatch fails loudly; the freshness filter is now membership in a positive fresh set, so a node with no node_metrics row at all loses its capacity row. All three considers were accepted: phase-4 note on the per-claim UPDATE loop, the disk-size oracle now matches MariaDB's probed half-away-from-zero JSON-to-BIGINT rounding, and limit_memory_mb clamps at zero.

mikalstill:
GitHub Actions was in a major outage when the last push and re-review request landed, and the events were not replayed. Re-requesting: @shakenfist-bot please re-review

PR #3617, opened 2026-08-03
Reap obsolete indirect dependency pins.

mikalstill:
Replaces the append-only indirect dependency pinning job with the
reconciling version from
[shakenfist/development](https://github.com/shakenfist/development/tree/main/templates/pin-indirect-dependencies).

## Why

The old job only ever *appended* to the pinned block. Once a package
stopped being required by anything it kept its pin forever, and
Renovate kept dutifully proposing bumps for it. The new job
regenerates the block on every run:

* Existing pins are demoted to pip **constraints**. A constraint only
  applies if something still requires the package, so surviving pins
  keep exactly their current versions -- Renovate remains the only
  thing that moves a version -- while packages nothing requires any
  more simply drop out.
* The resolve runs against a copy of `pyproject.toml` with the block
  stripped out. This matters: a stale pin left in place is itself a
  requirement forcing its own installation, so it would never look
  stale.

## What changed

* The logic moves out of the workflow YAML into
  `tools/pin-indirect-dependencies.sh`.
* The job moves from the shared `static` pool to an ephemeral vm
  runner, because it now apt installs its own build dependencies.
* The devpi env block gains the `PIP_EXTRA_INDEX_URL` pypi fallback
  the `workflow-standards` audit asks for.
* `pyproject.toml` gains the missing `# START_OF_INDIRECT_DEPS`
  marker, and the prose note about pydantic-core becomes a machine
  readable `# never-pin: pydantic-core` line the script honours.

## The first reconciled pass

Included here so CI proves the reduced set still works. It was
generated by running the script against a `debian:12` container rather
than a workstation, because the resolved closure is python-version
specific and the runner is debian-12.

Twenty one pins nothing requires any more are removed:

> anyio, async-extensions, coloredlogs, entrypoint, exceptiongroup,
> executor, funcs, futurist, humanfriendly, linux-utils,
> mixed-methods, named, orderings, property-manager, sniffio, solus,
> typing-aliases, tzdata, verboselogs, wraps, zipp

None of them are imported anywhere in the tree. The rest of the diff
is the block being sorted case-insensitively.

*(Prepared with Claude Code)*


mikalstill:

mikalstill:

mikalstill:

Issue #936, opened 2021-09-02
Replace hand-rolled instance-create validation with a declarative schema


mandoonandy:
Video/disk/network specs are now validated, but imperatively (`instance.py:286-325,585-663`); the original "silent failure on missing fields" is mostly closed.

Residual: move to a declarative marshmallow schema for completeness and to catch the remaining gaps. (Related: #528, #534.)

_Rescoped 2026-06 to the residual work above; the original report is preserved in this issue edit history. See CATEGORIZED_BUGS.md on the `bug-cleanup` branch._

mikalstill:
Closing as part of a bug cleanup sweep of the oldest open issues.

Duplicate of #528 at a narrower scope. Both are named in `docs/plans/index.md` as parent issues of `docs/plans/PLAN-api-input-validation.md`, which is in progress and covers declarative validation across the API rather than just instance create. Consolidating onto #528 so there is one tracker.

Issue #528, opened 2020-11-11
Broaden declarative type/validity checking across all API endpoints


mikalstill:
The API's structured body parameters — a `diskspec`, a `networkspec`, a `videospec` — are described to the validation layer only by their outer shape. A `dict` compiles to `fields.Dict()` and an `arrayofdict` to `fields.List(fields.Dict())`, neither of which carries a value schema, so the layer checks that `disk` is a list of mappings and then hands every key inside those mappings to the handler unexamined. What validation those keys get is hand-rolled, imperative, and incomplete: an unknown key is silently discarded, and a wrong-typed value reaches code that cannot cope with it.

**Residual:** teach the type vocabulary to carry an element schema, so that a structure's keys, their types and their enums are described declaratively in the same place the published OpenAPI document is rendered from, and so that a key nobody documented is refused rather than dropped.

---

### About the scope of this issue

This issue was filed in 2020 as *"Broaden declarative type/validity checking across all API endpoints"*, and its rescoped 2026-06 text read: *"Only ~4 endpoint files use `use_kwargs` typed schemas today (artifact/blob/instance/network); most endpoints lack declarative validation. Residual: extend schema coverage across the remaining endpoints."*

**That residual was overtaken by phase 4 of [`PLAN-api-input-validation.md`](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-api-input-validation.md), by another route.** Phase 3 compiles *every* parameter declaration in the tree into a marshmallow schema, and phase 4 turned rejection on — so every endpoint now has declarative validation, without `use_kwargs` being extended to any of them. `@use_kwargs` survives on four handlers only because it is what gets a query parameter *to* a handler; the compiled path is check-only by design and deliberately does not inject anything.

The work that is actually left is what [#936](https://github.com/shakenfist/shakenfist/issues/936) described — *"Replace hand-rolled instance-create validation with a declarative schema"* — and #936 was closed as a duplicate of this issue on 2026-09-12. The body above has therefore been rewritten to that scope, so that an implementing agent sent here reads a description of the work that remains rather than a description of work that is finished. Nothing the original text asked for has been dropped; it was delivered by a different mechanism than the one it proposed.

Tracked by phase 7 of [`PLAN-api-input-validation.md`](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-api-input-validation.md).

mikalstill:
[...]

PR #3620, opened 2026-08-03
Plan declarative API input validation, and audit the parameter declarations

### LINKED ISSUE #3609: GET /nodes/<node>/events with limit returns 400 "'<=' not supported between instances of 'str' and 'int'"
`GET /nodes/<node>/events` returns a 400 when `limit` is supplied,
because the handler compares the raw query-string value — a `str` —
against an `int`.

Reproduction, taken from a real request on the sfcbr cluster
(request-id `0c158974-2104-4e1e-a528-afe5fe11d702`, 2026-07-31):

```
GET http://sfcbr.shakenfist.com/nodes/sf-1/events
body: {'limit': '5'}
->  400  {'error': "'<=' not supported between instances of 'str' and 'int'", 'status': 400}
```

The three log lines for that request-id show the request parsing and
authenticating normally (keyname `deploy`, method GET, path
`/nodes/sf-1/events`) and then failing in the handler:

```
08:43:04.879  shakenfist.eventlog  GET api request received
              extra: {'url': 'http://sfcbr.shakenfist.com/nodes/sf-1/events', 'body': "{'limit': '5'}"}
08:43:04.887  shakenfist.eventlog  token used to authenticate request
              extra: {'keyname': 'deploy', 'method': 'GET', 'path': '/nodes/sf-1/events', 'remote-address': '192.168.1.3'}
08:43:04.893  shakenfist.eventlog  GET api response sent
              extra: {'status': 400, 'body': '{\'error\': "\'<=\' not supported between instances of \'str\' and \'int\'", \'status\': 400}'}
```

`limit` arrives as the string `'5'` and is never coerced to `int` before
being range-checked, so the comparison raises `TypeError`, which is
caught and returned to the client as a generic 400. The fix is to coerce
(and validate) `limit` in the argument-parsing layer for this endpoint.

Two things make this worse than a plain 400:

- **The error is unattributable from the logs alone.** It surfaces at
  INFO level as part of a normal "response sent" record, with no
  traceback and no exception class, because the API layer routes it as a
  client error. Nothing in the log identifies it as an unhandled
  `TypeError` in server code. It only became visible to us because the
  message text leaked the Python type names into the response body.
- **Leaking interpreter-level type errors into API responses** is not
  great as an interface — a caller passing a perfectly reasonable
  `limit=5` gets told about Python `str` and `int`.

Observed 5 times on 2026-07-31 across sf-1/sf-2/sf-3/sf-4, all from the
`deploy` key.

**Loki query**
```
{job="shakenfist"} | json |= "not supported between instances"
```

**first_seen_sha:** `788bb6f`

mikalstill:
## Summary

Replaces the abandoned #3610 with a plan and its first two phases.

#3610 tried to fix #3609 — a `limit` sent as a string returning a 400
containing a Python type error — by hand-rolling guards at each site that
reads an integer out of a request body. Six review rounds later it was closed
unmerged: seven hand-written guards, two of them wrong on the first attempt,
and the defect class still not covered.

The better approach came out of that review. **124 of our 129 handler methods
(96%) already declare every parameter's name, location, type and required-ness
via `swagger_helper()`** — 254 parameters, read today by nothing but the
documentation generator. Making that load-bearing gets validation across the
API without authoring a single new schema.

## Planning

* `PLAN-api-input-validation.md` — the master plan, scanning all 83 open
  issues for what it should address. **Twelve**, including #528
  ("Broaden declarative type/validity checking across all API endpoints",
  open since 2020-11-11) and #936 (2021).
* `PLAN-api-input-validation-phase-00-decisions.md` — phase 0's measurements
  and the reasoning behind each decision.
* `PLAN-api-input-validation-phase-01-declaration-audit.md` — phase 1.

## Phase 0: research and decisions

The plan rested on one hypothesis — are those declarations accurate, or six
years of drift? Measured rather than assumed:

* **Names: 229 of 236 (97%) match a kwarg the handler actually accepts**, and
  no declared type contradicts a signature default. So: compile the
  declarations, don't author new ones.
* **Locations: 116 of the 119 path parameters are declared as something other
  than `path`.** Only 3 are right. That makes the audit a *precondition* for
  compiling, since location is what tells a parser where to look for a value.
* webargs, marshmallow and pydantic are all already pinned, so the library
  choice is fit, not cost — webargs, because it exists to parse Flask request
  arguments and is already used at four sites.
* Decorator placement verified by running `flask_restful`, not by reading the
  comment: index 0 of `method_decorators` is innermost, so validation lands
  after authentication and before the per-method decorators.
* The error shape can stay as it is: the official client never parses the
  message, and only 11 of our assertions touch error text.

## Phase 1: declaration audit

No behaviour changes — the declarations are still documentation-only when this
lands. What changes is that they become true, and a test keeps them that way.

* Invalid locations now raise `InvalidAPIDeclaration` at import time.
  `swagger_helper()` has always validated the type token and never the
  location, which is how `'post'` and `'qeury'` survived.
* 116 path parameters corrected, derived from `app.py`'s `add_resource()`
  calls by `tools/fix-api-parameter-locations.py` (committed, so the diff can
  be reviewed by re-running it rather than by reading 116 single-token edits).
* Five wrongly named parameters fixed. **`POST /instances` has been
  documenting `sshkey` and `userdata` while the handler reads `ssh_key` and
  `user_data`** — anyone following the spec to configure cloud-init got
  neither, silently.
* Seven undeclared parameters documented.
* A structural test over the AST and the route table pins all of it.

### What the test found that the audit could not

`InstanceSnapshotEndpoint` carries no `swag_from` at all, so both its methods
were absent from the published API — including `max_versions`, which #3610
found flows unvalidated into `Artifact.from_url()`. Every by-hand measurement
compared declarations *against* handlers, so an endpoint with no declaration
was invisible to all of them. Now declared.

### One deliberate deferral

The five metadata `delete` endpoints accept `value` and none read it. The
right fix is to stop accepting it, but removing it today makes a caller who
sends it receive `delete() got an unexpected keyword argument 'value'` as a
400 — exactly the leak this plan closes. Deferred to phase 4 and recorded in
`UNDECLARED_BY_DESIGN` with that reasoning.

## Verification

`pre-commit run --all-files` passes; 1978 unit tests pass. The audit test was
checked against deliberate breakages: changing a path parameter's location
fails one assertion, renaming it to something the handler cannot receive fails
two.

Fixes #3609


mikalstill:

mikalstill:
Round two addressed in c4e187106.

The one `fix` item was right and it was mine: the derivation rule from the last round fell through to `body` for `ClusterOperationsEndpoint`, which reads its target parameters from `flask.request.args` as a fallback behind the body merge. Fixed by adding the missing derivation source rather than the missing declaration -- a name read from the query string is a query parameter whatever else it also is -- and the two declarations were then rewritten by the script rather than by hand.

Also taken: `header`/`formData` declarations are now reported and left alone instead of being derivable-to-`body`; an unknown type token raises `InvalidAPIDeclaration` naming the endpoint and parameter rather than a bare `KeyError`; and the script's report mode now runs as a pre-commit hook on every commit touching `external_api/`, so "re-running it is the check that the tree still agrees" is enforced rather than advertised.

The multiple-body-parameter finding is recorded as a phase 2 work item rather than fixed here, along with the missing specification-validity test. Confirmed the counts: 23 operations on develop, 29 on this branch. Collapsing them into one `body` parameter with a generated `schema` is a change to the renderer, not to any declaration.

Not taken, and why: per-class vs per-method query derivation, `open()` without context managers, and `assert` under `-O` are all unreachable on the current tree.

mikalstill:
Round three addressed in ae8a2cf61.

**Item 1** was the important one, and the mutation test proves it: `event_type` declared `query` passed every assertion while the script flagged it. Both halves of that are now closed. The derivation moved into `shakenfist/external_api/declarations.py`, which the script and the test both import, and the test asserts `audit()` returns nothing — the same question the script asks, so what one would rewrite is what the other fails on. I reproduced your exact mutation afterwards and the test now fails naming the endpoint, the parameter and where it really arrives.

That also resolves **item 2** as you predicted: one implementation, no "keep the two in step" obligation, and the `_literal` divergence you spotted is gone. `DerivationTestCase` covers the four sources on constructed sources, since the defects each was added for are by definition no longer in the tree.

**Item 3** fixed — `operation_type` is a registry key, now declared `string` with its own description. You are right that this is more than a documentation nit: compiled as written, a legitimate `net_op` would be rejected as a malformed UUID.

**Item 4** taken rather than deferred, since the derivation was being moved anyway: the webargs source is now per-method and reads the `use_kwargs` `location` keyword, resolving the schema argument it names. The docstring and the code now describe the same rule.

**Item 5** — you are right, and I checked: this module imports `base.py` and the test base, neither of which reaches an endpoint module, and the AST tests never execute a declaration. Reworded to the real reason.

**Item 6** — restated the criterion as a property enforced by a test, with the counts attributed to `b844fd98e` where they were measured.

**Item 8** taken, both halves: `bearer` is no longer declarable (with a test), and the duplicated `type` key on `integer` is now `format`.

**Item 7** filed as #3626 so the `openapi_spec_validator` test is not gated on the whole of phase 2. Your framing is the reason — it would have quantified the 23 → 29 regression automatically instead of by hand.

mikalstill:
Round four addressed in a9777d26c. All six actionable items taken.

**Item 1** was right and my last fix was half a fix: I corrected handler attribution and left name resolution alone. Reproduced your check — `A.get` did see `{'alpha', 'beta'}`. Resolution is now innermost-scope-first, and a scope contributes only its own assignments rather than everything nested inside it, so a class-level schema shadows a module-level one and a sibling class's is invisible. Same source as your repro, now asserted in `DerivationTestCase` both ways round, plus a case proving the module-level fallback still works.

**Item 3** turned up something while I was widening it. Asking "does this handler carry a declaration" instead of "does this handler with parameters declare them" first flagged eleven handlers, not three — because eight endpoints correctly declare an *empty* parameter list, which is a different state from carrying no declaration. So the guard now asks the former question via `declarations.documented()`, and `UNDOCUMENTED_BY_DESIGN` holds exactly the three health probes you identified. Worth noting the widening is real coverage rather than tidiness: `AdminLocksEndpoint.get` takes no kwargs, so the old test could never have flagged it if its `swag_from` vanished.

**Item 2** taken — the rewrite path is tested on a constructed module with two declarations on one physical line, which is the shape the bottom-up ordering exists for and which nothing in the tree has. `audit()` and `handlers()` now take a directory so the test can point them at a tmpdir.

**Item 5** taken: arity is checked before destructuring, in both `swagger_helper()` and `declarations()`.

**Item 6** taken: a `Resource` base is required. I checked first — all 90 current handler classes derive from `api_base.Resource` directly, so nothing is dropped by the filter.

**Item 4** — I took the rename rather than threading endpoint identity through 254 call sites. The test now says `section`, and the docstring says why that is all `swagger_helper()` can know.

**Item 7** recorded in the phase 1 outcome, with your reasoning: an undocumented spec-content change inside a mechanical audit is what makes a later bisect confusing.

mikalstill:
Round five addressed in 5885b5dd0. All items taken.

**Item 1** was right, and it is the fourth round running where the defect was in machinery added the round before. `documented()` was introduced *last* round for exactly this distinction and I then failed to use it in the assertion next door.

So rather than fix the instance I closed the class it belongs to, which is also item 2. Every source in `declarations.py` answered "not found" and "cannot read this" with the same empty set, which makes a skipped input a confident wrong answer rather than a missing one. `audit()` now returns a third value, `problems` — a non-literal `add_resource` argument, an unresolvable `use_kwargs` schema, a class subclassing another endpoint. The fixer refuses to rewrite when it is non-empty; the test asserts it *before* the drift list.

That ordering was not my idea. I wrote `tools/check-api-declaration-guards.sh`, which breaks each property the audit claims and confirms the guard fires — ten mutations. With the problems assertion second, an unreadable route reported ``blob_uuid is declared path but arrives in the body``, which would have sent a reader to change a correct declaration. The mutation pass found that; reading the test did not.

**Item 5** verified rather than taken on faith, and you are right: `API_ADVERTISED_HTTP_SCHEMES` is typed `str` while its own description calls it a space separated list, so a default deployment publishes `schemes: 'http'` and a two-scheme one publishes a single nonsense string. Recorded against #3626, since it has to be fixed or waived before the validator reaches anything else.

**Items 3, 4 and 6** taken: stale status line, phase 0 linked from both tables, and `declarations.py` annotated and added to the mypy rollout. mypy earned its place immediately — it rejected `list[ast.AST]` for the scope list, since only `Module` and `ClassDef` carry the `.body` that innermost-first resolution reads.

**Item 7** is a good catch and I have left the inconsistency alone deliberately: the declarations now accurately document it, which is what this phase is for. Noted against phase 4, which already plans to fold those webargs schemas into the compiled path.

On the rounds themselves: the pattern is not scope creep, it is incomplete generalisation — a finding gets fixed by adding a rule, the rule has an untested edge, the next round finds the edge. The mutation script is the durable answer, and it is checked in rather than kept as a one-off.

mikalstill:

Round six addressed. The `fix` (item 2) was real and I have independent evidence for it: rewriting `check()` to name its verdict rather than infer it from the absence of `Failed: 0` immediately exposed two of the ten mutations proving nothing — the unreadable-route one was a `NameError` on an undefined `BLOB_ROUTES`, and a class-name-collision mutation added in the same round turns out to be inexpressible in this tree, because `flask_restful` derives its endpoint name from the class name and refuses the second registration outright. The harness now checks the interpreter and the baseline first, diffs against the backup to confirm each mutation applied, and exits non-zero.

Item 3 is done: `docs/developer_guide/writing_an_endpoint.md`, with summaries in CLAUDE.md and AGENTS.md, and the stale hooks list refreshed.

Items 4, 5 and 6 taken as suggested. Item 1 is filed as #3629 (confirmed against `webargs/core.py:529`) and referenced from the phase 3 row, since decision D6 owns the general fix.

An adversarial pass on the fixes then found three more, all the same shape as the ones review rounds have been finding — a source the derivation cannot read answering with the same empty set as one with nothing in it:

- `add_resource()` whose first argument is not a plain name or attribute names no class the audit can match, silently emptying that class's route set; with no arguments at all it raised `IndexError` out of `audit()`.
- a webargs schema bound at a tuple of locations was read as "not query", so its keys derived to `body`. That is exactly the shape #3629's fix and D6's fallback introduce, so the audit would have rewritten the declarations that fix had just made true.
- a `location` keyword that is present but unreadable was indistinguishable from an absent one, which means webargs' json default.

Test coverage gaps from the review closed as well: `main()`'s refusal path is now exercised (asserting the file is untouched, not merely that it exits), and the class-name collision has a constructed-source test.

mikalstill:

mikalstill:

mikalstill:

mikalstill:

Since the last review: request_args_parameters() now has the problems channel (with negative coverage for all four unreadable forms), the collision-impossibility claim is corrected and the collision is now mutation 12, check() consults test counts before the exception-name grep, the checkout-vs-installed-wheel invariant is asserted, integer parameters render int64, and the branch has been rebased onto current develop (the audit now covers the eleven new federation handlers cleanly).

mikalstill:

Since the last review: every open() in the derivation and fixer now passes encoding=utf-8 (your LC_ALL=C reproduction runs clean, and the fixer can no longer truncate on the write side), test_the_audit_reads_this_checkout now asserts a checkout-only marker so the both-from-the-wheel case fails, the fixer splices at byte offsets with a test constructing the non-ASCII-before-literal line, route coverage is per handler, an endpoint/route symmetry test guards handlers() in both directions, the harness restores with rsync --delete, and the security-as-array fix shipped (spec errors 241 -> 129; carry list and #3626 updated).

PR #3621, opened 2026-08-03
Use the shared automated review workflow

mikalstill:
The automatic PR review has until now been an "automated_reviewer" job
written out in full in this repository's CI workflow, copied from a
snippet in the development repository's README. Eight projects each
carried a copy and they had drifted from one another. None of them set
timeout-minutes, so a wedged Claude Code run could hold a scarce
claude-code runner for GitHub's six hour default.

The job body now lives once, in
shakenfist/actions/.github/workflows/pr-auto-review.yml. What stays
here is the part which is genuinely per-project: the needs: list
naming this project's test jobs, which is what gates the review on CI
passing. A job skipped because a dependency failed never starts the
reusable workflow.

Adopting the shared workflow brings this project:

* A 60 minute timeout on the review. Real reviews finish in three to
  six minutes.
* The same-repository restriction. The reviewer runs Claude Code with
  --dangerously-skip-permissions while holding a token with
  pull-requests: write and issues: write, and is fed the PR diff,
  which is untrusted input, so a fork PR would put an attacker's text
  in front of a write-capable token.
* A bot-commit check which also recognises noreply@anthropic.com, the
  address a Claude Code commit made on a runner is authored as. Only
  clingwrap's copy caught that address before.

The check_bot_commit job stays, because automated_delinter also
consumes its is_bot output. It is no longer wired to the reviewer.

automated_delinter is the other unbounded Claude Code job here, so it
gains a 60 minute timeout for the same reason.

While here, bring the rest of the Claude Code workflow timeouts into line
with the templates: 60 minutes on the re-review, 10 on the address-comments
trigger and 10 on the fix-tests trigger.

Validated with each project's own `pre-commit` on the changed files, and by diffing actionlint findings before and after so pre-existing noise is not mistaken for a regression.

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -223,6 +223,10 @@ jobs:
   automated_delinter:
     name: "Automated delinter"
     runs-on: [self-hosted, claude-code]
+    # Without this the job inherits the six hour GitHub default, so a
+    # wedged Claude run would hold a scarce claude-code runner for the
+    # rest of the day.
+    timeout-minutes: 60
     needs: [sanity_checks, check_bot_commit]
     # The same-repo check keeps fork PRs from running their own copy of the
     # delinter script on the credential-bearing claude-code runners.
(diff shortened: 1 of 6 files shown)


mikalstill:

PR #3635, opened 2026-08-05
Stop unit tests leaking tempdirs in /tmp.

mikalstill:
## Summary

The `_SpoolRootMixin` classes in the eventlog and logship unit test files create a tempdir with `tempfile.mkdtemp()` to redirect `SPOOL_ROOT`, and register cleanups to reset the spool singleton and restore `SPOOL_ROOT` — but never register one to delete the directory. Every test therefore leaked a directory into `/tmp`; over 26,000 had accumulated on the development machine across the prefixes `sf-logship-test-`, `sf-logship-spool-test-`, `sf-logship-drainer-test-`, `sf-eventlog-test-`, `sf-spool-test-` and `sf-drainer-test-`. The logship tests inherited the bug verbatim when they were modelled on the eventlog ones.

This registers `shutil.rmtree(..., ignore_errors=True)` as the first cleanup in each of the six affected mixins. Cleanups run LIFO, so registering it immediately after `mkdtemp` means it fires last, after `reset_for_tests()` has closed the sqlite spool handles.

## Testing

- Ran all six affected test modules via tox: all pass.
- Counted matching directories in `/tmp` before and after the run: unchanged, so nothing new leaks.
- `pre-commit run --all-files` passes.
- Audited the rest of the test tree for the same pattern: all other `mkdtemp` users already register cleanups, and the remainder use `tempfile.TemporaryDirectory`.

### DIFF
--- a/shakenfist/tests/base.py
+++ b/shakenfist/tests/base.py
@@ -1,3 +1,5 @@
+import shutil
+import tempfile
 from unittest import mock
 
 import testtools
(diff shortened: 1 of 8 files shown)


mikalstill:

Since the last review: took item 2 (the six copy-pasted `_SpoolRootMixin` classes are now a single parametrised `SpoolRootMixin` in `shakenfist/tests/base.py`, with the cleanup-ordering comment from item 1 in the one shared copy) and item 5 (`test_util_exceptions.py` now imports `shutil` at the top and uses `addCleanup` instead of the four rmtree-only `tearDown` methods). Item 7 was skipped for the reasons the review itself gave.

PR #3659, opened 2026-08-07
Stringify uuids in the DELETE /artifacts response

### LINKED ISSUE #3656: DELETE /artifacts 500s with "Object of type UUID is not JSON serializable" after deleting the artifacts, breaking namespace delete
## Summary

`DELETE /artifacts` (bulk delete of a namespace's artifacts, used by `sf-client namespace delete` during namespace cleanup) returns a 500 whenever it actually deletes at least one artifact. The handler completes all the deletions, then crashes serializing the response:

```
TypeError: Object of type UUID is not JSON serializable
```

Because the deletions commit before the response is built, the operation succeeds server-side but the client sees a 500 and aborts — so a namespace delete fails partway, leaving the namespace record in place. An immediate retry finds zero artifacts, returns `[]`, and succeeds, which makes the failure look intermittent.

## Observed (production sfcbr cluster, 2026-08-08)

```
Deleting namespace k3s-test-002
2026-08-08 06:57:45.450688 ERROR: Internal Server Error: {"message": "Internal Server Error"}
```

Request-id `158768a1-2e58-4f8e-b0e3-0a2423f8aefb`: `DELETE /artifacts` with `{'confirm': True, 'namespace': 'k3s-test-002'}` deleted artifact `254cafee-f37c-4544-88cd-84e1e7141dfa` at 06:57:45.422, then logged:

```
Exception on /artifacts [DELETE]
  ...
  File ".../flask_restful/representations/json.py", line 21, in output_json
    dumped = dumps(data, **settings) + "\n"
  ...
TypeError: Object of type UUID is not JSON serializable
```

## Root cause

`ArtifactsEndpoint.delete()` in `shakenfist/external_api/artifact.py` builds its response with `deleted.append(a.uuid)`. Since 95974afc9 ("Store UUIDs as uuid.UUID objects in BaseObject.", 2025-12-26), `BaseObject.uuid` returns a raw `uuid.UUID`, which flask_restful's JSON representation cannot serialize.

This is the same class of bug as #3573 (raw `uuid.UUID` reaching a `json.dumps` boundary), but in the API response path rather than the log emitter. The codebase's serialization boundary for API responses is the Pydantic external-view models (`schema/external_view.py`, whose `BaseExternalView` has a `field_serializer` stringifying `uuid`) — this handler bypasses that by returning raw object attributes directly.

A sweep of `shakenfist/external_api/` shows this is the only response path returning raw `uuid.UUID` values; the similar appends in `auth.py`'s namespace delete only feed log fields and error counts.

## Fix

Stringify the uuids in the response list (matching the documented response schema, "A list of artifact uuids"), and add a regression test that exercises `DELETE /artifacts` through the flask test client so the real flask_restful serialization path is covered.

mikalstill:
## Summary

Fixes #3656.

Since 95974afc9 ("Store UUIDs as uuid.UUID objects in BaseObject."), `ArtifactsEndpoint.delete()` has returned a list of raw `uuid.UUID` objects. This response does not pass through an external view, so nothing stringifies them and flask_restful's JSON encoder raises `TypeError: Object of type UUID is not JSON serializable`. The deletions commit before the response is built, so every bulk artifact delete that actually deletes something 500s *after* doing the work — which aborts `sf-client namespace delete` partway through cleanup (observed on the sfcbr production cluster, 2026-08-08).

## Changes

- Stringify the uuids when building the `DELETE /artifacts` response, matching the documented response schema ("A list of artifact uuids") and the serialization the external-view models apply everywhere else.
- Add `shakenfist/tests/external_api/test_artifact_bulk_delete.py`: three regression tests exercising the route through the flask test client so the real flask_restful representation runs. Asserting on the handler's return value alone would not have caught this; the main test reproduces the exact production `TypeError` against the unfixed handler.

## Testing

- New tests pass under tox; the primary test fails with the production traceback when the fix is reverted.
- `pre-commit run --all-files` clean (flake8, full unit suite, namespace-scoping and auth-default consistency checks, mypy).

A sweep of `shakenfist/external_api/` found no other response path returning raw `uuid.UUID` values — the similar appends in `auth.py`'s namespace delete handler only feed log fields and error counts.

### DIFF
--- a/shakenfist/external_api/artifact.py
+++ b/shakenfist/external_api/artifact.py
@@ -521,7 +521,10 @@ class ArtifactsEndpoint(api_base.Resource):
         for a in Artifacts(namespace=namespace):
             a.add_event(EVENT_TYPE_AUDIT, 'deletion request from REST API')
             a.delete()
-            deleted.append(a.uuid)
+            # The uuid is stringified here: this response does not pass
+            # through an external view, and flask_restful's JSON encoder
+            # cannot serialize a raw uuid.UUID (issue 3656).
+            deleted.append(str(a.uuid))
 
         return deleted
 
--- a/shakenfist/tests/external_api/test_artifact_bulk_delete.py
+++ b/shakenfist/tests/external_api/test_artifact_bulk_delete.py
@@ -1,0 +1,46 @@
+"""The bulk `DELETE /artifacts` route must return a serializable body.
+
+The handler returns a plain list of the deleted artifact uuids rather
+than an external view, so nothing between it and flask_restful's JSON
+encoder stringifies them. When `BaseObject.uuid` became a `uuid.UUID`
+the deletions kept committing but the response crashed with
+`TypeError: Object of type UUID is not JSON serializable`, failing
+namespace deletes after their artifacts were already gone (issue 3656).
+
+These tests go through the flask test client so the real flask_restful
+representation runs -- asserting on the handler's return value alone
+would not have caught this.
+"""
+
+import json
+
+from shakenfist.artifact import Artifact
+from shakenfist.tests.external_api.test_artifact_access import (
+    ArtifactAccessFixture)
+
+
+class ArtifactBulkDeleteTestCase(ArtifactAccessFixture):
+    def _bulk_delete(self, namespace, body):
+        return self.client.delete(
+            '/artifacts', data=json.dumps(body),
+            headers={'Authorization': self._token(namespace)})
+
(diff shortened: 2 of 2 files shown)


mikalstill:
Superseded by #3667 (merged as d12cd6db7), which lands the identical `str(a.uuid)` fix with equivalent unit tests plus a functional CI test for the bulk delete path. Closing rather than re-queueing to stop the two duplicate fixes racing each other through the merge queue.

Issue #3670, opened 2026-08-08
test_artifact_show races suite concurrency for node CPU headroom (507 sufficient_idle_cpu)


mikalstill:
## What happened

`test_commandline_artifacts.TestArtifactCommandLine.test_artifact_show` failed in PR #3659's merge queue run on 2026-08-08:
https://github.com/shakenfist/shakenfist/actions/runs/31236340890
(job "Debian 12 tier (collection) / Smoke tests (collection)", step "Run functional tests").

The second instance create in the test (`test_commandline_artifacts.py:171`) was refused by the scheduler:

```
shakenfist_client.apiclient.InsufficientResourcesException: ('API request failed', 'POST',
    'http://localhost:13000/instances', 507,
    '{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}')
```

## Why systemic

This is the same 507 `sufficient_idle_cpu` capacity race already tracked per-test as #3602 (`test_disappearing_source_instance`) and #3498 (`test_network_plumbing_lifecycle`) — under suite concurrency the nested cluster transiently has no node with enough idle CPU headroom, and whichever test loses the race fails at `create_instance`. `test_artifact_show` is simply a new member of the family; this issue exists so the signature is findable by test name in future triage. The same run failed `test_federation` for an unrelated (since-fixed, #3652) reason, and PR #3659 touches only the DELETE /artifacts response serialization, which no part of this test exercises.

The underlying scheduler behaviour is discussed in #3636 and the scheduler-reservations plan (`docs/plans/PLAN-scheduler-reservations.md`).

Found during merge CI triage of PR #3659. *(Triage assisted by Claude Code)*

mikalstill:
First recurrence since filing, in PR #3685's merge queue run on 2026-08-09:

- [run 31339411465, job 93310792709](https://github.com/shakenfist/shakenfist/actions/runs/31339411465/job/93310792709) (00:16)
- Topology: "Debian 12 cluster" (the original report was a different topology, so this is not topology-specific).

```
File ".../cluster_ci_tests/test_commandline_artifacts.py", line 171, in test_artifact_show
    inst2 = self.test_client.create_instance(
shakenfist_client.apiclient.InsufficientResourcesException: ('API request failed', 'POST',
'http://localhost:13000/instances', 507,
'{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}')
```

[...]

Issue #3669, opened 2026-08-08
sf_instance dirtiness check false-positives on server-resolved blob_uuid in disk_spec (ansible module CI scenario 005)


mikalstill:
## What happened

Ansible module CI scenario `005.yml` (instance idempotency) failed in two of PR #3666's merge queue runs on 2026-08-08:

- https://github.com/shakenfist/shakenfist/actions/runs/31249876181 (job "Ansible modules (collection) / Smoke tests (collection)")
- https://github.com/shakenfist/shakenfist/actions/runs/31255571207 (same job)

The "Assert the instance is not new" task fails because the second `sf_instance` call with an identical spec reports `changed: true` — the module decided the instance was dirty and deleted/recreated it:

```
Instance dirty: disk specification has changed
    Requested: [{"base": "debian:11", "bus": null, "size": 10, "type": "disk"}]
    Existing:  [{"base": "debian:11", "blob_uuid": "73b4e968-7276-4b10-a03e-1258bcb930f1", "bus": null, "size": 10, "type": "disk"}]
```

## Root cause

`deploy/collection/plugins/modules/sf_instance.py` (the disk_spec dirtiness comparison, around lines 359-373) strips the internal `disk_base` key from the existing disk spec before comparing, but not `blob_uuid`. The server resolves `blob_uuid` into the stored `disk_spec` once the image fetch completes, so the comparison is timing-dependent: if the second module invocation lands before resolution the specs compare equal and the scenario passes; after resolution the extra `blob_uuid` key makes an identical request look dirty, and the module tears down and recreates the instance. In the failing runs the instance was still in `preflight` (23s after create) but `blob_uuid` had already been resolved.

This is another instance of the dormant-dirtiness-check family activated by the native collection module port (see #3308 for a sibling bug in networkspec float parsing).

## Why systemic, not PR-caused

PR #3666 only touches OpenAPI spec validation (tests, docs, a validator in `external_api/app.py`); it does not touch the collection modules or the instance API. The same scenario passed in the same PR's earlier merge run (https://github.com/shakenfist/shakenfist/actions/runs/31248151422), confirming the failure is a race, not a deterministic regression.

## Fix direction

Strip server-populated keys (`blob_uuid`, alongside the existing `disk_base` cleanup) from the existing disk spec before the JSON comparison in `sf_instance.py`.

Found during merge CI triage of PR #3666. *(Triage assisted by Claude Code)*

mikalstill:
[...]

Issue #3498, opened 2026-07-25
test_network_plumbing_lifecycle races suite concurrency for node CPU headroom (507 sufficient_idle_cpu)


mikalstill:
## Symptom

`shakenfist_ci.cluster_ci_tests.test_network_lifecycle.TestNetworkPlumbingLifecycle.test_network_plumbing_lifecycle` failed two merge CI runs in the last week with the same signature, both on the "Debian 12 tier" topology:

- [run 29794604996](https://github.com/shakenfist/shakenfist/actions/runs/29794604996/job/88523314661) (2026-07-21)
- [run 29623102870](https://github.com/shakenfist/shakenfist/actions/runs/29623102870/job/88021996332) (2026-07-18)

```
shakenfist_client.apiclient.InsufficientResourcesException: ... 507,
'{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}'
```

The failure is on `_create_instance_on(node_a)` -- a create targeted at a specific hypervisor -- while the suite runs with stestr concurrency 5 on the slim-tier topology, which has roughly half slim-primary's hypervisor capacity. Other tests' instances occupy the target node's CPU headroom, the targeted create gets a 507, and the whole merge run fails.

## Assessment

Given a targeted create and a genuinely full node, a 507 is arguably *correct* server behaviour -- this is primarily a test-vs-capacity race, and it is the single most frequent functional-test failure in merge CI this week (the wrong-node variant of the same test is tracked separately as a real placement bug).

## Options

- Make `_create_instance_on` retry with backoff on 507 (capacity is transient -- other tests' instances delete within minutes).
- Serialise the placement-sensitive lifecycle tests (stestr group or a lock) so they do not compete with the full concurrency-5 suite for node headroom.
- Reserve headroom: size slim-tier hypervisors so that a single targeted create always fits.

Found during a triage of merge CI failures for the week of 2026-07-18 to 2026-07-25.

mikalstill:
This bit us again on 2026-07-30, in the merge queue run for PR #3563:
https://github.com/shakenfist/shakenfist/actions/runs/30512428320
(job "Debian 12 tier (collection) / Smoke tests (collection)").

Same signature as reported: the second `_create_instance_on()` call in
`test_network_plumbing_lifecycle` failed with

```
InsufficientResourcesException: ('API request failed', 'POST',
'http://localhost:13000/instances', 507,
[...]

PR #3685, opened 2026-08-09
Add typed bounds to the API parameter vocabulary

### LINKED ISSUE #3642: API declaration audit: variadic handlers pass the accepted-parameters check vacuously
From the twelfth automated review round on PR #3620 (item 3, consider).

`declarations.handler_kwargs()` reads only `fn.args.args` and `fn.args.kwonlyargs`; `fn.args.vararg` and `fn.args.kwarg` are neither consulted nor reported. A handler written as `def post(self, instance_ref=None, **kwargs)` accepts arbitrary body keys (log_request merges the whole JSON body into kwargs), while `test_accepted_parameters_are_declared` iterates a near-empty list and passes. This is the "not found and cannot read this share one answer" defect shape the PR eliminated from every other derivation source.

Latent: no handler in `external_api/` is variadic today. But rule 5 of `docs/developer_guide/writing_an_endpoint.md` states the guarantee unconditionally, and phase 3 of PLAN-api-input-validation compiles validation from this same set.

Suggested fix: give `handler_kwargs()` an optional `problems` list like its siblings and append an entry when `vararg` or `kwarg` is set, so `audit()` refuses to proceed rather than reporting a clean tree; pin with a constructed-source DerivationTestCase case, since by definition the shape is not in the tree.

mikalstill:
Part 3 of 3 for phase 2 of `PLAN-api-input-validation`, following #3666 and
#3682. This completes the phase.

## What this is

Decision D9 from phase 0: the type vocabulary was too coarse to express the
bounds the issue list needs, so those bounds lived only in server code. The
events `limit` cap of 1000 has been enforced for years and was invisible to
every caller reading the published API.

**New type tokens** in `swagger_helper()`:

* `unsignedinteger` → `minimum: 0`. Applied to `max_versions` on the three
  endpoints declaring it — a negative value there is silently destructive,
  since `delete_old_versions()` slices `[:-max]`, so `-1` deletes the oldest
  version on every index add — plus blob read `offset`/`limit` and the upload
  truncate `offset`.
* `base64` → `format: byte`, Swagger 2.0's standard token for base64 content.
  Applied to instance create `user_data`.
* `macaddr` and `netblock` render with published validation patterns.
  `netblock` is applied to network create. `macaddr` has **no declaration site
  today** — interface MACs arrive inside the network `arrayofdict` specs — so
  it is defined here and its first consumer is the structured value types work
  (#936). Flagging that deliberately rather than quietly shipping an unused
  token.
* `arrayofstring` and `arrayofdict` become real array types with `items`,
  which only became legal once #3682 made body parameters render through
  schema objects. Generators now produce list-typed bindings instead of
  strings.

**An optional sixth tuple element**: a constraints dict with keys from
`minimum`/`maximum`/`pattern`, all valid Swagger 2.0 keywords, so a bound
renders into the published specification. Validated at import time in the
established style — unknown keys, non-number bounds, bounds on non-numeric
types, contradictory bounds, patterns that don't compile or sit on non-string
types, and a constraint restating a key its own token already renders all
raise `InvalidAPIDeclaration`. The five events `limit` declarations now carry
`{'minimum': 1, 'maximum': 1000}` and say so in their descriptions.

Note the published minimum is 1, not 0: the server replaces a `limit` of zero
or less with the default of 100, so 0 is not actually a rejected value. Whether
to keep that behaviour or start rejecting it is a phase 3 question, and the
declaration describes today's server rather than pre-empting it.

**Constraints are documentation only** until phase 3 compiles them into
request validation. Nothing about what the server accepts changes here.

## The #3642 rider

The audit machinery has to follow the tuple change, and while it was open:
`handler_kwargs()` gains the problems channel its siblings already have. A
variadic handler accepts names no enumeration can produce — `log_request`
merges the whole JSON body into kwargs — so it is now reported rather than
returned as a near-empty list that the accepted-parameters assertion iterates
vacuously. That is the "derivation sources must distinguish *not found* from
*cannot read this*" invariant, which is what #3642 is about. Pinned on
constructed source, since no handler in the tree is variadic.

## Verification

* `tools/check-api-declaration-guards.sh` grows from 12 to 16 mutations; all
  four new constraint guards report `caught-import`. **All 16 caught.**
* That harness's `NO-OP` verdict earned its keep here: retyping the blob
  `offset`/`limit` declarations moved them out from under mutation 3's search
  text, and the harness reported `NO-OP` rather than a false catch.
* Adversarially the other way: with the `declarations.py` changes stashed, six
  tests fail.
* The generated specification still validates with zero errors, so #3666's
  assertion holds.
* Full unit suite: 2591 passed, 0 failed. `pre-commit run --all-files` clean.

`Fixes #3642`

https://claude.ai/code/session_012nPqvWFVKNtS63aNm74kRq


mikalstill:
Pushed `4d849ee` addressing the review. No item was `action: fix`; these are the ones worth taking rather than filing.

**Taken:**

- **4 (document)** — release note now covers both generator-visible changes: array parameters binding as lists rather than strings, and `format: byte` potentially making a generator base64-encode `user_data` that callers already encode themselves.
- **2** — `netblock` loses the pattern and becomes format-only, rather than recording the IPv4 narrowing as a deviation. The pattern under-accepted (no IPv6, which `ip_network()` parses) *and* over-accepted (`999.999.999.999/99`), so it earned nothing as validation while setting phase 4 up to compile a documentation commit into a 400. `ip_network()` stays the single source of truth.
- **3** — the array-of-objects-outside-the-body case is now refused at import time, so the "every declaration defect stops sf-api starting" invariant has no exception. Mutation 17 retypes the instance `disk` declaration to a query location: `caught-import`. Query rather than path deliberately — a path location would also trip the route-symmetry check, and a mutation that trips two guards does not prove the one it was written for.
- **1** — `unsignedinteger` applied to `cpus`, `memory`, `key_ttl`, `version_id` and console `length`. Left at minimum 0 rather than 1, for the same reason as the netblock decision: nothing rejects a zero or negative `cpus` today, so 0 is the strongest claim the server actually backs. Tightening is a phase 3 warn-only question, recorded in the plan's Outcome.
- **5, 6, 7** — positive test for the six-element parse, the six-element case moved from the arity test to the constraints test, `subTest` on both loops, two stale docstrings, and the `constraints` local renamed to `access_notes`.

**Not taken:** 8 and 9 were informational. 8's negative-offset 500 is noted in the plan as a concrete case for phase 3's warn-only rollout to confirm.

Verification: all 17 mutations caught, 2593 unit tests pass, `pre-commit run --all-files` clean.

Worth flagging separately: this PR's five merge queue attempts all failed on known systemic issues, not on anything in the diff — #3669 every time (deterministic, fix in #3686), #3498 three times and #3670 once (507 `sufficient_idle_cpu`), and one attempt lost both cluster runners to under-cloud starvation, now filed as #3696. PR #3684, a docs-only change, shows the identical signature. Re-queueing before #3686 lands will just reproduce it.

mikalstill:
Pushed `3d52054`. The `fix` item was real and is fixed; four of the five considers taken.

**1 (fix) — instance create `metadata`.** Confirmed: the handler answers 400 to anything that is not a mapping and then iterates `.items()`, and the functional suite posts a dict. Retyped to `dict`. This is the netblock trap arrived at from the other direction — a declaration that was harmless prose until this PR made it machine-readable, and which phase 3 would have compiled into rejecting the only shape that works.

**2 — `dict` now renders `{'type': 'object'}`.** Taken, as you say it is the natural companion: the argument for promoting the array tokens applies unchanged. Corrects `video`, `bound_claims` and the interface hotplug `network` too. The non-body guard widened to refuse bare objects as well as arrays of them, since that is the same defect without the wrapper.

Worth calling out how item 1 is guarded, because the obvious answer does not work: both tokens are legal in a body, so no import-time check can distinguish them — only the endpoint's semantics do. The assertion therefore lives in `test_openapi_spec.py` against the published specification. **That exposed a real gap in the harness**: mutation 20 (metadata retyped back to an array) came back `NOT CAUGHT`, because `run()` only ever ran `test_parameter_declarations`. The harness now runs both modules, and all 20 mutations are caught.

**3 — `audit()` now calls `handler_kwargs()` for its problems**, rather than documenting the channel as test-only. You were right that the fixer's "refusing to derive from input which could not be read" guarantee was incomplete. Verified by making a real handler variadic: the fixer now prints `post is variadic (**kwargs), so the parameters it accepts cannot be enumerated` and refuses to rewrite.

**4 — fractional bounds on integer types rejected**, same rationale as the existing bool exclusion. Mutation 19, `caught-import`.

**5 — `argtypes` hoisted to module-level `ARGTYPES`**, and the docs now point at that name rather than at a function-local the reader cannot import.

**6 and 7** were informational. 6's `limit=0` question is already recorded in the plan's Outcome as a phase 3 decision.

Verification: all 20 mutations caught, 2595 unit tests pass, `pre-commit run --all-files` clean.

mikalstill:
Pushed `3a48912`. The `fix` was real and is the same class as last round's, which changed how I responded to it.

**1 (fix) — console `length`.** Confirmed: `-1` is a supported sentinel that `get_console_data()` special-cases, and the functional suite calls it directly (`base.py:119`, `test_console_log.py:71`). My `unsignedinteger` retype published `minimum: 0` over a value the CI itself depends on — publishing the API as narrower than it is, which is precisely what the netblock rationale in this same PR argues against. Reverted to `integer`, and the sentinel is now documented in the published spec, where it never was.

**2 — the mechanism, which is the real answer here.** Two instances of one class in two consecutive rounds, both caught by a human-equivalent reader and neither by a machine, says the class needed a mechanism rather than a third spot fix. `test_openapi_spec.STRUCTURED_PARAMETERS` now lists every parameter publishing a structure or a bound alongside the shape its handler actually accepts, checked against the served specification. An entry with **no** `minimum` asserts that none is published — that is the half which catches the console-length direction, and it is the half a naive pin would have missed.

Adversarially verified: with `metadata` retyped back to an array and `length` back to unsigned, the table fails naming both, with the messages `post /instances metadata: type is 'array', expected 'object'` and `get .../consoledata length publishes minimum 0, which this table does not expect`. Mutations 21 and 22 make that permanent; all 22 caught.

I took your (a) rather than (b) deliberately. A handler-body scan would not have caught this one — the `-1` comparison lives in `Instance.get_console_data()`, not in the handler — so (b) would have bought a false sense of coverage. The table is a registry, not a derivation: it constrains the next author to think rather than proving them right. Real type derivation is phase 3's problem, where warn-only measures declarations against live traffic.

**8 — taken, and it is a behaviour change.** The comment named a live data-loss foot-gun, so shipping the commit that documents it while leaving it open was the wrong trade. Negative `max_versions` now 400s, and the handler catches `TypeError` as well as `ValueError` so a list body value is a 400 rather than a 500. Called out in the release note as the one behaviour change in an otherwise documentation-only phase.

**3** deep copies (the nested `items` alias was newly introduced by this PR, as you note), **7** rejects Python-only regex constructs with a note that JSON Schema `pattern` is unanchored, **11** annotations, **5** CLAUDE.md's three missing rules plus the "publish what the server backs" rule from **4**, **6** the second `network` parameter named in the release note.

**9 and 10** were informational; `macaddr`'s lack of a declaration site is recorded in the plan's Outcome.

Verification: all 22 mutations caught, 2597 unit tests pass, `pre-commit run --all-files` clean.

mikalstill:
Pushed `8121df9`. All three `fix` items were real, and two of them are the same shape as each other: a bound published in one place while the code it describes lives in three places, or in none.

**1 (fix) — `max_versions` closed only one of three write paths.** Confirmed exactly as described: `LabelEndpoint.post` and `InstanceSnapshotEndpoint.post` both hand a caller's value to `Artifact.new()`, which reaches the same setter and the same `delete_old_versions()`. Took the second suggestion rather than the first, because the setter is the thing all three routes actually reach: the coercion and the check now live in `artifact.validated_max_versions()`, called from the setter so **any** writer inherits the guard, and from all three handlers so the refusal is a 400 rather than a 500. New tests cover the function, the setter, and both newly guarded routes end to end.

**2 (fix) — `key_ttl` published a value the server 400s.** Confirmed: `validate_key_ttl()` refuses `<= 0` and caps at `MAX_KEY_TTL_SECONDS`. Both declarations now carry `{'minimum': 1, 'maximum': MAX_KEY_TTL_SECONDS}` sourced from the constant, both `STRUCTURED_PARAMETERS` entries updated, and the plan outcome no longer lumps `key_ttl` in with `cpus`.

**3 (fix) — `STRUCTURED_PARAMETERS` incompleteness.** Your count was right; the derived assertion found exactly one site your list did not (upload truncate `offset`, which you had but under `/uploads/` rather than the real `/upload/`). `test_every_published_structure_or_bound_is_registered` now walks the served spec and fails until every structured or bounded parameter has an entry; twelve backfilled. Kept the registry hand-written on purpose — the derivation says *something is missing*, a human still says what the handler accepts.

**Considers taken: 4, 5, 6, 7, 8, 9.**

- **6** is the interesting one. Tried `validate=Range(min=0)` on the webargs schema first, as suggested — it produces a **500**, not a 422, because the app's error handler renders `UnprocessableEntity` that way. That is the serialisation hazard `base.py`'s `json_or_query` loader already documents, and a 500 for caller error is no better than the truncated 200 it replaced. So the bound is checked in the handler and answers 400. Nothing in the tree uses `validate=` today, so the webargs error path is a latent phase 3 problem rather than a live one, and it is now written down before phase 3 reaches for it.
- **5** generalised over `CONSTRAINT_KEYS` plus `items`, so an entry now describes the published shape in full. **8** builds a fresh mapping and reports a body/query name collision rather than resolving it by ordering; a stale entry naming a removed route is also reported rather than raising `KeyError` mid-table. **9** done. **4** and **7** recorded in the plan outcome: the events maximum must stay a clamp in phase 4, and `CONSTRAINT_KEYS` is three keys deliberately.

**10** needs no action, agreed.

**Verification.** Mutation harness 22 → **24, all caught**, with the two new guards written specifically for this round's mechanisms: a bound on a parameter nobody registered, and a spurious `maximum` on the console `length` entry that exists to assert nothing bounds it. Full unit suite **2612 passed, 0 failed** (2591 before). `pre-commit run --all-files` clean. Adversarially: with the fix reverted, each new test fails for its own reason.

**Convergence.** 3 fix / 0 doc / 6 consider, against last round's 1 fix / 5 consider. The fix count went up rather than down, but the content moved: this round found no new defect in the *original* change — item 1 and item 3 are both consequences of the retyping this PR did, and item 2 is about the guard added last round. That is the generator-loop signal, so this is intended as the last round. If the next review comes back with no `fix` items I would like to land it and let the remaining `consider` items ride to their auto-filed issues.

PR #3726, opened 2026-08-12
Compile the API parameter declarations, and warn

mikalstill:
Phase 3 of `PLAN-api-input-validation`, following #3666, #3682 and #3685.

Every request is now checked against its endpoint's published parameter
declarations, and **nothing is rejected**. `API_VALIDATION_MODE` defaults to
`warn`, which logs what would have been refused and changes no response.
Phase 4 flips it once those logs are understood.

Five commits, split so this can be staged as three PRs if you would rather
review it that way:

| Commit | |
|---|---|
| `d20ab74` | Master plan drift |
| `68b25d5` | The phase 3 plan |
| `c30476a` | **PR 1** — generate the derivation's input space |
| `e239251` | **PR 2** — compile the declarations into schemas |
| `6e1e65c` | **PR 3** — validate, and reject nothing |

## PR 1 — the generator, and the defect it found

The precondition the master plan sets rather than a step in it. Every
assertion in `test_parameter_declarations` compares a declaration against the
derivation, which takes the derivation on trust — and the derivation can only
be exercised by source shapes the tree happens to contain. All four defects
found while building the phase 1 audit were shapes it did not contain, and
mutating the tree cannot reach them either, since it only permutes shapes
already present.

So the axes are crossed instead: route form × `@use_kwargs` binding ×
`flask.request.args` read style, 225 synthetic endpoints, each asserting
`audit()` recovers what the case was *constructed* to mean and reports exactly
the problems its unreadable pieces earn. A second pass declares the one
location each case cannot mean and requires drift — without it the first pass
holds `drifted` empty and so cannot tell a right derivation from one that
returns whatever was declared.

**It found one on its first run.** `derived_location()` returned `path` before
consulting the query sources, so a handler whose declared parameters are all
path parameters could read `flask.request.args` with a key the walk cannot
name and the audit would report the tree clean — the confident wrong answer
`request_args_parameters()` exists to refuse, reintroduced one level up by its
caller. The comment directly beneath that early return asserts the opposite
rule for the other two sources. No handler in the tree has that shape, which
is why nothing had noticed. The tree still audits clean after the fix.

## PR 2 — compilation

132 handlers compile to marshmallow schemas, mounted in a registry built by
walking the mounted routes. Nothing consults it yet.

**Compiled from the rendered specification rather than from the declaration
tuples**, which is a deviation from the plan and a better answer than the one
it proposed. `swagger_helper()` already resolves a token into type, format and
constraints. Reading its output means there is exactly one interpretation of a
token in the process, so the compiled schema and the published specification
cannot disagree — where a second `ARGTYPES` mapping here would have been a
third independent reading, and a type token contradicting its handler is the
defect phase 2 shipped twice.

It also turns three rules the plan wrote as special cases into consequences:
`netblock` is format-only by deliberate choice so it compiles to a plain
string; the prose formats on `uuidorname`/`namespace`/`node`/`url`/`ipv4` do
the same; and the raw upload body renders as a schema whose type is not
`object`, which is the discriminator needed to leave it unparsed. None can be
got wrong by editing the compiler, because none is stated in it.

Two properties are chosen rather than derived, both to stop the compiler
inventing rejections no declaration describes: fields are never `required`
(`mode` on agent-put is declared required while omitting it has always
worked), and always accept `null` (a JSON null reaches handlers as `None`
today).

The strongest assertion cross-checks the runtime compilation against the
static audit — the same declarations read once through flasgger and once by
parsing source must yield the same names.

## PR 3 — warn-only validation

A warn record carries endpoint, parameter, reason, the offending value's
**type** and never its value, and **the status the request returned anyway**.
That last field is why the telemetry is an `after_request` hook rather than
part of the decorator: at validation time the outcome is unknown, and it is
what separates a rejection enforcement would *introduce* from a status code it
would merely *change*.

Three things found or fixed on the way:

* **The decorator had to propagate markers.** Sitting first in
  `method_decorators` puts it between the bound method and
  `_authenticate_unless_public`, which reads `__self__` for the resource class
  and `_sf_public`/`_sf_scope` for the policy markers. Without
  `functools.wraps` plus an explicit `__self__` copy, **every `@public`
  endpoint would have started demanding a token.** `base.py` already warns
  that its older decorators swallow attributes; the plan cited that as a
  reason to avoid `specs_dict` and did not notice the decorator walks into the
  same trap from the other side.
* **The `uuid` → `passed_uuid` remap is gone.** `passed_uuid` occurred once in
  the tree — the assignment itself — so no handler accepted it and sending
  `uuid` in a JSON body was a guaranteed 400 carrying interpreter text on
  *every* endpoint. It was not the collision dodge decision D8 cites it as.
  The collision it was meant to describe is now recorded properly, in
  `log_request`, because by the time the validator runs the overwrite has
  already happened.
* **A webargs error handler is registered for the first time**, so a malformed
  query parameter on the four `@use_kwargs` sites answers 400 in this API's
  `{"error", "status"}` shape rather than 422 in webargs' own. What is
  rejected is unchanged.

## D10 was framed wrongly, and the plan is corrected

I first wrote up `unknown=RAISE` as behaviour-preserving. Probing `POST /auth`
*without* mocking the namespace lookup disproved it: `arg_is_namespace`
short-circuits to 404 before the handler, so the `TypeError` never fires.

Validation at index 0 preempts every per-method decorator, so **neither option
is neutral** — RAISE changes the status wherever such a decorator answers
first, EXCLUDE wherever the request reaches the handler. The recommendation
stands as RAISE but only provisionally, and warn-only counts the two
populations separately so phase 4 decides with numbers. The preemption is
general: a request both malformed *and* referring to a missing object moves
404 → 400 at enforcement, which is a contract change phase 4 must put in the
release notes rather than discover in a bug report.

## Deliberately not done

`required` is recorded and never enforced. Semantic validation of the prose
formats is untouched (phase 6). `except TypeError` stays broad — narrowing it
is phase 5 and is gated on phase 4, since it currently absorbs the malformed
input this layer will start handling.

## Verification

* `tools/check-api-declaration-guards.sh` grows 25 → 31. Three new mutations
  break the *derivation* and three break the *compiler*, which no existing
  mutation did — they ask whether the trust the other 25 place in those is
  earned. Its `run()` gains both new modules, without which none could be
  caught.
* Mutation 31 caught a flaw in itself: written as "stop skipping undocumented
  handlers" it raised `AttributeError` at import and the harness reported
  `HARNESS BROKEN` rather than a catch — exactly the verdict that vocabulary
  exists for. Rewritten to express the defect without preventing the run.
* **All 31 mutations caught.** 2725 unit tests pass. `pre-commit run
  --all-files` clean. The published specification still validates with zero
  errors.
* `shakenfist/external_api/validation.py` is added to the mypy target list and
  passes `--strict`.

## What this does not include

**The measurement.** Warn-only is deployed nowhere, so the phase is
code-complete and evidence-empty. The next step is sfcbr plus a full
functional CI run, and the exit criterion is that every remaining finding is
*explained*, not that there are none. CI matters as much as sfcbr: the ansible
collection is a second implementation of this contract and the likeliest
source of undeclared keys.

https://claude.ai/code/session_012nPqvWFVKNtS63aNm74kRq

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -117,6 +117,6 @@ repos:
         description: Run mypy type checking on selected files (incremental rollout)
         entry: tox -e mypy
         language: system
-        files: ^shakenfist/(schema/|mariadb\.py|daemons/database/|operations/|resource_health\.py|node_health\.py|external_api/(declarations|base)\.py)
+        files: ^shakenfist/(schema/|mariadb\.py|daemons/database/|operations/|resource_health\.py|node_health\.py|external_api/(declarations|base|validation)\.py)
         pass_filenames: false
         types: [python]
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -114,11 +114,14 @@ non-numeric types, contradictory bounds, patterns that do not
 compile, patterns on non-string types, unanchored patterns, and a
 constraint restating a key its type token already renders (a second
 `minimum` on `unsignedinteger`) all raise `InvalidAPIDeclaration`.
-A pattern must be `^...$` anchored because its two consumers read an
-unanchored one differently: JSON Schema `pattern` is an unanchored
-*search*, while the compiled marshmallow validator uses `re.match`,
-which anchors at the start only. Full anchoring is the one form both
-read identically, so it is required rather than documented. Note that
+A pattern must be `^...$` anchored, with no top-level alternation
+(`^a|b$` is anchored on neither branch — wrap it in a group), because
+its two consumers read anything looser differently: JSON Schema
+`pattern` is an unanchored *search*, while the compiled validator
+requires the whole value to match (`re.fullmatch`, chosen because
+Python's `$` also matches before a trailing newline and ECMA-262's
+does not). Fully anchored and group-wrapped is the one form both read
+identically, so it is required rather than documented. Note that
 declared constraints are *published documentation* until phase 3 of
 [PLAN-api-input-validation](../plans/PLAN-api-input-validation.md)
 compiles them — a constraint does not reject anything yet.
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -269,8 +269,10 @@ questions: `unknown-parameter`, `type-mismatch`, `missing-required`
 and `body-path-collision`.
 
 Two things it deliberately does not do. `required` is recorded but
-never enforced — several parameters are declared required while
-omitting them has always worked, and what to do about that is phase 6.
+never enforced — not even in `enforce` mode, where missing-required
+findings are filtered out of the rejection decision. Several
+parameters are declared required while omitting them has always
+worked, and what to do about that is phase 6.
 And the prose `format` on a type token is documentation: `netblock`,
 `uuidorname`, `namespace`, `node`, `url` and `ipv4` compile to plain
 strings, because semantic validation of them is also phase 6. Only
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -111,9 +111,14 @@ exactly that reason, and now reads:
 Constraints are validated at import time in the same style as
 everything else: unknown keys, non-numeric bounds, bounds on
 non-numeric types, contradictory bounds, patterns that do not
-compile, patterns on non-string types, and a constraint restating a
-key its type token already renders (a second `minimum` on
-`unsignedinteger`) all raise `InvalidAPIDeclaration`. Note that
+compile, patterns on non-string types, unanchored patterns, and a
+constraint restating a key its type token already renders (a second
+`minimum` on `unsignedinteger`) all raise `InvalidAPIDeclaration`.
+A pattern must be `^...$` anchored because its two consumers read an
+unanchored one differently: JSON Schema `pattern` is an unanchored
+*search*, while the compiled marshmallow validator uses `re.match`,
+which anchors at the start only. Full anchoring is the one form both
+read identically, so it is required rather than documented. Note that
 declared constraints are *published documentation* until phase 3 of
 [PLAN-api-input-validation](../plans/PLAN-api-input-validation.md)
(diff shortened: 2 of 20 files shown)


mikalstill:

All eight fix items and the document item are addressed in efd9f40a2, with the four consider items triaged: the pattern-anchoring check (9), the merged json_or_query view (10) and the conditional raw-body discard (11) were taken; the functional-CI self-check (13) is recorded in the plan's measurement notes rather than added here, and the repeated-query-key observation (12) is noted there too.

Of note for the re-review: the HTTPException carve-out in suppress_exceptions_to_client is restricted to aborts carrying a crafted response, so a bare abort() still suppresses to a 500 and the every-response-is-{error,status} invariant holds; and each fix was mutation-tested by hand against the new request-level tests before pushing.

mikalstill:

Round 2 addressed in 4346369fc. The fix item (enforce rejected on missing-required, contradicting the documented contract) is fixed by filtering those findings out of the enforcement decision, with a request-level test that enforce plus an omitted required parameter still reaches the handler. Considers 2, 4 and 5 were taken (an 'off' mode, non-printable stripping in finding names, and the validator now reads the body log_request stashed on flask.g instead of re-fetching); the document item adds validation-detail and route to the operator guide's field table, and the warn line now carries the route template for aggregation. Items 6 and 7 were informational and are unchanged. Each behavioural change was mutation-tested before pushing.

mikalstill:

Round 3 addressed in fc86609f1. Both fix items are fixed with request-level tests: findings are now stashed before the enforce decision (an enforced 400 emits its finding line carrying mode=enforce), and a new test pins that a finding on a successful request leaves the 200 response byte-for-byte unchanged. Considers taken: unrecognised type tokens drop their bounds, _schema_findings() catches anything a schema raises, build_registry() refuses colliding class names at mount time, the body is stashed unconditionally with a corrected comment, and the import moved to the right block. The release-notes count is corrected. Item 6 (functional-CI self-check) remains recorded in the plan's measurement notes as before, and item 9 was informational. All four behavioural changes were mutation-tested before pushing.

mikalstill:

(Retry: the previous re-review run failed with 'No result from Claude' after the model run completed — run 31642436427's review step produced no JSON to post.)

Issue #3749, opened 2026-08-13
test_lifecycle_pause_cycle races suite concurrency for node CPU headroom (507 sufficient_idle_cpu)


mikalstill:
## Summary

`guest_ci_tests.test_state_changes.TestStateChanges.test_lifecycle_pause_cycle`
failed in merge CI when its `create_instance` call was rejected by the
scheduler with a 507:

```
shakenfist_client.apiclient.InsufficientResourcesException: ('API request failed',
  'POST', 'http://localhost:13000/instances', 507,
  '{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}')
```

This is a new test joining the known CPU-headroom race family (#3602, #3670,
#3728; #3498 closed but recurring): under merge-queue concurrency the nested
cluster's nodes have no CPU headroom left at the `sufficient_idle_cpu`
scheduling stage, so an unlucky test's instance create is refused outright.

## Occurrence

- 2026-08-13, PR #3736's merge queue:
  [run 31739307168](https://github.com/shakenfist/shakenfist/actions/runs/31739307168)
  (`gh-readonly-queue/develop/pr-3736-327239ac...`), job "Guests (collection) /
  Smoke tests (collection)", step "Run functional tests".
- Failure at `test_state_changes.py:175` → `_start_target('pausecycle')` →
  `create_instance` → 507.

## Why systemic, not PR-caused

- PR #3736 changes `interface_float` skip behaviour in
  `net_iface_op.py`/`net_iface_ip_op.py`; the failure is at instance *create*,
  before any float is attempted, in a test the diff does not touch.
- The identical 507 hit a different test (`test_network_plumbing_lifecycle`,
  #3498's exact signature) in a different job on the *next* merge-queue run
  for the same PR
  ([31740876315](https://github.com/shakenfist/shakenfist/actions/runs/31740876315)).
- Both runs' merge group bases contain `6cd64208` (PR #3724, "Charge placed
  instances at CPU admission"), so this family survives that fix.

Found during merge CI triage of PR #3736.

The `automated-fix-attempted` label is applied at filing time: the prior
automated fix for this family (#3724, closing #3498) landed and did not
eliminate the race, so this needs systemic scheduler/capacity work rather
than another same-day patch.

mikalstill:
Closing in favour of #3772, which consolidates this whole family — #3498, #3602, #3670, #3728, #3749 and #3767 — into one issue.

[...]

Issue #3728, opened 2026-08-12
test_stray_torn_down_while_a_hosted_network_survives races suite concurrency for node CPU headroom (507 sufficient_idle_cpu)


mikalstill:
Found during merge CI triage of PR #3711's sibling, PR #3714.

- Run: https://github.com/shakenfist/shakenfist/actions/runs/31585972641 (2026-08-12)
- Merge group for PR #3714 (JWT rejection logging — touches only `external_api/base.py` and auth tests, nothing in the scheduler or instance-create path)
- Failing job: "Debian 12 tier (collection) / Smoke tests (collection)", step "Run functional tests"; the only test failure in the run

```
{1} shakenfist_ci.cluster_ci_tests.test_stray_vxlan.TestStrayVxlanInstanceProtection.test_stray_torn_down_while_a_hosted_network_survives [78.858225s] ... FAILED
    Traceback (most recent call last):
      File ".../cluster_ci_tests/test_stray_vxlan.py", line 316, in test_stray_torn_down_while_a_hosted_network_survives
        inst = self.test_client.create_instance(
    shakenfist_client.apiclient.InsufficientResourcesException: ('API request failed', 'POST',
        'http://localhost:13000/instances', 507,
        '{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}')
```

This is the same flake family as #3602 (test_disappearing_source_instance), #3670 (test_artifact_show) and #3498 (test_network_plumbing_lifecycle): under merge-queue suite concurrency the nested cluster's nodes run out of schedulable CPU headroom and an instance create in an unlucky worker gets a 507 at the `sufficient_idle_cpu` stage. Supporting evidence from this run:

- `test_single_disk_snapshots` finished at 587s in a concurrent worker just before the failure, so the cluster was under sustained load.
- The same run also skipped `test_local_placement_works` with "Target node does not exist", another concurrency-lag symptom seen alongside these 507s before (see #3602's 2026-08-09 occurrence).
- The scheduler filter-ordering fix from #3722 (for #3565) was already in this merge group's base, so it does not eliminate this family.

[...]

Issue #3602, opened 2026-08-02
test_disappearing_source_instance races suite concurrency for node CPU headroom (507 sufficient_idle_cpu)


mikalstill:
## Symptom

`shakenfist_ci.cluster_ci_tests.test_imagefetch.TestHTTPFetch.test_disappearing_source_instance`
fails merge CI with:

```
shakenfist_client.apiclient.InsufficientResourcesException: ('API request failed', 'POST',
'http://localhost:13000/instances', 507,
'{"error": "No nodes remaining at scheduling stage sufficient_idle_cpu", "status": 507}')
```

Both of the test's creates use `force_placement=n`, where `n` is the first
hypervisor returned by `get_nodes()`. The 507 lands on either of them:

- `test_imagefetch.py:130` — the `inst1` create
- `test_imagefetch.py:152` — the `inst2` create (the more common one)

## Occurrences

Every observed instance is on the **Debian 12 tier** topology, which matches
the capacity observation already made in #3498 (the tier topology has roughly
half the hypervisor capacity of slim-primary).

| Run | Date | PR | Line |
|---|---|---|---|
| [30743921619](https://github.com/shakenfist/shakenfist/actions/runs/30743921619/job/91486140357) | 2026-08-02 | #3596 | 152 |
| [30720961791](https://github.com/shakenfist/shakenfist/actions/runs/30720961791/job/91424615854) | 2026-08-01 | #3592 | 130 |
| [30720572409](https://github.com/shakenfist/shakenfist/actions/runs/30720572409/job/91423571096) | 2026-08-01 | #3582 | 152 |
| [30720239679](https://github.com/shakenfist/shakenfist/actions/runs/30720239679/job/91422708660) | 2026-08-01 | #3582 | 152 |
| [30579273872](https://github.com/shakenfist/shakenfist/actions/runs/30579273872/job/90995222260) | 2026-07-30 | #3562 | 152 |
| [30525600914](https://github.com/shakenfist/shakenfist/actions/runs/30525600914) | 2026-07-30 | #3562 | 152 |

In a sample of seven failing merge_group runs from the last four days, five
listed this test. It is currently the most frequent functional-test failure in
the merge queue, ahead of the `test_network_plumbing_lifecycle` variant that
#3498 was filed for.

## Assessment

This is the same root cause as #3498: a create targeted at a specific
hypervisor competes with the rest of the stestr concurrency-5 suite for that
[...]

PR #3754, opened 2026-08-14
Scheduler reservations phase 3: atomic placement admission

mikalstill:
Phase 3 of `docs/plans/PLAN-scheduler-reservations.md`: instance placement now draws down the phase-2 capacity counters atomically at admission time, instead of merely reconciling them after the fact. The full phase plan (survey, decisions P1–P8, per-step implementation notes and validation evidence) is in `docs/plans/PLAN-scheduler-reservations-phase-03-primitive.md` on this branch.

## What this does

- **New sf-database RPCs `AdmitInstancePlacement` / `ReleaseInstancePlacement`**: one MariaDB transaction performs the guarded counter drawdown (`used_* + claim <= limit_*`, with the decaying `expected_demand` term in the WHERE per D13), the masked placement attribute write, and the INSTANCE_LOCATION reference rewrite. A denial is a clean no-op with per-dimension detail. The namespace-claim branch is built but dormant until phase 4.
- **Pick-then-claim (D7)**: `find_candidates()` output is now a preference list. The create path and the preflight redirect walk candidates, treating a capacity denial as "next candidate" rather than failure; exhaustion is a 507 exactly as before.
- **All placement writers rewired**: the cleaner and queue-startup reconciliation record ground truth non-enforcing (P5) — they probe first and emit a loud over-limit event when reality exceeds the ledger, then write unguarded. A Prometheus counter (`database_admit_instance_placement_unguarded_total`) tracks how often that happens.
- **Release is reference-gated**: INSTANCE_LOCATION rows are the sole authority for what is charged; a caller-named node filters located rows rather than overriding them, making repeat release idempotent (a review-caught defect — a second delete of an errored instance used to decrement the counters twice, silently leaking capacity).
- **Legacy `node_attributes.instances` column removed** (operator-approved, no release-cycle wait), along with its dual-write and read-union. The issue-3498 stopgap and `_committed_vcpus()` are deleted; `summarize_resources()` reads the materialised counters.
- **`SCHEDULER_DISK_OVERCOMMIT` (default 5.0)** scales the free-space headroom term of the derived disk limit, since disk claims are virtual sizes over sparse files.

## Validation

Live concurrency tests (repeatable via `test_mariadb_capacity_admission_live.py` against a real MariaDB) cover: an 8-way race admitting exactly one; a 50-instance burst admitting exactly the 5 that fit, denied on the correct dimension; move semantics without double-charging; crossing moves without deadlock; and a randomised soak after which the phase-2 reconciler computes zero drift. ~85 unit tests cover the transaction shapes, retry classification, and probe/guard structure.

Validation forced one significant design fix: on MariaDB >= 11.6.2, `innodb_snapshot_isolation` defaults ON and a transaction whose read view is established by a plain SELECT aborts later guarded UPDATEs with ER_CHECKREAD (a 50-way burst produced 46 retry-exhausted HTTP 500s). The transactions now open with a guarded UPDATE; probe reads run on a separate autocommit connection. CI's MariaDB 10.11 cannot exercise this, so the invariant is pinned structurally in unit tests and documented in AGENTS.md; a second live-suite job on a newer MariaDB is recorded as future work.

## Soak watch items (plan step 9)

- Reconciler drift stays zero with admission live.
- `database_admit_instance_placement_unguarded_total` stays flat after the first reconcile pass.
- No 507 regression in CI throughput.

https://claude.ai/code/session_01T9m4SUWdprQHd4kA47RyVk


mikalstill:
Review comments addressed in c125a1a56 *(assisted by Claude Code)*:

- **Item 1 (fix)**: the enforce=False probe now waives the demand clause (`enforce_demand and enforce`), so a demand-only refusal no longer emits the over-limit event or costs a second RPC. New test pins it, mutation-checked.
- **Item 2 (document)**: upgrade-ordering (database tier before API nodes) is now stated in the operator guide and the plan's risks table, and the gRPC wrapper names UNIMPLEMENTED in its error string so the mixed-version 500 diagnoses itself.
- **Item 3 (taken)**: the release delete is now filtered to the probed nodes, matching the decrement, so a row and its charge always travel together; docstrings updated on both sides, mock matches, router test pins the filtered DELETE.
- **Item 4 (taken, docstring form)**: the admission form is kept — a release-only repair would leave an instance charged nowhere when the authoritative node has no row yet — and the transient one-instance overcount is now documented.
- **Item 6 (taken)**: `disk_spec_virtual_gb()` and `json_dumps()` promoted to public names; `_json_dumps` remains as an alias for the ~90 internal callers.
- **Item 5 (declined, recorded in future work)**: extracting the shared pick-then-claim walk waits for phase 5's third call site to make the right parameterisation visible; both copies are separately tested today.
- **Item 7 (declined, recorded in future work)**: behavioural ER_CHECKREAD coverage needs a MariaDB 11 runner/DSN, not new test code.

mikalstill:
Second review round addressed in c6148d0db *(assisted by Claude Code)*:

- **Item 1 (fix)**: the denial raise in `place_instance()` is now guarded on `enforce` — a non-enforcing caller whose unguarded retry hit the vanished-capacity-row race events the denial and returns instead of raising into the cleaner's pass. New unit test, mutation-checked.
- **Item 6 (document)**: the ER_CHECKREAD behavioural CI gap is now tracked as #3759 rather than living only in the plan.
- **Item 2 (recorded)**: the P7 fail-open's effect on the claim guard is now an explicit phase-4 prerequisite in the master plan; splitting the flag belongs with the phase that makes claims enforceable.
- **Item 3 (documented)**: the early-out now states place_instance() is not a repair path, and the scheduler guide explains restore_instances() owns missing-reference-row repair.
- **Item 4 (documented)**: the doubled-walk steady state on demand-saturated clusters is noted in the operator guide as the accepted trade and a mis-sized-constants signal.
- **Item 5 (taken)**: the two walk copies now cross-reference each other; extraction stays deferred to phase 5 per the plan, as the review itself endorses.

mikalstill:
Third review round addressed in 5ddb9c93f *(assisted by Claude Code)*. Both fix items were verified against the tree and are **incorrect** — no code change was made for them:

- **Items 1 and 2 (refuted)**: both rest on the premise that the `node_attributes` Table is generated from `NodeAttributesData`. It is not — it is hand-declared in `_get_node_attributes_table()` (`shakenfist/mariadb.py:10184`), which still carries `sa.Column('instances', sa.JSON(), nullable=True)`. An INSERT omitting a nullable column stores NULL (no 1364, which requires NOT NULL without default), and the v3 seeding migration's `src.c.instances` (`mariadb.py:2959`) resolves against that same declaration. Round 1's item 9 verified exactly this and recorded the removal as safe; if issues are auto-filed for these two items they should be closed as invalid.
- **Item 6 (taken)**: the kernel of truth — with no DROP COLUMN migration the column survives in upgraded databases, so ARCHITECTURE.md, CLAUDE.md and the database operator guide now say precisely which parts went (the dual-write and the read-side union) and that the column itself remains in place, nullable and unread, as a rollback fallback until a later release drops it.
- **Item 3 (taken)**: the admission delete-all now carries a comment recording that it sweeps a duplicate's row without crediting that node, why that direction is acceptable, and why the release path differs.
- **Items 4 and 5**: the review itself notes both deferrals are properly recorded (phase 5 walk extraction, phase 4 claim-guard split); nothing further.

No re-review is requested for this round: no valid fix items remain, and the last two rounds' remaining findings are recorded deferrals. Ready to enqueue from my side.

mikalstill:
The merge queue rejected this branch twice. Both runs failed the `tier`
and `cluster` topologies with `507 no node had capacity for this instance,
1 candidates refused it` — five creates across
[31841411662](https://github.com/shakenfist/shakenfist/actions/runs/31841411662)
and
[31847351915](https://github.com/shakenfist/shakenfist/actions/runs/31847351915).
`7c1307a` repairs both causes.

**What the journal shows.** Tracing one failed create: three hypervisors
enter scheduling and all three survive every filter, including
`sufficient_idle_cpu`. `schedule have lowest cpu load` then cuts the list
to a single node, and the guard refuses that node — `cpus limit 3.0, used
3.0` (demand was also over, so `demand_only` is false and the P9 re-walk
correctly does not fire). The walk had nothing left to try, so a cluster
with two idle nodes returned a 507. PR CI never saw this: it runs only the
single-node smoke job, and the multi-node topologies run only in the merge
queue.

**Cause 1, from this PR: the CPU pre-filter went measurement-only.** This
phase replaced the issue-3498 `_committed_vcpus()` placement walk with a
purely measurement-denominated `_has_sufficient_cpu()`, reasoning that the
guard was now the real admission. But `cpu_total_instance_vcpus` counts
running libvirt domains and refreshes once a minute, so a node whose ledger
is full measures as completely idle for as long as its instances spend
fetching images — it stayed in the candidate list and went on to win the
ordering. The filter now charges `max(measured, used_cpus)` again, reading
the capacity counters in one query per scheduling decision rather than
rebuilding a second ledger from every candidate's placement rows. A node
with no capacity row is charged nothing (admission lets it through
unguarded too, P7), and an instance being rescheduled is not charged for
itself on the node it is already on.

**Cause 2, older than this PR: the load bucketing filtered rather than
ordered.** `candidates = by_load[lowest_load]` discarded every node outside
the best band. A band says a node looks busier right now, not that it
cannot host the instance, and since the guard can refuse the head of the
list, discarding the rest converts one refusal into a failed create. The
stage now orders the whole list best-band-first, weighted-shuffled within
each band as before. The comment beside that shuffle already claimed
callers walk the list on failure — which they could not usefully do when it
had been cut to one entry.

Three new unit tests, each mutation-checked by breaking the guard it covers
and confirming the right test fails: a full ledger drops the node, an
instance is not charged for itself, and a busier node is ordered last
rather than dropped. The four load-ordering tests that asserted the bucket
was a filter now assert the ordering. Full `pre-commit run --all-files` is
green, including all 2,967 unit tests. The operator guide and the phase
plan's bug log record the failure and both fixes.

One unrelated failure in the same run, for the record: the `Guests` job's
`Check SF process CPU usage` step failed on `kvm on node sf4 has consumed
1.80 of a CPU, threshold is 1.75`. Every test in that job passed; `kvm` is
the test workload rather than a Shaken Fist daemon, and the check lives in
`shakenfist/actions`, so it needs nothing from this PR.

mikalstill:
Not a review — a note from merge CI triage of PR #3764, because this branch
turns out to own the fix for a flake family that was just consolidated.

## The 507 family is now one issue: #3772

Mikal asked me to stop tracking the 507 `sufficient_idle_cpu` CI failures
per-test. #3498, #3602, #3670, #3728, #3749 and #3767 are closed and replaced
by **#3772**, which carries ~16 recorded occurrences across six tests from
2026-07-18 to 2026-08-15. Future sightings go there.

Per-test tracking had stopped paying for itself: every new victim test earned
a new issue, none could be fixed alone, and the one automated fix that landed
(#3724) closed #3498 and was followed by three recurrences of the exact
signature it was meant to prevent.

## Two things for this branch specifically

**1. The phase table in `docs/plans/PLAN-scheduler-reservations.md` still says
phase 3 is "Not started".** I wrote that into #3772's body on the strength of
the table before finding this PR open, and had to post a correction. Rows
00a/0/1/2 are accurate; it's just row 3.

This matters more than ordinary staleness now. The session working on PR #3766
(the AGENTS.md / ARCHITECTURE.md docs restructure) reports that it strips
phase-number language out of `docs/` prose, and that the sequencing constraints
left in `docs/developer_guide/subsystem_internals.md` explicitly depend on
`PLAN-scheduler-reservations.md` staying accurate. So once #3766 lands, that
row is load-bearing for readers of `subsystem_internals.md`, not just for the
plan. That session is not editing `docs/plans/`, so the fix belongs here.

**2. A check on `7c1307aff` you don't need to repeat.** That commit's *second*
cause — the load-per-thread bucketing filtering the candidate list to the
lowest band rather than ordering it — does **not** explain the #3772 family as
it exists on `develop` today. On develop the bucketing runs at
`scheduler.py:633`, *after* the `sufficient_idle_cpu` stage at line 479, so a
507 raised at that stage genuinely means every candidate failed the CPU check,
not that the list had already been narrowed to one.

The fix is still right, and it becomes load-bearing the moment this phase lands
and the capacity guard is the real admission gate with callers walking an
ordered list — which is precisely the failure this PR's own merge queue hit. I
flagged it only so the two 507s don't get conflated in either direction, in
this PR's history or in #3772's.

The first cause — this phase having replaced the #3724 `_committed_vcpus()`
ledger with a purely measurement-denominated `_has_sufficient_cpu()`, and
`cpu_total_instance_vcpus` measuring a node as idle for the whole minute its
instances spend fetching images — reads correctly to me, and the analysis in
that commit message is the clearest statement of the mechanism I found
anywhere.

## What #3772 asks for beyond this PR

I framed it as three independent levers, with this phase as lever 1 and by far
the most important. The other two are decisions nobody has made, and I would
rather they stayed visible than be assumed away by phase 3 landing:

- **Lever 2: nothing treats a 507 as transient.** CI capacity frees within
  minutes as other tests' instances delete, but the first refusal is fatal.
  Whether retry belongs in the client SDK, the CI base class, or as
  server-side admission queueing is open.
- **Lever 3: topology and node roles.** The Debian 12 tier, where nearly all
  of these land, has three "hypervisors" of which `primary` is also network
  *and* database node and `sf1` is also database node (node dump recorded in
  the now-closed #3767). The CPU arithmetic does not discount that overhead,
  and the suite runs at stestr concurrency 5 with several victim tests using
  `force_placement`, which cannot fall back to a less loaded node.

Suggestion on #3772: keep it open past this PR landing, as the place to
confirm the family actually stops rather than assuming it. Phase 3 makes
admission race-free and gives the walk somewhere to fall through to, but it
does not make a full cluster not-full.

No action needed on the code from my side — I have not touched this branch or
its worktree.

*(Note assisted by Claude Code)*

mikalstill:

mikalstill:
Round 5 addressed in c8b52bf46.

**Item 1 (fix) — taken.** Confirmed: `_capacity_by_node()` is called
from `find_candidates()` and `summarize_resources()`, so the read is on
the create path and the queues preflight. `_grpc_get_scheduler_node_capacity()`
now passes `timeout=BOUNDED_QUERY_TIMEOUT, max_slow_failures=1`, with the
reasoning in the docstring. Pinned by
`test_the_capacity_read_uses_the_bounded_budget`, which closes the
coverage gap noted in the test-coverage section.

**Items 4 and 5 (document) — taken.** Both were factual errors:

* `subsystem_internals.md` now says the *charge* is shared and the limit
  is published live, matching the accurate comment in
  `summarize_resources()`. It also names the published fields
  (`cpu_measured`, `cpu_committed`, `cpu_committed_row_present`) so which
  input binds is answerable from the response.
* `scheduler.md` now matches `database.md`: the 507 reports the count,
  the per-candidate dimensions are on the `schedule failed, every
  candidate refused by capacity guard` audit event.

**Item 6 (consider) — taken, differently.** Rather than hoisting
`placement_filter` and changing the filter signatures, the decision now
runs inside `inst.attribute_memo()`. That is the mechanism built for
this, collapses every attribute read in the pass rather than only the
CPU one, and touches one line. Nothing in the block writes an attribute.
Pinned by `test_prefilter_reads_the_placement_attribute_once`.

**Items 2, 3, 7 — declined, with item 2 newly recorded as future work.**
Item 2's narrowing has to be made identically in both copies of the walk,
and the plan already commits to extracting those into one helper in phase
5 (item 3). Doing it now doubles the divergence risk that extraction
exists to remove, to save RPCs on a path that only runs when a create was
about to fail outright — so it is recorded against the extraction, along
with the test you suggest. Item 7 stays documentation plus deployment
ordering, per your own note that the operator is the release manager.

Both new tests were mutation-tested: reverting each fix makes its test
fail for the right reason.

Issue #3788, opened 2026-08-16
trusted_issuer and mapping_rule are missing from _STATIC_TABLE_GETTERS, so their orphaned state rows are never reaped


mikalstill:
`trusted_issuer` and `mapping_rule` are absent from `_STATIC_TABLE_GETTERS`
in `shakenfist/mariadb.py`, so the orphan reconciler never sweeps them. This
is the same defect as #3588, in two more object types, and it has been
present since federated identity landed.

## What the registry gates

`ORPHAN_RECONCILABLE_OBJECT_TYPES` is derived from `_STATIC_TABLE_GETTERS`
(`mariadb.py:23339`), and `reconcile_orphaned_objects()` iterates it **twice**
(`daemons/cluster/scheduled_tasks.py:637` and `:651`) — once to delete
phantoms (a state row whose static row is gone) and once to repair zombies
(a static row with no state row). So absence from this one dict disables
both directions of reconciliation, not one. A type absent here is silently
skipped: not logged, not counted, not an error.

`constants.OBJECT_NAMES_TO_CLASSES` matters too, but downstream — zombie
repair writes a deleted state row, and the regular deleted-object sweep then
hydrates the object to hard delete it. Both types are correctly registered
there, so registering them here is sufficient; the pairing is worth knowing
about because getting only one of the two is what produced #3588.

## Evidence

Both are ordinary state-bearing `dbo` subclasses:

* `shakenfist/trusted_issuer.py:41` — `object_type = ObjectType.TRUSTED_ISSUER`,
  with `state_targets` through to `STATE_DELETED`
* `shakenfist/mapping_rule.py:244` — `object_type = ObjectType.MAPPING_RULE`,
  likewise

Both have static-values tables with a `uuid` primary key, and both already
have the getter the registry needs:

* `mariadb.py:13372` — `_get_trusted_issuers_table()`, `primary_key_fields=['uuid']`
* `mariadb.py:13989` — `_get_mapping_rules_table()`, `primary_key_fields=['uuid']`

And they are the *only* two missing. Comparing the registries:

```
in OBJECT_NAMES_TO_CLASSES but not in _STATIC_TABLE_GETTERS:
    mapping_rule -> mapping_rule.MappingRule
    trusted_issuer -> trusted_issuer.TrustedIssuer

in _STATIC_TABLE_GETTERS but not in OBJECT_NAMES_TO_CLASSES:
    (nothing)
```

[...]

Issue #3759, opened 2026-08-14
Live capacity suite needs a MariaDB 11 CI job so the ER_CHECKREAD invariant has behavioural coverage


mikalstill:
Scheduler-reservations phase 3 (PR #3754) fixed an ER_CHECKREAD retry-exhaustion bug that only manifests when `innodb_snapshot_isolation` is ON (the default from MariaDB 11.6.2): both placement transactions opened with plain SELECTs, and under snapshot isolation 46 of a 50-way concurrent admission burst exhausted the retry budget and surfaced as HTTP 500s on instance create. The fix is structural (a guarded UPDATE is now each transaction's first statement) and is pinned by `SnapshotIsolationInvariantTestCase`, which asserts statement ordering against a mock router.

What CI cannot currently do is reproduce the *behaviour*: the live suite (`shakenfist/tests/test_mariadb_*_live.py`) runs in the `schema_enum_widening` job on a debian-12 runner with MariaDB 10.11, where the variable does not exist, so `PlacementAdmissionConcurrencyLiveTestCase` passes there whether or not the fix is present. A refactor that introduces a new read into the transaction via a helper the mock router does not classify as a SELECT would slip through the structural test and ship a user-visible 500 on the create hot path.

The fix needs infrastructure rather than test code: a second live-suite job against a `mariadb:11` container (or a debian-13 runner) reusing the existing DSN mechanism, with `innodb_snapshot_isolation` left at its server default. The harness already reports the server regime (version, collation, snapshot-isolation setting) each test ran under, so the job is a runner plus a DSN.

Found during the automated review of PR #3754 (round 1 item 7, escalated to a document item in round 2); recorded in `docs/plans/PLAN-scheduler-reservations-phase-03-primitive.md` future work. Labelled `automated-fix-attempted` because the work is CI-infrastructure scoping, not a same-day code patch.

*(Filed with Claude Code assistance)*

Issue #3522, opened 2026-07-26
Auth path surfaces an unreadable namespace key set as 401 'JWT token uses non-existent key' instead of a 5xx


mikalstill:
When `GetNamespaceAttributes` fails for infrastructure reasons (in this
case gRPC RESOURCE_EXHAUSTED because the namespace's keys blob exceeded
the 4MiB message cap — see #3521), the
API auth path treats the unreadable key set as "the key doesn't exist"
and rejects the request 401 with `JWT token uses non-existent key`. A
database/transport failure is being conflated with an authentication
failure — the same class of error-conflation as #3373.

On the sfcbr cluster this produced 2,959 ERROR events across sf-1, sf-3
and sf-4 from 2026-07-26T18:09:32+10:00 onward (logger
`shakenfist.external_api.base`), and a real outage: nodes reaching
MariaDB via the sf-database gRPC gateway 401'd every request, while
database-tier nodes reading MariaDB directly kept working. Behind the
load balancer, auth therefore succeeded only ~1/3 of the time, and the
CI conductor's main loop died on the first 401 of each cycle, so no CI
workers were started for the evening.

Found by the nightly sfcbr error-mining report for 2026-07-26 (new
signature #2), root-caused during the same evening's conductor outage.
First seen under deploy `c32668f`, which is incidental — the trigger was
the keys blob crossing 4MiB that day.

## Symptom

```
JWT token uses non-existent key
```

Logged at ERROR, no request context. Clients receive 401.

## Why this deserves its own fix

Even with the service-key bloat fixed, the auth path's failure mode is
wrong twice over:

1. A failure to *fetch* the key set should surface as a 5xx (service
   unavailable), not a 401 that tells the client its credentials are
   bad. Clients responded to these 401s by re-authenticating in a loop,
   which is part of why the event count is so high.
2. The intermittency (working DB-tier nodes vs failing gateway nodes)
   made this look like a key-rotation/propagation problem and cost real
   diagnosis time. An accurate error would have pointed straight at the
   database path.

Loki query:

```
{job="shakenfist"} | json | level=~`ERROR|CRITICAL` | logger_name=`shakenfist.external_api.base`
```

PR #3790, opened 2026-08-16
Namespace capacity claims (scheduler reservations phase 4)

mikalstill:
Phase 4 of `docs/plans/PLAN-scheduler-reservations.md`. The plan for this
phase is
`docs/plans/PLAN-scheduler-reservations-phase-04-claims-api.md`, added by
the first commit here.

A **namespace capacity claim** is a namespace's promise of aggregate
capacity. Creating one is itself a guarded admission decision against the
cluster, and an instance create in a claimed namespace draws the claim down
instead of the cluster's unclaimed sums. Claims are admin-only, and served
at `/auth/namespaces/<namespace>/claims`.

## Advisory, deliberately

Exceeding a claim in this release is **recorded, not refused**: the
placement is admitted and an audit event
(`placement admitted over namespace capacity claim`) is written against the
instance. Phase 5 flips it.

The switch is a module constant, `mariadb.CLAIM_ENFORCEMENT_HARD = False`,
and deliberately not a configuration option — an operator-visible `hard`
setting whose 403 path does not yet exist is a trap, and the flip should
arrive with the code that implements refusal.

## The parts worth reviewing carefully

* **The migration-aware mirror guard**
  (`claimed + requested + GREATEST(0, unclaimed_used - migrated) <= total`).
  Creating a claim moves the namespace's existing usage off
  `cluster_capacity.unclaimed_used_*` onto the claim in the same
  transaction, so the guard has to discount what it is migrating or it
  counts the namespace against itself. `migrated` is non-zero for a create
  and **zero for a grow**, since a grow moves nothing.
* **Transaction ordering.** Every new transaction opens with a guarded or
  floored `UPDATE`, never a `SELECT`. Under
  `innodb_snapshot_isolation` (on by default from MariaDB 11.6.2) a leading
  `SELECT` establishes the read view early and makes later guarded updates
  on contended rows abort with `ER_CHECKREAD`. Phase 3 lost a step to this.
  `_direct_update_namespace_claim()` opens with the cluster update only
  conditionally — a pure expiry change or a shrink touches no cluster
  counter — and when that condition is false the first statement is the
  guarded claim update, which is equally DML.
* **Two states, kept apart.** `state` is existence
  (`created`/`deleted`), where every other object publishes it;
  `coverage_state` is whether the claim covers placements
  (`active`/`expired`). An expired claim reads as
  `state: created, coverage_state: expired`.
* **One aggregation, two callers.** Create-time seeding and reconcile-time
  recompute share a single SQL fragment, so they cannot drift and make the
  counters flap every five minutes.

## Testing

3,241 unit tests, 3,138 passed, 103 skipped (the live modules need
`SF_MARIADB_TEST_DSN`), 0 failed. `pre-commit run --all-files` clean. Live
suites were run against MariaDB 11.8 with `innodb_snapshot_isolation` on,
including phase 3's concurrency suite, which still admits exactly one
placement per contested slot with no reconciler drift. Every behavioural
step was mutation tested — the guard was broken on purpose and the test
confirmed to fail with the right message.

Two caveats worth stating rather than burying. The live runs used the
`pymysql` driver rather than CI's `mysqldb`; the SQL is identical and CI
settles it. And CI's own MariaDB predates `innodb_snapshot_isolation`, so
it cannot see an `ER_CHECKREAD` regression in the five new transactions —
that is issue #3759, now with more surface behind it. The sfcbr soak is
what actually exercises that path.

Functional coverage is in
`shakenfist/deploy/shakenfist_ci/cluster_ci_tests/test_namespace_claims.py`,
which reaches the API through `apiclient.Client._request_url()` rather than
client verbs. That is decision D7: the collection installs
`shakenfist-client` from PyPI, so a test written against new `apiclient`
methods cannot pass in CI until a client release exists, and no server pull
request can produce one. The verbs are
shakenfist/client-python#364.

## Known gaps, recorded rather than fixed

* No **REST** surface publishes the `cluster_capacity` singleton, so an
  operator cannot ask the API what is claimed and what is unclaimed. The
  Prometheus gauges (`scheduler_capacity_cluster_total`, `_claimed`,
  `_unclaimed_used`) do publish it, so this is a REST gap rather than an
  observability gap, but `/admin/resources` is the obvious home and it
  would be a small change.
* The claim decrement on a namespace `hard_delete()` cascade is asserted at
  the mock boundary and by inspection, not live, because the cascade is
  gated behind `CLEANER_DELAY` (an hour). The sfcbr soak covers it.
* #3788 (`trusted_issuer` and `mapping_rule` missing from
  `_STATIC_TABLE_GETTERS`) was found while doing this work and is
  deliberately left to its own commit.


mikalstill:
Round 1 handled in `2138413d6`. Five of the seven findings taken, one
declined with reasoning, two needed nothing. Every guard added was
mutation tested — broken on purpose, with the named test confirmed to
fail and then restored.

**Findings 1 and 2 (fix)** turned out to be one defect wearing two hats:
a failure that reads as an absence. `_direct_get_namespace_claim(s)()`
returned `None`/`[]` on `OperationalError`, and both read servicers
turned any exception into a well formed empty reply, so nothing
downstream could tell "no claims" from "the database fell over". The
reads now raise `DatabaseUnavailable` and the servicers set
`StatusCode.INTERNAL` so the client wrapper raises rather than unpacking
an absence. `NamespaceClaim.hard_delete()` now checks `result['success']`
and raises, matching `new()` and `update()`.

Worth recording: this is issue #3522's fix applied to a new object type.
That it had to be found again by review, rather than being impossible to
write, is the more interesting result — the same shape will recur on the
next object with a static table.

**Finding 3 (consider)** taken. `_blocked_by_shrink_floor()` now only
inspects dimensions actually being shrunk, tested against the freshly
read row. Two tests: a stale grow on an advisory over-limit claim now
retries and reports `conflict` (503) instead of `below_usage` (409), and
a stale shrink still reports `below_usage`, so restricting the loop
cannot silently disable the diagnosis.

**Finding 5 (consider)** decided rather than changed, and pinned by
`test_a_deleted_claim_is_still_listed`. A claim has no soft delete, so a
`deleted` object state does not mean "already accounted for" — it means a
row still holding `claimed_*` while it waits for the reaper. The listing
shows it because the listing's job is to account for held capacity;
`arg_is_claim_ref` keeps its 404 because there is no useful action to
offer. The disagreement is now deliberate and documented on
`NamespaceClaims.__iter__`.

**Finding 4 (consider)** declined. The extra point read per claim is real
waste, but N is one claim per namespace, the pattern is shared with every
comparable iterator, and the fix wants a row cached on the object whose
staleness after an `update()` would be a new footgun. It belongs in a
change that does it for all of them rather than one that makes claims the
exception.

**Findings 6 and 7** agreed, no change. The expired-claim release
asymmetry was already recorded under Future work in the phase plan.

Test coverage gaps from the review are closed except the REST-layer 503
propagation, which is handled by an app-wide decorator rather than
anything claim-specific. 3,250 unit tests, 0 failures; `pre-commit
run --all-files` clean.

PR #3801, opened 2026-08-18
Actually delete namespaces in functional CI

mikalstill:
`_remove_namespace()` asked whether a namespace *name* was `in` the result of
`get_namespaces()`, which is a list of `external_view()` **dicts**. A string is
never in that list, so the guard was always false, the delete never ran, and
**no namespaced functional test deleted its namespace between 2020
(`2b7666fa0`) and today**. Every run leaked one namespace per test class, and on
a long-lived cluster they stayed until the cleaner collected them.

Found while addressing the review on #3794, which predicted the phase 7 Loki
detector's `addCleanup()` on a minted key would 404 against a namespace
`tearDown()` had already deleted. The mechanism was right and the prediction was
wrong, and the reason it was wrong is this bug: the namespace was still there.
That fix landed anyway, because it becomes necessary the moment this PR does.

## The same mistake, four times

The fix is a shared `namespace_names()` helper rather than four repairs, because
the evidence is that this is a mistake people make repeatedly and that reviewing
for it does not catch:

| Site | Consequence |
|------|-------------|
| `base.py` `_remove_namespace()` | Six years of leaked namespaces |
| `smoke_ci_tests/test_auth.py` | Vacuous `assertNotIn` |
| `cluster_ci_tests/test_auth.py` | Vacuous `assertNotIn` |
| `cluster_ci_tests/test_upgrades.py` | A `skipTest()` guard that could not detect the condition it names |

The last one is the worst of the three harmless ones: `test_upgraded_data_exists`
would have skipped even on a cluster that *did* have an `upgrade` namespace. It
still skips, because nothing builds one, but the guard can now tell the
difference.

## Two behaviours that change once deletion is real

* **`_make_namespace()` no longer pre-deletes an existing namespace.** That call
  could never fire, and *making* it fire would be wrong: namespace deletion is
  soft, and the create endpoint refuses any name whose row still exists whatever
  its state, so delete-then-create would answer `403 namespace exists` and read
  as though the delete had failed. Every caller uniquifies, so a collision is
  something the test author should be shown. Removing it is provably a no-op.
* **`_remove_namespace()` retries a 400.** The delete endpoint refuses a
  namespace still owning a live instance or network, and `tearDown()` removes
  those through a *non-blocking* client, so a namespace whose objects have
  stopped being listed can be seconds short of deletable.
  `RequestMalformedException` specifically rather than `APIException`, and with a
  deadline, so a refusal that will not clear still fails.

## Two guards, because the failure mode was to succeed quietly

* `smoke_ci_tests/test_ci_harness.py` — functional. Drives the helper against a
  real API, and runs an empty namespaced test in-process to inspect what
  `tearDown()` leaves behind, since a test cannot observe its own teardown.
* `shakenfist/tests/test_ci_namespace_membership.py` — static AST check, no
  cluster, inside the required check. It rejects the *shape* of the comparison
  rather than a list of known sites, so a fifth occurrence fails before it
  merges.

Both were mutation tested by restoring the bug; the static one reports
`base.py:107`.

## Functional evidence

Dispatched on this branch rather than waiting for the merge queue, because the
blast radius is the whole suite: [run
32123955277](https://github.com/shakenfist/shakenfist/actions/runs/32123955277),
**7 of 8 jobs green** — smoke, Debian 12 cluster, Ubuntu 24.04 cluster, Guests,
Ansible modules, node lifecycle, schema ENUM widening. This is the first time
any of the 42 namespaced test classes has deleted a namespace at teardown, which
exercises `Namespace.hard_delete()` and the delete endpoint's artifact reaping on
a path CI has never taken. The new guards ran and passed:

```
test_ci_harness.TestNamespaceHelpers.test_a_namespaced_test_deletes_its_namespace [6.211464s] ... ok
```

**The one failure is pre-existing.** Debian 12 tier lost two tests, both to an
identical `507 No nodes remaining at scheduling stage sufficient_idle_cpu` from
`create_instance`, 21 seconds apart on different workers. The same job fails the
same way on `develop`'s merge queue in three of the last four runs, six
occurrences per run, with a different random victim each time
(`test_object_names`, `test_floating_ips`, `test_stray_vxlan`,
`test_network_lifecycle`); this run drew `test_commandline_artifacts` and
`test_imagefetch`. Same mechanism, arbitrary casualties: the half-capacity tier
topology losing a scheduler race under under-cloud contention, which the
workflow comment already anticipates.

I did consider whether this PR could cause it — namespace deletion now reaps
artifacts, which is real I/O on a capacity-constrained cluster. The `develop`
history rules it out: the failure predates the change and picks victims at
random rather than clustering on artifact-heavy tests.

The second commit only touches the new test file, so the evidence above still
stands for the fix itself. Happy to re-dispatch against the final head before
merging if you would rather.

## Deliberately not in scope

`test_namespace_create_delete_and_list` creates namespaces `a`, `b`, `d` with
fixed names and never deletes them, so it can only pass once per cluster. Fresh
CI clusters hide it; a long-lived cluster would not. Unrelated to this fix and
untouched.


mikalstill:
Review addressed in `ede0d9a`. Both `fix` items, three of the four
`consider` items, and one `none`.

## Fixed

**1 — the AST guard could pass on having read nothing.** It now counts
the files it scanned and asserts against a floor of 20, a long way below
the 54 in the tree, because the assertion is about the walk having
happened rather than about the file count. Mutation tested by pointing
`_suite_root()` at a directory which does not exist: it fails with *"The
CI suite scan found 0 Python files under …_renamed, so an empty offence
list below would mean nothing. Check `_suite_root()`."* Restoring the
original bug in `base.py` still reports `base.py:115`, so the new
assertion did not displace the one that matters.

**2 — the harness test could leak the namespace it guards against.**
`self.addCleanup(self._remove_namespace, namespace)` immediately after
`namespace = case.namespace`, with the reason recorded at the call site.
Your reasoning about the `setUp()` window between `create_namespace()`
and `add_namespace_key()` is exactly right, and the failure path is the
one that mattered.

## Taken from `consider`

**3 — the retry now matches before it retries.** Only the two refusals
it was written for, keyed on the `you cannot delete a namespace with`
prefix that both carry; every other 400 re-raises on the first attempt.
It also logs the refusal text on the first retry rather than only at the
timeout, so a teardown which is going to be slow says why while it is
happening.

I also closed the coverage gap you noted for that loop, which cannot be
reached from a real cluster on demand — producing the refusal it retries
means racing a delete against a live instance, and producing the
refusals it must not retry means a request the client will not send. So
four tests in `test_ci_harness.py` substitute the client instead of the
cluster: a refusal which clears is waited out, an unrelated 400 raises
after exactly one delete, the deadline re-raises, and a 404 between the
listing and the delete is success. All four mutation tested — retrying
every 400, retrying none, and dropping the 404 tolerance each fail the
matching test and only that one.

**7 — `test_upgrades.py` is deleted.** You are right that a
permanently-skipping test reads as coverage while providing none, and
fixing the guard only changed *why* it skipped. What it would take to
reinstate it, and the one upgrade hazard the suite does cover (the
`Schema ENUM widening` job), is now recorded under a new *Coverage the
functional suite does not have* section in
`docs/developer_guide/ci.md`.

**8 — the dual import is gone**, in all five files rather than the one
you flagged. `base.namespace_names()` at every call site, which is both
the project convention and visibly the shared helper — worth having,
given the whole point is that four files each had their own.

**9 — the copyright header is on the smoke test.**

## Not taken

**4 — nothing changes in code**, which I think is what you were saying
too. Failing loudly is the point, and the answer to the question you
actually raised is a run rather than a diff: dispatched against this
head as
[32217801316](https://github.com/shakenfist/shakenfist/actions/runs/32217801316).
The `delete_wait` route you describe is real, and if it shows up it will
show up there.

**5 and 6 — noted, no action.** The artifact reaping is intended and
worth knowing when the first post-merge cluster run's blob metrics are
read; the module-wide binding in the AST check is the trade the
docstring already states, and a false positive costs a rename.

PR #3829, opened 2026-08-21
Bring the plan docs onto the shared consistency blocks

### LINKED ISSUE #3733: Consistency: Pre-push audit file
This project is not yet compliant with the **Pre-push audit file** consistency audit.

**Audit spec:** [development/audits/push-audit.md](https://github.com/shakenfist/development/blob/main/audits/push-audit.md)
**Template:** [development/templates/shared-blocks/README.md](https://github.com/shakenfist/development/blob/main/templates/shared-blocks/README.md)

### Automated check details

missing shared block plan-phase-references (copy it verbatim from templates/shared-blocks/plan-phase-references.md in the development repository)

---
*This issue was created automatically by the consistency audit workflow.*

### LINKED ISSUE #3810: Consistency: Plan index
This project is not yet compliant with the **Plan index** consistency audit.

**Audit spec:** [development/audits/plan-index.md](https://github.com/shakenfist/development/blob/main/audits/plan-index.md)
**Template:** [development/templates/shared-blocks/README.md](https://github.com/shakenfist/development/blob/main/templates/shared-blocks/README.md)

### Automated check details

1 table(s) not led by Date then Plan columns: line 36 starts "Plan | Phase"; 5 master plan(s) not listed in the index: PLAN-artifact-fetch-retry-with-backoff.md, PLAN-attribute-field-masks.md, PLAN-ci-node-exec-assertions.md, PLAN-netserv.md, PLAN-qemu-futures.md

---
*This issue was created automatically by the consistency audit workflow.*

### LINKED ISSUE #3811: Consistency: Plan template
This project is not yet compliant with the **Plan template** consistency audit.

**Audit spec:** [development/audits/plan-template.md](https://github.com/shakenfist/development/blob/main/audits/plan-template.md)
**Template:** [development/templates/shared-blocks/README.md](https://github.com/shakenfist/development/blob/main/templates/shared-blocks/README.md)

### Automated check details

missing shared block plan-status-vocabulary (copy it verbatim from templates/shared-blocks/plan-status-vocabulary.md in the development repository)

---
*This issue was created automatically by the consistency audit workflow.*

mikalstill:
Works through three of the open consistency-audit issues. They are separate
commits because they are separate fixes, but they share a subject: the
scaffolding around `docs/plans/`.

Every claim below was verified by running the audit's own checker against a
checkout of this branch, rather than by trusting the filed issue text:

```
python3 scripts/audit-check.py --repo-path <worktree> --repo-name shakenfist
```

from `shakenfist/development`. All three checks now pass, and the eight
pre-existing audit failures on this repository are unchanged.

## Fixes #3811 — plan template

`PLAN-TEMPLATE.md` carried seven of the eight required shared blocks.
`plan-status-vocabulary` is added under `## Execution`, where the phase table
it governs is written.

## Fixes #3733 — pre-push audit file

The issue names one missing block; the audit against the current tree reports
two. `llm-doc-discipline` replaces the two ad-hoc bullets telling the
documentation reviewer to make `ARCHITECTURE.md` and `AGENTS.md` "reflect"
every new module and convention — the same feedback loop `readme-discipline`
exists to stop, one file over. Their useful half, the local list of what counts
as a structural change here, survives as a trailing bullet.
`plan-phase-references` goes into the same section.

## Fixes #3810 — plan index

Larger than its issue text implies. The checker skips date and status
validation entirely on a table whose first two columns are not `Date` and
`Plan`, so simply prepending a Date column would have traded one finding for
three louder ones: 157 rows of per-phase status would have become visible, and
the index carried its own `Status Definitions` section (`Stub`, `Planning`,
`Future`, `Dissolved`, `Blocked on preconditions`) contradicting the shared
vocabulary.

So the index becomes one row per master plan — `Date | Plan | Intent | Status |
Phases` — matching the audit's model and the layout `ryll` already uses. Every
master plan already tracks its own phases in its own Execution table, and
comparing the two showed the index's per-phase descriptions were duplicates
that had drifted.

Five plans were unregistered and are now listed, with statuses read from the
tree rather than the plan text: `artifact-fetch-retry` (`defer_with_backoff` is
present), `attribute-field-masks` (`instance_location` reference rows are
present) and `ci-node-exec-assertions` (the primitive and its consumers are
present) are Complete; `netserv` and `qemu-futures` describe themselves as
thoughtbubbles, so Proposed. Four of the five were also missing from
`order.yml`, which made them invisible in the docs site navigation.

## Fourth commit — the phase tables

The shared block governs a master plan's own Execution table as well as its
index row, and nothing mechanical checks the former, so those had drifted the
same way. Seven master plans carried 24 non-vocabulary cells. Most were stale
evidence qualifiers (`Complete (pending CI)`); Kerbside's nine phases said
`Done`. Where a cell held facts recorded nowhere else they moved rather than
went — a Landing record in the Loki plan, a Phase status notes section in
scheduler-reservations, whose cells had grown into paragraphs.

Cross-checking the index against the phase tables afterwards caught seven rows
whose arithmetic was wrong, mostly from my own reading of those tables in the
previous commit.

## Known follow-ups, not in scope

* The session plan-status hook parses the new layout, but with one row per plan
  it reports a plain status instead of `4/10 phases complete, next: ...`. That
  is a change to a hook outside this repository.
* 23 plan files still refer to the "Plan Status table" in their own
  index-maintenance sections. They are historical records of instructions
  already carried out, and new plans are written from `PLAN-TEMPLATE.md`.
* The step tables inside phase plans keep their prose statuses. They are a
  working log of one phase's implementation rather than a plan status, the
  shared block does not reach them, and flattening them would destroy the
  record.

https://claude.ai/code/session_01UyJQKHRD4NAR72TKfULJUf


mikalstill:
Thanks — I checked each item against the tree rather than taking it at face
value. Six `fix` and five `consider`; two commits pushed
(`a5476fb`, `148002f`).

**Taken, and two of them were real defects in this branch:**

* **#4** — the important one.
  `PLAN-fix-cluster-operation-targets-unique-constraint.md` opened with a prose
  `Proposed` while the index row said `Complete`. Confirmed: the composite
  constraint is in the tree. I had read that plan's status from the code and
  never read its own header, so the branch shipped a file contradicting itself.
  Fixed, and the other two prose-status plans normalised with it.
* **#8** — also mine. The template note forbade exactly the `Complete | 5 of 7`
  row this branch ships for remove-primary. Reworded to match index.md.
* **#2** — fair. "Every master plan..." overclaimed, and I knew, because those
  four plans were the exemptions in my own cross-check script. The preamble now
  says most, and names them.
* **#1** — over-counted, but right in substance. Eleven of the 23 are
  instructions telling an author to add one row per phase, which is the thing
  this PR exists to stop, in files the template's execution model sends
  sub-agents to read. All 23 updated: instructions reworded, past-tense records
  given the corrected table name and their tense left alone.
* **#3**, **#11** — both correct, both done.
* **#6** — taken, as `tools/check-plan-status.py` plus a pre-commit hook. It
  would have caught #2 and #4 on its own, and it is the query I had already
  hand-written twice while doing this work. Mutation tested against five
  deliberate breakages.

**Not taken:**

* **#5** — this reads the block as documenting tooling the repository does not
  have. The block says *the consistency audit*, which is the fleet-level one in
  `shakenfist/development`: implemented at `scripts/audit-check.py:2742`, and
  already reporting these violations as issue #3732. So it is not a rule the
  tree fails on day one — it is a rule that already applied, with its own open
  issue. Fixing those lines here would be doing #3732's work inside a PR about
  #3810/#3811/#3733. What the finding is right about is that a reader goes
  looking in `tools/`, so the block now says where the audit runs.
* **#7** — accurate about `issue-fix.yml:183`, and it grades itself correctly as
  degraded triage precision rather than breakage, having verified no content was
  lost. It is a workflow prompt change rather than a docs one; leaving it to the
  auto-filed issue.

**Held for the author** — both need knowledge I do not have:

* **#9** — `qemu-futures` at `—` versus `0 of 3` is a judgement call about
  whether that table is settled.
* **#10** — verified that no phase-00 decisions plan exists and phases 1-9 all
  shipped. Whether that is `Abandoned` or a token-format spec still owed is
  Michael's call.

mikalstill:
Your #1 was right, and it was the important one. I reproduced it before
changing anything: phase files with no inbound markdown link, develop versus
this branch, **24 → 38**. The fourteen newly unreachable are exactly the
fourteen you named.

**The fix goes wider than the fourteen.** The cause is that many master plans
list phase files as bare filenames in the Execution table, and the old index
rows were the thing carrying them. Converting those to links fixes **36 of the
38** — the same defect was already on develop for health-checks,
network-facade, node-resource-health, replace-last-cluster-operation and
sql-pushdown-filtering, which this branch merely exposed. The remaining two had
no mention anywhere to convert: the scheduler-reservations findings document
was named in a bare parenthetical that reads like a link whose text half went
missing, and remove-primary phase 6 step 6 links out to its parents with
nothing pointing in. Both now have a way in. **Zero unreachable, against 24 on
develop.**

Your suggested fifth check is in: every markdown link between plans must
resolve, and every phase plan must have an inbound link. Writing it immediately
caught a real bug in the checker (the absolute `PLAN-TEMPLATE.md` URL in the
index needed excluding), which is a fair sign the rule bites.

**#2** — correct, and the commit subject was writing a cheque the change did
not cash. `shakenfist/tests/test_plan_status.py` is modelled on
`test_doc_anchors.py` exactly as you suggested, including the closing
regression guard over the real `docs/plans/`. 17 tests. I left the earlier
subject alone rather than rewrite pushed history; the claim is now true.

**#3, #4, #5** — all three taken, and all three were better findings than
"consider" suggests:

* the em-dash hatch is now a named `PROVISIONAL` set, held to the same standard
  as `HAND_COUNTED`;
* `cells()` splits on unescaped pipes only — your point that ten phase plans
  already quote LogQL this way is what made it worth doing now rather than
  later;
* the vocabulary is parsed from the shared block instead of duplicated, and a
  missing block is now a loud failure rather than an empty allowlist. That one
  was the same drift this PR exists to close, one level up.

Mutation tested, ten cases, and the tests were mutation tested too — disabling
the orphan check and reverting the escaped-pipe fix each turn exactly one test
red, so they are not passing for the wrong reason.

Still open for @mikalstill, unchanged from last round: whether `qemu-futures`
should read `0 of 3` rather than `—`, and whether kerbside's phase 0 is
`Abandoned` or a token-format spec still owed.

mikalstill:
All six reproduced against the tree before changing anything. Five taken, one
partly.

**#1 was the sharp one, and it caught me making the same mistake twice.**
`functional-tests.yml` gates its unit test job on a filter using
`predicate-quantifier: every` with `!docs/**`, so a plan-only edit matches
nothing, `code_changed` is false, and the job — and the regression guard inside
it — never runs. That is exactly the change the guard exists to police. The
filter now has a `plans` output for `docs/plans/**` and `PLAN-TEMPLATE.md`, and
the job runs when either is true. I chose that over adding to `docs-tests.yml`
because `functional-tests.yml` already gates the merge queue, so the coverage
lands where the merge actually happens.

Worth saying plainly: this is the second time on this branch I asserted CI
coverage without tracing trigger → filter → job. The docstring now states where
the coverage comes from *and* what it depends on.

**#2 was right, and my own mutation testing missed it.** I had mutated the
orphan check and the pipe fix, but never the vocabulary check — and the test
mutated the *completed* phase, moving the arithmetic too, so the arithmetic
rule fired and the test would have passed with the vocabulary check deleted.
It now mutates the unfinished phase. I took the "better still" as well:
`main()` is split into a `problems()` the tests read, and every fixture test
asserts on which problem fired rather than on the exit code. Re-mutated after —
disabling the vocabulary, arithmetic and each new status rule turns exactly one
test red apiece.

**#3** — confirmed, `Complete | 0 of 8` exited 0. Both one-sided rules
implemented as you described, with a fixture test each direction.

**#4** — confirmed and worse than stated: `mkdocs.yml.tmpl` hard-codes three
plan entries, and nothing outside my own new checker reads `order.yml` at all.
I took the second option: `order.yml` is described as the reading order and
registration it actually is, in the template, five plan files and the checker
docstring. The inbound-link rule now rests on the repository's own
navigability, which is where these are read. Making the nav claim *true* is a
docs-sync change and a bigger piece of work than this PR.

**#5** — taken for everything unambiguously stale: retiring etcd described as
pending when `shakenfist/etcd.py` and `etcd3gw` are gone (the vestigial
`is_etcd_master` flag deliberately remains), remove-primary phase 7 and the
whole BYO-MariaDB plan as future when both are Complete, and the triage
paragraph deferring to "when remove-primary is close to landing". The six
table-only plans are now accounted for. **I left the triage itself alone** —
which of the unordered plans goes first is @mikalstill's call, not a defect
this PR should resolve.

**#6** — taken. All four phases have merged to develop, which the merge queue
gates on CI, and both plans now say the caveat is spent rather than silently
dropping it.

One thing that fell out: my test fixtures were named `PLAN-thing.md`, which the
fleet audit's `plan-source-references` check reads as dangling references to
real plans. Renamed to `FIXTURE-*`.

---

**On convergence.** Round 1 was 6 fix / 5 consider, round 2 was 2 fix / 3
consider, this round 3 fix / 2 doc / 1 consider — and #1, #2 and #3 are all
about machinery added in rounds 2 and 3, which is the generator-loop signal
from the skill. The difference is that all three were real: a guard that did
not run, a test that could not fail, and a documented rule nothing enforced.
Round 4 would be reviewing the paths filter and the `problems()` split, and I
would expect it to be mostly `consider`.

My recommendation is to land after this round unless something `fix`-shaped
comes back. Still open for the author either way: `qemu-futures` at `—` versus
`0 of 3`, kerbside phase 0 as `Abandoned` or a spec still owed, and now the
sequencing re-triage.

mikalstill:
Round four addressed in three commits. Every item was reproduced against the tree before being acted on.

**1 (fix) — example tables parsed as phase tables.** Real, and worse than "latent". Parsing the two named files before the fix gives three phantom rows in `PLAN-qemu-futures.md` and two in `PLAN-grpc-bounded-replies.md`, exactly as described. The template instructs authors to include that block, so the next small plan created from it would have had its arithmetic computed from the template's placeholder rows. Fenced blocks are now blanked with offsets preserved (line numbers in messages stay honest) and a table row must begin in column zero. The same blanking applies to link extraction — I checked first that no `.md` link in `docs/plans/` currently sits inside a fence, so nothing became orphaned by that.

**2 (fix) — the merge-queue claim.** Correct, and I took the second option. `sanity_checks` is in `can_enqueue`'s needs and not `can_merge`'s; that split is the workflow's deliberate design — cheap checks gate the pull request, expensive integration jobs gate the merge — and moving a lint-and-unit-test job onto the merge gate is a change to make on its own terms, not as a side effect of adding a checker. The docstring now claims only PR-entry coverage. Third time on this branch I have overclaimed CI coverage; the correction is in the commit.

**3 (consider) — order.yml.** Taken, with one correction. Entries are now parsed line-anchored and checked both directions. But commenting an entry out is not simply "the way to break the check": `tools/sync_component_docs.py` in shakenfist/actions documents it as the sanctioned way to keep a page out of a synced navigation. So a commented-out entry is reported as its own distinct problem rather than being silently accepted or conflated with a forgotten one.

**4 (document) — what order.yml is for.** Taken, and the guess in the suggestion is wrong: nothing in shakenfist/development reads it either. The real consumer is `sync_component_docs.py`, which owns the format and its meaning as a per-directory nav allowlist. Both the template and the checker docstring now say that.

**5 (document) — updating_docs.md.** Taken, including the division between the two checkers.

**6 (consider) — the PROVISIONAL plans.** Taken. Both exemption sets are now named for readers.

**7 (consider) — untested branches.** Taken. Ten new tests: the parsing rules, HAND_COUNTED, the PROVISIONAL em-dash (which pins the character), ellipsis rows, stale index rows, and both order.yml directions. The exemption sets are substituted rather than named, so the tests exercise the mechanism while the whole-tree regression test answers whether the membership is still right.

**8 (consider) — the short-row skip.** Taken; it is the silent-skip class, and hiding a row is the drift this checker exists to catch. Indentation fixed.

**9, 10 (info)** — no action. On 9: reusing `sanity_checks` stays the right call, and splitting a dedicated job is the same "change the gate structure" decision as item 2, better made separately.

Mutation tested rather than read: ten rules disabled in turn, each turning exactly the test about it red. That harness earned its keep — my first two attempts to mutate the column-zero rule did not actually change the parse and read as a missing test rather than a bad mutation.

**Convergence.** 6 fix → 2 fix → 3 fix → 2 fix / 2 doc / 4 consider. Both fix items this round were about machinery this PR added, which is the generator-loop signal, but item 1 is a latent defect that would have bitten the next plan created from the template rather than a defect in a defect's repair. **My recommendation is to land after this round.** The remaining surface is the fence handling and the new fixtures; a fifth round would be reviewing those, and unless it returns something `fix`-shaped the outstanding items should ride to the auto-filed issues.

Separately, and outside the review: the QEMU stack plan is now `Not started` rather than `Proposed` at Michael's direction — nothing has happened there but discussion, and it is a plan we intend to get to. The Kerbside phase 0 question is answered the same way, and needed no edit: it already reads `Not started`, not `Abandoned`.

mikalstill:
Round five addressed in three commits. **This is the last round — I am not requesting another review.** Reasoning at the bottom.

**1 (fix) — the `plans` paths filter can never match.** Correct, and the most serious finding on this PR. Verified against the action's own documentation rather than reasoned about: "With `predicate-quantifier: 'every'`: 'true' - if any changed file matches **all** of the filter's rules". Two mutually exclusive patterns means the output is always `'false'`, and a non-empty string being truthy means the `|| 'true'` fallback never fires either. The gate I added in round three has been dead code since I wrote it.

Fixed with the single-pattern form, which behaves identically under every quantifier and so cannot fail this way again. This is the third consecutive round in which a claim I made about CI coverage was false — the claim, then the docstring, now the mechanism — and each correction was narrower than the mistake. The commit records that.

**2 (fix) — duplicate index rows.** Real. Rows are collected per plan and a repeat is reported with both line numbers.

**5 (consider) — `index_rows()` not masking fences.** Taken; it was the one parser not reading through `outside_code()` while two documents claimed otherwise.

**3 and 4 (fix) — stale claims in plan files.** Both real, and both under-enumerated here: item 3 names three files, `grep -rn 'consumes .order.yml.' docs/plans/` finds five; item 4 names six, which is right. Fixed all twelve. Worth noting for the record that the `mkdocs.yml.tmpl` half of the item-3 paragraph is *true* — `ci_sync_docs.sh` in shakenfist/actions does regenerate `mkdocs.yml` from the template — what was false is that it consumes this repository's plan `order.yml`; the ones it reads belong to the external repositories synced into `docs/components/`.

**6 (consider) — reachability vs. link existence.** The observation is right and the message overclaimed. I took the softening, not the transitive walk. The walk is the better invariant and I have noted it in the docstring as the obvious next move, but adding machinery in the round I intend to land is precisely how this review got to five rounds.

**7 (consider) — the unenforced exemption-list claim.** Half taken, the other way round from the suggestion. The preamble was asserting that its list and the checker's agree with nothing enforcing it, which is the exact failure this PR exists to remove — so the assertion is deleted rather than a fifth parser added to defend it. Adding `**Status**:` lines to two hand-counted roadmaps is a change to those plans, not this one; declined.

**8 (consider) — 45 minutes for a plan-only edit.** Accepted as a chosen cost and written into the workflow comment. A separate cheap job needs its own entry in branch protection, and one gate in one place is worth the runner minutes.

**9, 10, 11 (info)** — no action. On 10: `PLAN-remove-primary.md` phase 2 stays `Abandoned`; the vocabulary requires `Superseded` to name a superseding *plan* and here it is a phase, so the reviewer's own reading is the one I agree with.

---

**On convergence, which is the actual question.**

Counts by round: 6 fix → 2 fix → 3 fix → 2 fix → 4 fix. That is not converging, and the raw trend argues for stopping. But the composition matters more than the count, and it says something more specific: **the reviewer under-enumerates.** Item 3 named three of five instances. Round four's fence finding named two of two, but round three's CI finding fixed the symptom and left the mechanism, which is what round five caught. Fixing exactly what is named therefore guarantees another round, every round — not because new defects are being created, but because each round samples a class rather than closing it.

So the remaining findings this round were not really four defects. They were two defects (a dead paths filter, a lossy dict) plus two *classes* of stale prose, both of which a grep closes completely. I closed them by grep, not by the list. That is the difference between a loop that terminates and one that does not.

What that leaves for a sixth round is the fence handling, the duplicate-row check, and three new fixtures — all added in the last two rounds. Reviewing those is reviewing my repairs, and four of the last five rounds' `fix` items were already about machinery this PR introduced. That is the generator signal, and I do not think another sample of it buys anything the tests do not already assert: twelve rules, each mutated in turn, each turning exactly the test about it red.

Landing. Anything the reviewer would still say rides to the auto-filed issues.

PR #3843, opened 2026-08-22
Scheduler reservations phase 4a: a satisfiable demand guard

### LINKED ISSUE #3813: Scheduler demand guard is arithmetically unsatisfiable on nodes with fewer than four schedulable threads

## The problem

(`shakenfist/mariadb.py`) is:

```
```

expected_demand*.

## Observed

```
"dimensions": [
],
"enforce_demand": true,
"failing_stage": "node",
```

## Why it matters

  absolute sum.
  a load figure.

same-day patch.

mikalstill:
Phase 4a of [PLAN-scheduler-reservations](docs/plans/PLAN-scheduler-reservations.md), inserted between phases 4 and 5. It fixes #3813, a live defect in the admission code phase 3 shipped, and clears the way for phase 5 — which makes that code the sole gate on placement, so the fix has to land first.

**This PR does not close #3813.** Closing it waits on the sfcbr soak, which is step 5 and is operator work.

## The defect

The D13 demand clause denominated its budget per schedulable thread and its charge per requested vCPU, and the two were never reconciled:

```
cpu_load_1 + expected_demand + demand_add <= SCHEDULER_TARGET_LOAD x cpu_schedulable
```

At the seed constants a 1-vCPU instance was charged `1 x 2.5` against a budget of `0.75 x cpu_schedulable`, so a node needed 3.34 schedulable threads before it could admit that instance *at zero measured load and zero expected demand*. A 4-vCPU instance needed fourteen. The CI hypervisors publish `cpu_schedulable: 2`.

The P9 waiver meant this did not fail creates — both walkers re-walk with the clause waived. What it cost was two full candidate sweeps per create, and a spreader that never operated, on the small clusters where spreading matters most.

## What the survey found

Three findings, all corrected at their source in this branch.

**The seed constant was transcribed from the wrong row of its own measurements.** D13 attributes 2.5 to "the 00a-1 measurements". That appendix's demand-per-vCPU row reads 0.12–0.35 steady with a ~0.6 burst peak, and names "~0.33 steady / 0.6 conservative" as the seed for exactly this constant. The only 2.5-shaped figure there is a different row: 2.3–3.0 *allocated vCPUs per thread*, the packing figure behind `CPU_OVERCOMMIT_RATIO`. That reclassified the fix from a tuning question deferred to the phase 0 step 3 analysis into a correction of it.

**The master plan's phase 6 correction misdescribed the mechanism.** It said the create "places through a single forced candidate and the affinity stage has nothing left to rank". The re-walk iterates the same list in the same order, so ranking including affinity survives intact. What is lost is spreading: nothing makes the top-ranked candidate less attractive to the next create in a burst, against a metrics row up to 60s stale. Phase 6 now inherits a corrected premise, and D6's three positions stay live rather than moot.

**Nothing outside the admission transaction reads `expected_demand`** — no occurrence in `scheduler.py`. That is what makes this a contained change rather than a scheduler rework.

## The fix

Two changes, because either alone leaves the defect in place.

The clause now compares only the node's existing state, with `demand_add` still added to `expected_demand` by the same UPDATE so it counts against the next decision:

```
cpu_load_1 + expected_demand <= SCHEDULER_TARGET_LOAD x cpu_schedulable
```

Both sides are now node state in units of runnable threads. Check-then-charge is safe because the comparison and the increment are one guarded UPDATE in one transaction, so concurrent admissions against a node serialise. And `SCHEDULER_DEMAND_PER_VCPU` drops to 0.6.

Retuning alone would still leave 22 of the 80 node-size by instance-size cells unsatisfiable, which is why the clause changes too. The runner-up design — flooring the budget at `demand_add` — is a smaller diff but keeps a per-request term on a per-node budget, so it patches over the units error rather than correcting it. That reasoning is decision E2 in the phase plan, and it is the one most worth arguing with.

`_capacity_dimension()` gains a `charged` flag, false for demand only, so the reported `exceeded` is the test the guard actually made. This is not just diagnostics: `CapacityAdmissionDenied.demand_only` is derived from the exceeded set, so a wrongly charged demand dimension would fire the P9 waiver on denials it must not touch.

## Verification

Full suite: **3,364 tests, 0 failures**, including the live MariaDB modules. `pre-commit run --all-files` clean.

The phase plan predicted, from the arithmetic and before any code was written, that the satisfiability sweep would refuse 54 of 80 cells against the pre-fix code. Mutation-tested against a real server, it refuses exactly 54. Both new unit tests also fail against the pre-fix clause.

The live suite was run against **MariaDB 11.8**, past the 11.6.2 boundary where `innodb_snapshot_isolation` turns a transaction's leading SELECT into `ER_CHECKREAD`. CI runs 10.11 and cannot see that (#3759), so the whole 364-test capacity suite passing there is worth more than the new tests alone.

## Two departures from the plan

Both recorded in the phase plan's step notes rather than smoothed over.

The satisfiability sweep could not live in step 1 as planned: `test_mariadb_capacity_admission.py` runs against a mocked connection and asserts on compiled statement text, so it can show the charge has left the WHERE clause but cannot evaluate whether the arithmetic admits anything. It moved to the live suite, where the arithmetic that broke actually lives.

Two existing live tests had premises the fix invalidated — one published a load that only exceeded the budget once the charge was added, the other crossed the budget on its fifth round and now needs a sixth. Both were restated rather than relaxed: same facts, at loads and round counts true of the new clause. The unit suite passed throughout, because the live modules skip without a database, which is precisely why the phase wanted live coverage.

## Still outstanding after this merges

- **Step 5**, operator: deploy to sfcbr and run one CI burst, discharging phase 3's step 9 soak, phase 4's step 10, and phase 4a's validation together (decision E6 — soaking claim accounting while every placement took the waiver path would have measured the waiver). Five observations to record separately.
- **Step 6**: flip the phase 3, 4 and 4a statuses and close #3813, gated on step 5.

## Scope guards

No diff to `CLAIM_ENFORCEMENT_HARD`, and the two duplicated `place_walk` copies are untouched — both are phase 5's.


mikalstill:
Addressed in afef6bede. Five items changed code, two were documentation accuracy, one is declined with reasoning.

**1 [FIX] MockMariaDB demand comparison** — fixed. Correct and worth catching: the mock charged the placement where the real guard no longer does, so the P9 waiver tests in `test_external_api.py` were exercising a walk production no longer takes. It is now check-then-charge, with the coupling to `demand_only` named in both `_mariadb_admit_instance_placement()` and `set_node_capacity()`. Added the caller-side test you suggested — `test_an_idle_node_admits_a_large_instance_on_the_first_walk`, an idle two-thread node admitting an 8-vCPU create with no waiver event. Mutation-tested: it fails against the un-updated mock.

**2 [DOC] Demand residue survives release** — fixed. You are right that the guidance was optimistic, and the sentence pointing operators at the wrong diagnosis was one I wrote. The guide now states that `expected_demand` is not credited back on release and only decays at the reconcile pass, gives `scheduler_capacity_node_expected_demand` against live instance count as the way to tell residue from load, and splits the waiver-event reading into two cases. Also added to Future work for step 5 to measure, with crediting-on-release and a shorter reconcile interval as the options.

**3 [DOC] Boundary case** — fixed in all three places, including phase 4a's own E3, which carried the same imprecision.

**4 [CONSIDER] `demand_add` unused** — taken. Your argument that the plan's "keep the parameter" instruction was about the RPC rather than the private helper is right, and an unused argument documented as deliberately unused is exactly how it gets wired back in. The signature is now the guard: `test_the_demand_clause_does_not_charge_the_placement` asserts `demand_add` is not a parameter, and that the only floats bound into the comparison are the target load and the coalesce default. The plan's Definition of done was updated to match, since it previously required the parameter to stay.

**5 [CONSIDER] `charged` positional** — taken, keyword-only. I considered deriving it from an `UNCHARGED_DIMENSIONS` frozenset as you offered; keyword-only keeps the divergence visible at the call site, where the comment explaining it already lives, rather than making a dimension's behaviour depend on its name from a distance.

**6 [CONSIDER] Degenerate sweep axis** — this was the best catch in the review, and it caught me overstating evidence. Rather than keep the redundancy and soften the prose, I split it: a clause-level sweep over 16 node sizes, and `test_an_idle_small_node_admits_every_instance_size`, which drives five instance sizes through a real `_direct_admit_instance_placement()` on a two-thread node. That puts the instance-size axis where it can still change the outcome, and answers your separate point that `sa.select(clause)` never proved an idle node actually admits. Both plans now describe what the tests run; the 54-of-80 mutation figure holds for the two together.

**7 [CONSIDER] Re-read can suppress a waiver** — documented, not fixed, per your own read. Added to the phase plan's risks as the false-negative direction the existing risk did not cover, with the empty-`exceeded`-set fix recorded in Future work and step 5 tasked with looking for unexplained 507s. I did not want to change `demand_only`'s semantics on a race I cannot produce, in a phase whose job is to make the clause satisfiable.

**8 [CONSIDER] No functional-CI coverage** — declined, with the reasoning in Future work. The suggested assertion (no `waiving demand guard` event on a create) would flake: a CI cluster legitimately at target during a burst fires that event correctly, so the test would fail for the right reason. The property is now covered at the database level by the live suite, which does run in CI, and at the caller level by the new mock-based test; real-hardware proof stays with step 5.

Also took the third item from your Test Coverage section — `test_an_allocation_denial_with_quiet_demand_is_not_waivable`, since the existing under-target test passed for the weaker reason that nothing at all was exceeded.

Full suite: 3,367 tests, 0 failures, including the live MariaDB modules against 11.8. `pre-commit run --all-files` clean.

PR #3846, opened 2026-08-22
Make await_timeout bound the whole instance create.

mikalstill:
The caller half of shakenfist/client-python#369.

`sf_instance` builds its client with `ASYNC_BLOCK`, so `create_instance()`
waited up to an hour for the instance to leave `creating`, returned it still
transitional, and only then did `await_instance_create()` start its own
`await_timeout` clock on the same condition. The task took the sum — about
4200 seconds — and failed with `not created within 600 second timeout`,
which is honest about the await and silent about the hour before it.

3600 + 600 = 4200, against 4206 / 4209 / ~4218 observed on
shakenfist/kerbside#284. See shakenfist/kerbside#355 for the full analysis.

## The change

Ask the client not to wait when we are going to await ourselves, so
`await_timeout` is the budget for the operation rather than for its second
leg.

## The compatibility wrinkle, which is the interesting part

That `timeout` argument is new in client-python, and
`deploy/collection/requirements.txt` requires `shakenfist-client`
**unpinned**. The control node's client is whatever pip resolved and may
predate it — and passing an unknown kwarg there is a `TypeError` that would
stop every instance creation in the fleet, including the under-cloud
playbooks this same module drives.

So two mechanisms rather than one:

* **Feature-detect** with `inspect.signature` before passing the argument.
  Guessing a version to test against would itself be a guess about
  backports.
* **Deduct** whatever the create spent from the await budget. This is what
  an older client gets: it cannot bring the task down to `await_timeout`,
  but it does stop the two waits being added together — 3600 rather than
  4200.

Also guards the elapsed time on the path where no create happens: an
instance needing no replacement gets the whole budget for its await.

5 new tests covering both client generations; 13 pass in that module.

## Landing

Either this or shakenfist/client-python#369 can land first — that is what
the feature detection buys. This one is what actually reaches kerbside CI,
since its oVirt lane checks this repository out at `develop` and builds the
collection from it (`tools/install-collection.sh`); the client comes from
pip separately, which is why the deduction fallback matters.

Refs shakenfist/kerbside#355.

https://claude.ai/code/session_01MXko24m6vN6o9P1q44gDyx

### DIFF
--- a/docs/user_guide/ansible.md
+++ b/docs/user_guide/ansible.md
@@ -215,8 +215,8 @@ Delete a network:
 | metadata<br/>*dictionary* | Metadata key-value pairs to set on the instance. |
 | state<br/>*string* | The state of the resource. Valid states are `present` or `absent`, defaults to `present`. |
 | uuid<br/>*string* | The UUID for the instance. Either `name` or `uuid` must be included in all requests with `state: absent`. If you specify a UUID and the instance does not exist in the Shaken Fist cluster, this argument will be ignored as UUIDs are randomly assigned on network creation. |
-| await<br/>*boolean* | Whether to wait for the instance to be created. Only works for when state is `present`. Default is `false`. |
-| await_timeout<br/>*integer* | How many seconds to wait in an `await`. Defaults to 600. |
+| await<br/>*boolean* | Whether to wait for the instance to be created. Only works for when state is `present`. Default is `false`. The wait is bounded by `await_timeout`, as is the creation which precedes it. |
+| await_timeout<br/>*integer* | How many seconds an `await` may take, counted from the start of the operation rather than from the start of the wait. Where an instance is being replaced, the time spent deleting the old one and creating the new one is deducted from this number before the wait begins. It is a budget rather than a hard deadline: a deletion or creation already underway is not interrupted, and a `shakenfist-client` of v0.8.3 or earlier cannot be told not to wait while creating. Defaults to 600. |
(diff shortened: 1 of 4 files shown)


mikalstill:
All four `fix` items, the `document` item, and five of the six `consider` items are addressed in 7d0be9c.

Notable changes since the last review:

* The budget now starts before `_delete_and_wait()` rather than at the create, so a replacement spends one budget rather than two, and the `DOCUMENTATION` is rewritten to be honest that this is a budget rather than a hard deadline.
* client-python#369 has landed, so `timeout=0` is now a verified fact rather than an assumption: the client tests `if timeout is None`, and `await_instance_create()` checks the instance once before consulting its clock. Both are recorded at the decision site.
* An exhausted budget fails with the module’s own message naming where the budget went.
* Which create path ran is now logged, and durations use `monotonic()`.
* `exit_json` and `fail_json` raise distinguishable exceptions in the tests, and the fake clock is an explicit pair of readings.
* `ansible_module_ci/004.yml` asserts the budget was reported, which is the only place the real client is exercised on this path.

Item 6 (ASYNC_CONTINUE) is declined, with the reasoning left in a comment as suggested: the async strategy is a property of the client, and this same client runs `_delete_and_wait()`, where `delete_instance` blocks today.

Nine mutations of the module were checked to fail the suite, including one that bails out before the create — which the previous test harness did not catch, exactly as item 9 predicted.

Issue #3857, opened 2026-08-23
baseobject.upgrade() raises AttributeError on an object written by a newer node


mikalstill:
## Summary

`DatabaseBackedObject.upgrade()` cannot read an object whose stored
version is *higher* than the reading node's `current_version`. It
raises `AttributeError` from inside a loop looking for an upgrade step
that will never exist, and the guard written to catch exactly that
case is dead code.

This is the "old node reads a row written by an already-upgraded node"
half of a rolling upgrade, and
[docs/operator_guide/upgrades.md](../blob/develop/docs/operator_guide/upgrades.md)
currently promises rolling upgrades are safe without qualifying it:

> This means it is safe to perform a rollout across a cluster without
> downtime, although you might see small transient failures such as
> single API requests failing as processes restart.

Found while planning phase 2 of
`docs/plans/PLAN-agent-operation-deadlines.md`, which declined an
object version bump because of it. Nothing is currently broken —
this is about what happens the next time anyone bumps a
`current_version`.

## Reproduction

```
$ python3 -c "
from shakenfist.operations.agentoperation import AgentOperation
o = AgentOperation.__new__(AgentOperation)
print('current_version =', AgentOperation.current_version)
o.upgrade({'uuid': 'aaaabbbb-0000-4000-8000-00000000000a', 'version': 4})
"
current_version = 3
AttributeError: 'AgentOperation' object has no attribute '_upgrade_step_4_to_5'
```

## Why

`shakenfist/baseobject.py`, in `upgrade()`:

```python
while static_values['version'] != self.current_version:
    step = '_upgrade_step_%d_to_%d' % (static_values['version'],
                                       static_values['version'] + 1)
    step_func = getattr(self, step)
    if not step_func:
        raise exceptions.UpgradeException(
            'Upgrade step %s is missing for object %s'
            % (step, self.object_type))
```

Two things go wrong together:

1. The loop condition is `!=` rather than `<`, so a version *above*
   `current_version` walks upwards forever rather than being
   recognised as "this row is from the future".
2. `getattr(self, step)` has no default, so it raises `AttributeError`
   before the `if not step_func` guard can run. That guard — and the
   `UpgradeException` it was written to raise — is unreachable.

The Pydantic path gets (2) right and (1) wrong:
`upgrade_pydantic_data()` in the same file uses
`getattr(cls, step_name, None)` and raises the intended
`UpgradeException`. So the missing default in `upgrade()` looks like a
[...]

PR #3858, opened 2026-08-23
Agent operation deadlines phase 2: schema

mikalstill:
Phase 2 of [PLAN-agent-operation-deadlines](docs/plans/PLAN-agent-operation-deadlines.md). Schema only: it adds the four columns the rest of the plan needs, carries them across all three MariaDB access layers, and proves the migration against a real server. Nothing enforces a deadline yet, and no client can set one until phase 3.

## What lands

`agent_operations` gains `deadline` and `progress_timeout`; `agent_operation_attributes` gains `last_progress` and `attempts`. Both table versions go 2 -> 3 with additive `ADD COLUMN IF NOT EXISTS` migrations. The direct, gRPC and servicer converters all carry the new fields, `AgentOperation` grows matching properties, and `external_view()` returns all four.

## What the survey found

Three findings. Two of them contradict the master plan, and both were corrected at their source in this branch rather than left for the next reader to trip over.

**The plan's NULL semantics were self-inconsistent.** It had NULL meaning "no deadline" on one column and "server default" on the other. This branch makes NULL mean the same thing in all four places — *no client intent was recorded, so the server default applies* — and reserves an explicit `0.0` for "the client asked for none". That ordering matters for safety: NULL is what every legacy row and every row written by a not-yet-upgraded API node contains, so NULL has to be the reading that produces a working timeout rather than the one that disables it. `0.0` is an unambiguous sentinel because a real deadline is an absolute timestamp of order 1.7e9.

Expressing three values over a wire that has two required proto3 explicit field presence — the `optional` keyword, with `HasField()` on the read side. The precedent is `NamespaceKeyAttributesProto.expiry`. Every daemon except `sf-database` itself reaches MariaDB across that boundary, so a converter that collapsed NULL into `0.0` would have left the direct path correct and the path everything actually uses wrong.

**The plan's proto3 assumption was wrong.** It warned that a proto3 `optional` field cannot be set to `None` from Python and briefed a conditional-assignment workaround. Checked against the regenerated bindings: passing `None` leaves the field unset and `HasField` false, which is exactly the wanted behaviour. The values are passed straight through.

**The planned object version bump was declined.** `current_version` stays at 3 and there is no `_upgrade_step_3_to_4`. `baseobject.upgrade()` walks `while version != current_version`, so a row written by an upgraded node is walked *upward forever* by an older one, and `getattr(self, step)` with no default raises `AttributeError` before the `UpgradeException` guard at `baseobject.py:208-211` can fire — that guard is unreachable. Agent operations are the one object type whose writer (sf-api) and reader (sf-sidechannel, on the hypervisor) are reliably different machines, so a bump would break rolling upgrades for precisely the object that cannot tolerate it. The columns are additive and default to NULL, so nothing needs the bump anyway.

That `baseobject` defect is real but is not this phase's to fix — it is filed as #3857, labelled `automated-fix-attempted` to reserve it for a planning session, since choosing between its three candidate fixes is a design decision about mixed-version cluster behaviour and likely changes what `docs/operator_guide/upgrades.md` promises.

## The decision most worth arguing with

Declining the version bump. The counter-argument is that skipping it leaves the object version no longer describing the row's shape, so a future reader cannot tell a phase-1 row from a phase-2 one. The answer is that the columns are nullable and additive, so shape is discoverable from the data itself, and that a bump would trade a cosmetic invariant for a real rolling-upgrade break in the one object type that spans machines.

## Verification

`pre-commit run --all-files` clean. **3,290 unit tests pass.** All **116 live-MariaDB tests** pass — the 10 new ones plus every pre-existing live module — against a disposable MariaDB 11.4 container.

The live tests cost no CI wiring: `tools/ci-enum-widening-test.sh` discovers live modules by regex on purpose, and fails the job if zero live tests ran.

Every load-bearing assertion was mutation-tested. Four of five mutations were caught. The fifth is worth stating plainly rather than hiding:

**Removing `DEFAULT 0` from the `attempts` ALTER did not fail any test.** MariaDB fills a new `NOT NULL` numeric column with its implicit type default instead of rejecting the ALTER, so the tests could not have caught it. The `DEFAULT 0` stays — it makes the behaviour deterministic across `sql_mode` settings rather than dependent on them — but three places that claimed the tests proved it were corrected: the migration comment, the test comment, and the live module docstring.

Two of the phase plan's own definition-of-done checks were also wrong when written and were rewritten in the closing commit: one counted literal `ALTER TABLE` statements where the implementation loops over column-definition tuples, and the other used a `sed` range whose start pattern also matched its end line, silently swallowing the second function.

## Two gotchas the live tests hit

Raw SQL against the uuid columns must use the **undashed** form (`uuid.hex`). These are `sa.Uuid` / CHAR(32) columns, unlike `object_states.object_uuid` which is dashed — the dashed form silently matches nothing rather than erroring. This is CLAUDE.md's "two uuid formats" pitfall, and there is now a docstring warning on it.

Two `external_view()` tests needed `references_to_grouped_dict` mocked, because it reaches `get_jwt_identity()` and so needs a Flask application context. That requirement predates this work and is noted as such in the test.

## Scope guards

No enforcement code, no API parameters, no config defaults — those are phases 3 and 4. `AGENT_OPERATION_EXECUTION_TIMEOUT` is untouched.

The API reference examples show `null` and `0` rather than plausible-looking values, because until phase 3 ships no client can set a deadline and until phase 4 ships nothing writes progress. Those *are* the honest values today.

https://claude.ai/code/session_01AJb2ff4gsvsFcJqfp7kruV


mikalstill:
Review addressed in f45ef2cd9. Summary of what was taken and what was not.

## 1 — [FIX] swallowed ALTER, version bumped anyway — **fixed**

Correct, and the phase plan caused it. Step 2a's brief told the implementer to copy the `try/except` from `_ensure_instance_attributes_schema`'s v2->v3 block — but that block is a `DROP COLUMN IF EXISTS`, where a swallowed failure leaves a harmless leftover column. The v1->v2 `ADD COLUMN` block directly above it has no handler at all, and the review is right that the asymmetry is deliberate. The brief copied the wrong one of two adjacent precedents.

Both migration blocks now let the ALTER propagate, with a comment recording why: `IF NOT EXISTS` already makes the benign case a no-op, so the handler could only ever have caught genuine failures, and bumping the version past a column that was never added produces a schema which passes `verify_schema_versions()` while every insert fails on an unknown column — and which re-running `ensure-mariadb-schema` can never repair.

New live test `test_failed_migration_does_not_advance_the_version` pins it, covering both ensure functions. It induces a real failure by removing the table out from under a version row that still claims 2. It asserts against `sa.exc.DatabaseError` rather than a leaf class, because that induced failure surfaces as `ProgrammingError` (unknown table) while a metadata lock timeout would be `OperationalError`; the test is about the version, not which leaf was raised.

Mutation-tested: restoring the `try/except` makes it fail with `end_version: 3, migrated: True` — exactly the described defect.

## 2 — [CONSIDER] no coverage of the `_grpc_*` client half — **taken**

Added `AgentOperationGrpcClientTestCase`, six tests. Rather than asserting on the client in isolation, it patches the stub, captures the request protos `mariadb.py` actually builds, and feeds them to the servicer converters — a client -> wire -> server round trip, so a client that set presence differently from `_agentop_to_proto` is caught rather than hidden by testing one side twice.

That also closes the fourth missing-coverage item: `test_update_mask_survives_the_round_trip_to_the_direct_layer` drives `_grpc_update_agent_operation_attributes(data, fields=['results'])` through `UpdateAgentOperationAttributes` and asserts the mask arrives at `_direct_update_agent_operation_attributes`. Deleting `fields=fields or []` from the client makes it fail with `['results'] != []`.

## 3 — [CONSIDER] `_attributes()` can return None — **taken**

`_attributes()` now returns early when the create succeeds, and falls back to the freshly constructed defaults when it loses the race *and* the re-read comes back empty. New unit test `test_attributes_survive_a_lost_get_or_create_race`; reverting the fallback makes it fail with `AttributeError: 'NoneType' object has no attribute 'results'`.

Agreed this is pre-existing rather than introduced here, and that consolidating four readers onto one helper is the right moment to add the guard.

## 4 — [INFO] old sf-database unrestartable after the migration — **no action, agreed**

Inherent to every table version bump given the strict-equality check, and already the documented procedure. Recorded, not changed.

## 5 — [INFO] stale definition-of-done grep — **replaced rather than tweaked**

The finding understates it: the check was not merely stale, it was unfalsifiable. `deadline` appears about a dozen times in `shakenfist/daemons/` as gRPC call-deadline prose, so a bare word match could never have printed zero, before or after this branch. Excluding the servicer would not have rescued it.

Replaced with a match on attribute access, excluding the four transport converters step 2b required:

```
grep -rnE '\.(deadline|progress_timeout|last_progress|attempts)\b' \
    shakenfist/daemons/ --include='*.py' \
    | grep -vc '^shakenfist/daemons/database/main.py:'
```

This prints `0` today, and adding a single `op.deadline` to a daemon takes it to `1` — mutation-tested, so the replacement is falsifiable in the way the original was not.

## Also addressed

The fresh-`create_all` nullability gap from the test-coverage list: `test_migration_runs_on_a_freshly_created_schema` now asserts nullability on all four columns. The migrated path and the greenfield path are produced by different code — a hand-written ALTER versus `create_all()` from the pydantic model — so the existing `test_migration_preserves_nullability` said nothing about a fresh install.

Step 2a's wrong brief is recorded under a new **Departures from the plan** section in the phase plan rather than quietly edited away.

## Verification

`pre-commit run --all-files` clean. **3,414 unit tests pass**, and **all 117 live-MariaDB tests pass** against MariaDB 11.4.

One note for anyone reproducing the live suite locally: use the `mariadb+` dialect, not `mysql+`. `tools/ci-enum-widening-test.sh` uses `mariadb+mysqldb://`, and under `mysql+pymysql://` two unrelated tests in `test_mariadb_enum_columns_live` fail on an `INET4` compile error for `ipam_reservations.address`. That is a dialect artefact rather than a regression — confirmed by reproducing it with this branch's changes stashed.

Four mutations tested, all four caught: re-swallowing the ALTER, dropping the gRPC field mask, reverting the `_attributes()` fallback, and adding a fake enforcement consumer to a daemon.

Issue #3864, opened 2026-08-23
A completed cluster operation's events are unreachable 30 seconds later, so its audit trail cannot be read


mikalstill:
Found while planning step 7 of `docs/plans/PLAN-queue-performance.md`, which
needed to read back the per-operation queue-wait events that step 1 of that plan
added.

## The behaviour

A cluster operation is hard deleted 30 seconds after it reaches a final state:

* `_deleted_object_delay()` returns 30 for any object type whose name ends in
  `_op` (`shakenfist/daemons/cluster/scheduled_tasks.py:735`), against
  `CLEANER_DELAY` for everything else.
* `FINAL_OBJECT_STATES` includes `complete` (`shakenfist/constants.py:191`), so
  the clock starts when the operation succeeds, not when it is deleted.
* `DatabaseBackedObject.hard_delete()` calls `delete_object_events()`
  (`shakenfist/baseobject.py:689`), which deletes every `event_objects` row for
  that object (`shakenfist/mariadb.py:5835`).

The `events` rows survive that, but orphaned: nothing joins them to an
operation, a queue or a target any more, and the daily prune sweep
(`_direct_prune_orphan_events`, `shakenfist/mariadb.py:5639`, scheduled at
`shakenfist/daemons/cluster/main.py:619`) then deletes them.

So every event an operation ever emitted — the audit events for its tasks, the
`execution duration` event carrying its latency, the `coalesced sibling ops`
event — is readable for about thirty seconds and then is not. An operator
looking at an operation that completed a minute ago has nothing to read.

## Why it matters beyond the audit trail

`docs/operator_guide/networking/overview.md` documents the dispatcher's
diagnostic events as an operator-facing debugging tool, and stated that the
`execution duration` event is retained for `MAX_USAGE_EVENT_AGE` (30 days).
That was wrong in both directions and is corrected by the queue-performance
step 7 change. But the correction only documents the gap; it does not close it.

Working around this meant reading the events out of the log stream instead
(`eventlog.add_event_multi` echoes each event as an `Added event` log line when
`LOG_EVENTS_TO_LOKI` is set, which is the default), which is what
`tools/queue-wait-report.py` does. That works, but it means the authoritative
record in MariaDB is the *less* durable of the two copies, which is backwards.

## Things worth deciding

* Whether an operation's own events should outlive it at all, and if so for how
  long — `MAX_USAGE_EVENT_AGE` and friends already express that intent for every
  other object.
[...]

Issue #3863, opened 2026-08-23
A dependency wait costs a flat 15 seconds on the user_waiting queue, and dominates instance start latency


mikalstill:
Measured while closing out step 7 of `docs/plans/PLAN-queue-performance.md`, which
re-measured cluster operation queue latency after the coalescing work landed.

## What was measured

25h58m of `execution duration` events from `sfcbr` (17,936 samples,
2026-08-22T16:33Z to 2026-08-23T18:30Z), plus a 33 minute cluster CI window
(1,248 samples) from run 32597511463. Both were produced with
`tools/queue-wait-report.py`.

The queue-wait tail this plan set out to remove is gone: `net_op` now runs at a
p50 wait of 0.78s and a p90 of 1.81s, against a starting point of over 60
seconds. What is left is not queueing.

| Operation type | n | wait p50 | wait p50 (never deferred) | defers |
|---|---|---|---|---|
| `node_inst_netdesc_op` | 1013 | **15.78s** | **0.77s** | 962 of 1013 |
| `node_inst_net_iface_op` | 28 | 15.51s | 0.67s | 18 of 28 |

Excluding deferred operations collapses the wait to the poll floor, so
essentially all of it is deferral. The same shape appears in CI
(`node_inst_netdesc_op` p50 15.52s falling to 0.13s undeferred).

## Why

`BaseClusterOperation.defer()` takes `delay: int = 15`
(`shakenfist/operations/baseoperation.py:390`), and `sf-queues` takes that
default: `shakenfist/daemons/queues/workitem.py:152` calls
`op.defer(waiting_on=[dep_op])` with no delay.

`sf-net` does not, and an earlier version of this issue said it did. Its
`_apply_defer` (`shakenfist/daemons/network/workitem.py:95-101`) passes an
explicit delay which starts at `INITIAL_DEFER_DELAY` (0.1s) and doubles to a
`MAX_DEFER_DELAY` cap of 15s, so a defer on a queue `sf-net` drains costs
well under a second until it has retried several times. That does not weaken
this issue: `node_inst_netdesc_op` and `node_inst_net_iface_op`, the two
operation types the numbers above are about, both run on the per-node
`-clusteroperation-` queues that `sf-queues` drains. It does narrow it — the
fix belongs in the `sf-queues` dependency-wait path, and `sf-net` is a
worked example of what the fix could look like.

So an operation whose dependency is not yet complete is re-enqueued a flat 15
[...]

PR #3865, opened 2026-08-23
Queue performance step 7: measure the queue, and decline fairness

mikalstill:
Closes out `docs/plans/PLAN-queue-performance.md` by doing its step 7 --
re-measuring cluster operation queue latency now that the coalescing work has
been in production for three months, and deciding the fairness question it left
open.

## The verdict

**The tail this plan set out to remove is gone, and explicit fairness is not
needed.**

`net_op`, the family containing `network_apply_update_dnsmasq`, against a
starting point of over 60 seconds:

| Window | n | p50 | p90 | p99 | max |
|--------|---|-----|-----|-----|-----|
| `sfcbr`, 25h58m | 6310 | 0.78 | 1.81 | 7.77 | 27.60 |
| Cluster CI, 33m (merge run 32597511463) | 313 | 0.56 | 1.72 | 2.19 | 23.96 |

A p90 of 1.8s *is* `IDLE_POLL_MAX_SECONDS`, so the median operation now waits
one dispatcher poll interval and nothing else.

## Three tails, none of them starvation

Each was excluded on evidence rather than argument, which is the part of this
worth reviewing:

* **`user_waiting`, p50 15.78s.** Entirely deferral. Restricted to operations
  which never deferred, the same p50 is **0.77s**, and 962 of 1013 samples had
  deferred at least once. A dependency wait re-enqueues an operation a flat 15
  seconds into the future.
* **`background_high_io`, p90 403s, defer count zero** -- which is the shape
  starvation would have. It is not: during each of the eight worst waits, the
  *same queue* completed between 1,281 and 1,334 seconds of its own work (more
  than the wait, because several workers run concurrently). A starved queue
  does no work during its wait; this one was saturated throughout.
* **`networknode`/`background`, p50 4.45s.** Bursts. In the largest, the
  background lane ran 34.5s of its own work in a 29 second span while
  `networknode`/`user_facing` ran 2.9s in total.

So the `FIELD()` ordering's theoretical starvation risk did not materialise in
26 hours of production traffic which contained exactly the bursty contention it
would appear in. Bounded-staleness ordering and reserved slots are not being
added, and the tool is committed so that reopening the question means bringing
a measurement.

## What is in the change

* `tools/queue-wait-report.py` and its unit tests. It reads Shaken Fist JSON
  log lines on stdin rather than talking to Loki, because the same lines arrive
  from Loki, from `journalctl` on a node, and from the CI artifact bundle, and
  a stdin tool works with all three. Every percentile is reported twice, once
  over all samples and once over operations which never deferred -- that split
  is the point of the tool, not a nicety, since without it a dependency wait
  reads as a queueing problem.
* The phase plan, its findings, and the master plan's outcome section.
* A documentation fix. `docs/operator_guide/networking/overview.md` claimed the
  `execution duration` event is retained for `MAX_USAGE_EVENT_AGE` (30 days).
  It is not: an operation is hard deleted 30 seconds after it completes and
  `hard_delete()` drops the `event_objects` rows joining its events to it, so
  the events are unreachable within a minute and the orphaned rows are pruned
  daily. The page now says that, and says to read the log stream instead.
* The master plan gains an Execution table, so its index arithmetic is
  recomputed by `tools/check-plan-status.py` rather than counted by hand, and
  it leaves the checker's `HAND_COUNTED` list.

## Filed, not fixed

Two latency sources the phase measured and deliberately left alone:

* #3863 -- a dependency wait costs a flat 15 seconds on the `user_waiting`
  queue, and is now the largest user-visible latency in the sample.
* #3864 -- a completed operation's events are unreachable 30 seconds later, so
  its audit trail cannot be read. Working around this is why the measurement
  had to come from the log echo.

A third candidate, the `background_high_io` tail, was conditional on the
measurement confirming it was not self-inflicted. It found that it was, so
there is nothing to file.

## Review notes

* The three exclusions above are load bearing. If one of them is wrong, the
  verdict changes.
* The `background_high_io` exclusion deviates from the plan, which asked for a
  disk-busy correlation. The same-queue-occupancy test answers the same
  question from the log stream alone and is a stronger negative -- it shows the
  queue was busy, not merely that the gate was open.
* No production code changes. The only non-documentation file is a new
  standalone analysis tool which nothing imports.

https://claude.ai/code/session_01G31Z9FWwYcbHYgt7YB3427

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
         description: Run flake8 style checks on changes since HEAD~
         entry: tox -e flake8
         language: system
-        files: ^shakenfist/
+        files: ^(shakenfist|tools)/
         pass_filenames: false
         types: [python]
 
@@ -52,7 +52,7 @@ repos:
         description: Run Python 3 unit tests with stestr
         entry: tox -e py3
         language: system
-        files: ^shakenfist/
+        files: ^(shakenfist|tools)/
         pass_filenames: false
         types: [python]
 
--- a/docs/operator_guide/networking/overview.md
+++ b/docs/operator_guide/networking/overview.md
@@ -514,9 +514,18 @@ itself):
       terminal state write).
     * `wait_seconds` — time between when the op was first inserted
       into `cluster_operations` and when the dispatcher claimed it.
-      Large values (>10 s) mean the queue was backed up; very
-      large values point at either worker saturation or a stuck
-      op blocking the worker.
+      This is not queueing alone: it also counts deliberate
+      deferral while an op waits on a dependency, and the seconds a
+      `*_high_io` background op spends gated off while the local
+      disk is busy. Both are designed behaviour, and each is large
+      enough to be the whole of a tail on its own -- the worst
+      operation type measured for the queue-performance plan had a
+      15.78 s median wait which fell to 0.77 s once deferred ops
+      were excluded. So read a large value against `defer_count`
+      and against the node's disk busy metric before calling it a
+      backup; only once those are excluded does it point at worker
+      saturation or a stuck op blocking the worker. See "Reading
+      these events back" below.
     * `defer_count` — how many times this op was re-enqueued via
       `defer()` / `defer_with_backoff()` before finally running. A
       first-time pickup is `0`; non-zero values indicate dependency
@@ -574,11 +583,17 @@ are orphaned rows which the daily prune sweep then deletes. The
 never come into it -- they bound how long an event may live, not how
 long the object it describes does. This is issue #3864.
 
-What does persist is the log stream. Every event is echoed as an
-`Added event` log line carrying the whole `extra` dict, unless
-`LOG_EVENTS_TO_LOKI` has been turned off, so wherever the cluster's
-logs are shipped is where an operation's history can still be read
-after the fact.
+What does persist is the log stream. Every event is echoed as a log
+line carrying the whole `extra` dict, unless `LOG_EVENTS_TO_LOKI` has
+been turned off, so wherever the cluster's logs are shipped is where
+an operation's history can still be read after the fact.
+
+Do not grep that stream for `Added event`. That is the message the
+echo is written with, but the logger merges the caller's fields over
+the record last and one of those fields is the event's own `message`,
+so the string never reaches the shipped JSON. Filter on the event's
+message instead -- `execution duration` for the events above -- which
+is what the record's `message` field actually holds.
 
 `tools/queue-wait-report.py` in the source repository summarises the
 `execution duration` events from such a stream: give it Shaken Fist
--- a/docs/operator_guide/networking/overview.md
+++ b/docs/operator_guide/networking/overview.md
@@ -560,6 +560,33 @@ itself):
   one, so a moderate gap between sibling-arriving and
   `coalesced sibling ops` firing is expected.
 
-The `execution duration` event is retained for
-`MAX_USAGE_EVENT_AGE` (default 30 days); the `coalesced sibling
-ops` event for `MAX_STATUS_EVENT_AGE` (default 7 days).
+### Reading these events back
+
+Neither event survives its operation for long, so plan to read them
+from the log stream rather than from the database.
+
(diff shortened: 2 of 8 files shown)


mikalstill:
Addressed in 4b11f1ca0. Every claim was checked against the tree before acting on it; item 2 needed a correction to the review itself, and one item turned up something the review did not catch.

**No measured number changed.** I re-ran the original 26 hour capture through the corrected tool: `node_inst_netdesc_op` 15.78s falling to 0.77s over 1013/51 samples, `node_blob_op` p90 403.56s, `networknode`/`background` 4.45s, `net_op` p50 0.78s / p90 1.81s — identical, and still zero `unknown` rows. The grouping key was always right; only the labels and the notes were wrong.

### Fixes

**1. `per-network`** — confirmed and fixed. Now `per-node (cluster op)` and `per-node (network)`, with the docstrings, the module comment and the tests saying the split is by draining dispatcher rather than by target scope.

**2. The flat-15s note** — confirmed, with one correction to the review. The schedule follows the *dispatcher*, and `sf-net` drains `networknode-clusteroperation-*` as well as `<uuid>-network-*` (`shakenfist/daemons/network/workitem.py:174-176`), so those cluster-wide queues get the 0.1s backoff too — the suggested rule "a defer on a `-clusteroperation-` queue is a flat 15s" is not quite right. The tool now prints a per-class schedule table, and `test_every_queue_class_has_a_defer_schedule` fails if a class is added without one. Corrected in the phase plan's survey, the master plan's finding, and in **#3863**, which cited `_apply_defer` as an example of "calls it with no argument" when that is the backoff path.

The capture corroborates the correction from the other side: all 980 deferred samples in 26 hours are on queues `sf-queues` drains, and every `sf-net` lane records a defer count of zero — so nothing measured here was ever attributed to the wrong schedule.

**3. "These six"** → "These five".

**4. The `wait_seconds` bullet** — amended to say it counts deliberate deferral and the disk-busy gate, that either can be a whole tail on its own (with the 15.78s → 0.77s number), and to read it against `defer_count` before calling it a backup.

### Considered, and taken

**5.** `--min-samples` now prints `(N rows with fewer than M samples omitted, K in total)` under each table, including when it empties one. Help text corrected.
**6.** `load_json_object()` retries at each subsequent `{`, so a logcli label set no longer costs the whole capture. Test added.
**7.** The page now says not to grep for `Added event` and to filter on the event's `message` instead, with the reason (`JsonFormatter.format` does `obj.update(record.extra_fields)` last, and `add_event_multi` puts the event message in there).
**8.** Added `QueueWaitReportTestCase`: `main()` end to end, the empty-stdin path the definition of done names, `--min-samples` both ways, and banner/column alignment. Plus the missing `*_op`-less event and a group holding both populations. 13 tests → 22.
**9.** Took the second half — `tools/` added to the `files:` scope of the `flake8` and `py3` hooks, so an edit to this tool alone is now checked and tested locally. Left the `SkipTest` half; tox runs from a checkout.
**10.** `percentile()`'s docstring now says "the value at the nearest index to `fraction * (n - 1)`".
**11.** Took the label: `exec (all)`, which is safe now that banners size their columns (below).

### One the review did not catch

The banner row was genuinely broken. `wait_seconds (defer_count == 0)` is 31 characters over a 21-character span, so it overflowed and pushed `exec` off its own columns — the undeferred half of every row was mislabelled in any real run — and the line carried trailing whitespace. `column_widths()` now shares a banner's deficit across the columns it spans, and `test_columns_line_up_under_their_banners` covers it.

I mutation-tested the four load-bearing new assertions (banner padding, logcli parse, min-samples footnote, defer-schedule completeness); each fails when its fix is reverted.

### Filed rather than fixed

Verifying item 1 turned up a third thing: `get_node_background_node_queues()` lists `f'{node_uuid}-background'`, which has no family segment (`shakenfist/operations/baseoperation.py:70`). Every enqueue builds `f'{target}-{family}-{priority}'` (`shakenfist/schema/operations/util.py:53`), so nothing can ever write that name — yet it rides in every background dequeue's `IN` list and `ORDER BY FIELD()` on every node, matching nothing. Zero of the 19,184 measured samples had an unparseable queue name, which corroborates it. Filed as **#3867** rather than fixed here: this is a measurement phase and that is a production change.

Issue #3879, opened 2026-08-24
No functional CI coverage for cluster operation coalescing


mikalstill:
Found by the phase 8 push audit of `docs/plans/PLAN-queue-performance.md`, as a
contributing cause of #3878.

`grep -rln coalesc shakenfist/deploy/shakenfist_ci/` returns nothing. Cluster
operation coalescing — steps 4 and 5 of the queue-performance plan — has no
functional test anywhere.

The unit coverage that exists cannot substitute for it, because every layer is
mocked:

* `shakenfist/tests/operations/test_baseoperation.py` mocks
  `mariadb.claim_coalescible_siblings`, so it verifies the dispatcher decides to
  call the primitive, not that the primitive does anything.
* `shakenfist/tests/test_mariadb_work_queue.py` mocks `_get_engine`, so it
  verifies the shape of the SQL statement, not that it matches a row.

That combination let #3878 — a join which can never match, disabling coalescing
entirely — sit on `develop` for three months with a green test suite.

`CLAUDE.md` states the project's preference directly: "We prefer functional tests
to unit tests if we can only have one." This is a case where only the functional
test could have caught the defect.

## What the test should assert

The behaviour is already documented in
`docs/operator_guide/networking/overview.md`: enqueue duplicate coalescible work
against one target — the canonical case is several instances starting on the same
network, each enqueueing `network_apply_update_dnsmasq` — and assert that a
`coalesced sibling ops` event fires carrying `sibling_count`, `sibling_uuids` and
`tasks`, and that the folded siblings reach `complete` with the message
`coalesced into sibling op`.

The operator guide already names the absence of that event as the diagnostic for
a coalescing bug. Turning that sentence into an assertion is the fix.

*(Filed by Claude Code during the phase 8 push audit of the queue-performance plan.)*

mikalstill:
Closed by phase 9 of `PLAN-queue-performance.md`, landed in #3905.

## What was built

`shakenfist/deploy/shakenfist_ci/cluster_ci_tests/test_coalescing.py`
fires a burst of instances at a single network and asserts that a
coalescing event appears on that network — so the opening check in
this issue, `grep -rln coalesc shakenfist/deploy/shakenfist_ci/`, now
returns a file.

The assertion is deliberately "either coalescing event": `enqueue-side
dedup: reused pending op` **or** `coalesced sibling ops`. Both mean a
coalescing query matched a row, and #3878 broke both at once, so
[...]

PR #3880, opened 2026-08-24
Queue performance phase 8: audit the plan, and make coalescing work

### LINKED ISSUE #3878: Cluster operation coalescing has never worked: both SQL primitives join undashed to dashed uuids
Found by the phase 8 push audit of `docs/plans/PLAN-queue-performance.md`.

Cluster operation coalescing has never worked. Both SQL primitives join
`cluster_operations` to `object_states` on a uuid expression that cannot match,
so both always return "nothing found". Steps 4 (worker-side fold) and 5
(enqueue-side dedup) of the queue-performance plan are inert in production and
have been since PR #3194 merged on 2026-05-26.

## The defect

`cluster_operations.uuid` is `sa.Column('uuid', sa.Uuid())`, which SQLAlchemy
renders as `CHAR(32)` on MariaDB and stores **undashed**.
`object_states.object_uuid` is `sa.String(36)` and stores the **dashed** form.
This is the two-uuid-format pitfall documented in `CLAUDE.md`.

Both functions join them with a bare cast, which changes the declared width but
does not reinsert the dashes:

* `_direct_find_existing_coalescible_op` — `shakenfist/mariadb.py:22473`
* `_direct_claim_coalescible_siblings` — `shakenfist/mariadb.py:22576`

```python
states_table.c.object_uuid == sa.cast(
    cluster_ops_table.c.uuid, sa.String(36)),
```

The repository already has the correct helper for exactly this join,
`_dashed_uuid_expr()` at `shakenfist/mariadb.py:23595`, whose docstring says it
exists "so a comparison against object_states can use its primary key index".
Neither coalescing function uses it.

## Proof

```
stored cluster_operations.uuid  : 'aaaaaaaabbbb4ccc8dddeeeeeeeeeeee'
stored object_states.object_uuid: 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee'

JOIN rows with bare cast (as shipped): 0
JOIN rows with _dashed_uuid_expr     : 1
```

Reproduced against a real database by building both tables, inserting one
`cluster_operations` row through the SQLAlchemy `Uuid` bind processor and one
`object_states` row with the dashed string every caller passes to
`set_state()`, then running each join.

## Corroboration from production

`coalesced sibling ops` — the event `BaseClusterOperation.execute` emits on a
survivor when it folds siblings — has fired **zero times in seven days** on
`sfcbr`:

```
loki-query '{job="shakenfist"} |= "coalesced sibling ops"' --tenant sfcbr --since 168h
→ 0 lines
```

`docs/operator_guide/networking/overview.md` already names this as the
diagnostic: "A *complete absence* of these during a CI run that's known to be
enqueueing duplicate work would point at a bug in either the enqueue-side dedup
(`mariadb.find_existing_coalescible_op`) or the worker-side fold
(`mariadb.claim_coalescible_siblings`)." That prediction has been true for three
months and nothing was watching it.

## Why no test caught it

Nothing anywhere executes the join:

* `shakenfist/tests/operations/test_baseoperation.py` — all ten coalescing tests
  mock `mariadb.claim_coalescible_siblings`, so they test the caller's decision
  to call it, not the SQL.
* `shakenfist/tests/test_mariadb_work_queue.py` — `ClaimCoalescibleSiblingsTestCase`
  mocks `_get_engine` and returns canned rows from `fetchall`. Its own docstring
  says "the SQL-shape assertions cover the WHERE/JOIN clauses", which is exactly
  the limitation: it asserts the shape of the statement, never that it matches a
  row.
* There is no functional coverage at all —
  `grep -rln coalesc shakenfist/deploy/shakenfist_ci/` returns nothing.

## Suggested fix

Replace the bare cast in both functions with `_dashed_uuid_expr(...)`. That is
two lines, and it is the easy part.

The harder part is that this **activates a concurrency path which has never
executed in production**. The fold marks *other* workers' operations complete
and depends on the routing invariant documented at
`shakenfist/daemons/network/workitem.py:60-77` (ops sharing a target must land
on the same worker). That invariant has never been exercised with the fold
actually firing. Before or alongside the fix this wants:

* a regression test that runs the join against a real database rather than a
  mocked engine, and fails with the bare cast;
* functional CI coverage that asserts `coalesced sibling ops` fires when
  duplicate work is enqueued — the absence of which is why this went unnoticed;
* a review of whether any other query joins these two tables with a bare cast.

Filing rather than fixing inline because turning on dormant coalescing is a
behavioural change to a live cluster, not a documentation correction.

*(Filed by Claude Code during the phase 8 push audit of the queue-performance plan.)*

mikalstill:
Adds a closing push-audit phase to `PLAN-queue-performance.md`, runs it, and fixes what it found.

## The audit found that coalescing has never worked

Steps 4 and 5 of the plan — enqueue-side dedup and the worker-side fold, the coalescing half of a plan called "Queue performance **and coalescing**" — have been inert since PR #3194 merged on 2026-05-26.

Both primitives join `cluster_operations` to `object_states` on an `ON` clause that can never match, on **two independent axes**, either of which alone is fatal:

1. **Uuid form.** `cluster_operations.uuid` is `sa.Uuid` — undashed `CHAR(32)` on MariaDB. `object_states.object_uuid` is the dashed 36-character form. The repository already had `_dashed_uuid_expr()` for exactly this join, with a docstring saying so; neither function used it.
2. **Enum form.** `object_states.object_type` is `sa.Enum(ObjectType)` and stores the enum *name* `'NET_OP'`; `cluster_operations.operation_type` is a plain string holding the *value* `'net_op'`. The join compared those two columns directly. A *bound* value would be coerced by SQLAlchemy's Enum bind processor — which is why every other `object_states` query in `mariadb.py` safely passes a string — but a column-to-column comparison has nothing to do the coercing.

Corroborated in production before fixing: `coalesced sibling ops` had fired **zero times in seven days** on sfcbr. `docs/operator_guide/networking/overview.md` already named that absence as the diagnostic for this exact bug — the documentation was right and predictive, and nothing was watching it.

Fixes #3878.

## Why the test suite was green

Nothing anywhere executed the join:

- `tests/operations/test_baseoperation.py` — all ten coalescing tests mock `mariadb.claim_coalescible_siblings`, so they assert the dispatcher *decides to call* the primitive.
- `tests/test_mariadb_work_queue.py` — mocks `_get_engine` and feeds canned rows to `fetchall`. Its own docstring concedes it covers "the SQL-shape assertions".
- No functional coverage at all: `grep -rln coalesc shakenfist/deploy/shakenfist_ci/` returns nothing.

New `shakenfist/tests/test_mariadb_coalescing.py` runs the **production statements against a real database**. That needs sqlite shims for `JSON_LENGTH`, `JSON_UNQUOTE` and `UNIX_TIMESTAMP`, which sqlite lacks — registering shims rather than rewriting the query for sqlite is deliberate, since a query rewritten for the test proves nothing about the query that ships. Both halves of the fix are mutation-tested: reverting either fails three tests.

#3879 tracks the functional coverage, which is the half this PR does not close.

## The tests corrected the fix

An intermediate version also changed the `UPDATE` in `claim_coalescible_siblings`, on the theory it had the same enum defect and would report siblings folded without folding them. Reverting that change failed **nothing** — which prompted a check of the bind processor. The `UPDATE` binds a value, so it was always correct. There is no third defect. The enum is still bound there for consistency and the comment says so rather than claiming a bug that never existed. That correction is recorded in the phase plan too.

## What else the audit covered

Wave 1 clean (pre-commit, `tox`, proto stubs fresh, style greps). The other three wave-2 headings found nothing of substance and each names what it examined: three-layer pattern satisfied, no SQL-pushdown or cached-FK violations, no interpolated SQL (the variable-length `FIELD(queue_name, ...)` — the obvious injection site — is properly parameterised), `MAX_DEQUEUE_BATCH` clamps the batch at the gRPC trust boundary, and the wait event carries nothing sensitive. Five findings total, graded and dispositioned in the phase plan.

## Reviewer notes

- **This activates a code path that has never executed in production**, in which one worker marks another's operations complete. Until #3879 lands, the first real fold will also be the first test of the routing invariant at `daemons/network/workitem.py:60-77`. That is stated plainly in the plan rather than buried.
- **Review found a second defect the fix would have activated, now fixed in 33e29b831.** `network_ensure_mesh` was declared coalescible, but it does node-local work (`_apply_ensure_mesh` diffs *this* host's FDB) under a coalescing key that is the network alone — `cluster_operations` has no queue column and NetOp's model has no `node_uuid`, so the fold could not tell one hypervisor's mesh op from another's. The networknode survivor would have marked every other node's pending mesh op `complete` without doing their work. It is no longer coalescible, and the convention that made the queue-blind fold SQL safe is now enforced (`InvalidCoalescibleEnqueue`) rather than described in three comments. #3884 tracks the multi-column key that would let per-node tasks coalesce properly.
- Two new `# type: ignore[call-arg]` on `ObjectType(...)`: `ObjectType.__new__` takes `(string, proto_id)`, so mypy reads a by-value lookup as a short constructor call. Matches four existing suppressions (`node_health.py:204`, three in `mariadb.py`).
- The audit's baseline is the plan's own commit range, not `develop...HEAD`, because the work is already merged and that diff is empty. Decision 1 in the phase plan pins the ranges.
- PR #3873 also adds a push-audit step to this plan among twenty others; it is `CONFLICTING` and predates the Execution table. That conflict is being resolved from the other direction.

https://claude.ai/code/session_01G31Z9FWwYcbHYgt7YB3427


mikalstill:
## Review comments addressed

Thank you — action item 1 was a real defect and a serious one. Every item
is dealt with below; the code changes are in 33e29b831.

### 1 [FIX] Cross-queue fold — confirmed, fixed

Verified rather than accepted on description. `_direct_claim_coalescible_siblings`
has no queue predicate (and cannot have one — `cluster_operations` has no
queue column), and an AST walk of every `net_create_and_enqueue` call site
confirms the exact collision:

| Site | Tasks | Target |
|------|-------|--------|
| `network/network.py:316` | `[create_network_node, ensure_mesh]` | networknode |
| `external_api/instance.py:1091` | `[create_network_node, ensure_mesh]` | networknode |
| `network/network.py:980` | `[ensure_mesh]` | `<node_uuid>`, family `network` |
| `daemons/network/maintain.py:673` | `[ensure_mesh]` | `<node_uuid>`, family `network` |
| `daemons/network/maintain.py:728` | `[ensure_mesh]` | `<node_uuid>`, family `network` |

`network_ensure_mesh` is the *only* task ever enqueued to a non-default
target, and every one of those three sites is per-node. So the networknode
survivor's `task_names` includes `network_ensure_mesh`, and the three
per-node ops are single-task, same network, state `queued` — matched and
folded.

**I took neither of the two suggested options**, because both leave the real
problem in place. Joining `work_queue` (option a) would need
`JSON_UNQUOTE(JSON_EXTRACT(payload,'$.operation_uuid'))` — the op uuid lives
in an unindexed JSON blob — so it is a full scan of a hot table per fold.
Dropping the task (option b) is right but, on its own, is a convention that
the next per-node coalescible task quietly breaks.

The root cause is narrower than "cross-queue": **`network_ensure_mesh` is a
node-local task declared under a network-only coalescing key.**
`_apply_ensure_mesh` diffs *this* host's FDB, and there is no column
recording which node an op was targeted at — `target` is queue routing and
NetOp's model has no `node_uuid`. So "same network" is the finest grain the
fold can express, and it cannot be made finer without a schema change. The
declaration was wrong, not just the queue filter.

So: the task is no longer coalescible, and the invariant that makes the
queue-blind SQL safe is now **enforced** rather than described — a
coalescible task enqueued to a non-`networknode` target raises
`InvalidCoalescibleEnqueue`. `tests/schema/test_net_op_coalescing.py` checks
it at runtime and *also* walks every call site statically, so a future
per-node enqueue fails the suite whether or not a test executes that line.
Filed #3884 for the multi-column key and the `node_uuid` it needs.

The `workitem.py:63-66` invariant comment is corrected as you asked, and now
points at the enqueue guard as the thing that actually holds.

> the plan itself states "this change turns on a code path which has never run"

That sentence was written as a caveat and should have been read as a
question. The audit asked whether the join was wrong; it did not ask what
the corrected join would then match. That is recorded in the phase plan as a
finding about the audit, not just about the code.

### 2, 5, 6, 7, 8, 9, 10 [CONSIDER] — all taken

- **2** dedup path emits an audit event on the existing op and the network,
  mirroring `coalesced sibling ops`, and it is documented in the operator
  guide alongside it.
- **5** both `except ValueError` branches log, and so does the target-column
  whitelist rejection, which was equally silent.
- **6** extracted `_coalescible_preflight` and `_coalescible_states_join`.
  Both copies being wrong in the same two ways is the argument.
- **7** the fixture moved to `shakenfist/tests/dbfixture.py`;
  `test_mariadb_orphans.py` now uses it, which fixes the module-state leak it
  still had. Snapshot tightened with a `callable()` test so it no longer
  sweeps up the `_get_*_table` functions.
- **8** module docstring now states what the tests do *not* prove — sqlite
  emits nothing for `FOR UPDATE`, so every test here runs uncontended.
- **9** the `now` shim is gone.
- **10** docstring amended: the fold is conditional, and when skipped the
  outcome is a harmless duplicate run of an idempotent task.

### 3, 4 [DOC] — taken

New rule in `docs/developer_guide/coding_rules.md`: *"A column-to-column
comparison has no bind processor"*, covering both axes. It makes the point
you identified — a bound plain string works only because `ObjectType`
subclasses `str`, which is what makes every *correct* `object_states` query
in `mariadb.py` correct by accident and the broken join look like all of
them. CLAUDE.md's uuid pitfall now names the enum column in the same breath
and links to the rule.

Step 7's measurement caveat is in both the plan's status section and at the
head of "What step 7 measured", and re-measuring the fold's cost is folded
into #3879.

### 11 [CONSIDER] — you were right, and I fixed the arithmetic

Finding 2 was one blocking row with a half-done disposition, which the
definition of done does not admit. Split: the unit coverage discharges the
blocker (it is the coverage whose absence let the defect ship), and
functional coverage is now its own advisory row tracked as #3879. The DoD
now says what "resolved" means so the two sections agree.

### Test coverage

Added for every gap you listed except the two that are not expressible here:
different-queue fold (now unreachable by construction, so tested as the
guard firing), multi-task survivor folding single-task siblings,
`ObjectType` ValueError, `find_existing` on a different network, and
`find_existing` on a different task. `FOR UPDATE` serialisation and
functional coverage remain #3879, now stated in the docstring rather than
left implicit.

Suite is 3466 tests, 0 failures; `pre-commit run --all-files` and mypy
clean. Every new assertion is mutation-tested: re-adding `ensure_mesh` to
`COALESCIBLE_TASKS` fails 3, removing the enqueue guard fails 2, reverting
either half of the join fails 4, silencing either preflight warning fails 2.

*(Review addressed by Claude Code)*

Issue #3882, opened 2026-08-24
Scheduler capacity reconciler does not log the drift it corrects


mikalstill:
The scheduler capacity reconciler recomputes `scheduler_node_capacity`, `namespace_claims` and `cluster_capacity` wholesale from ground truth every five minutes. Since scheduler-reservations phase 3, admission and release also move those counters incrementally on every placement, which makes the reconciler a drift *corrector* rather than the sole write path.

It does not report the drift it corrects.

`reconcile_scheduler_capacity` logs membership and timing only:

```python
LOG.with_fields({
    "nodes": len(reply["nodes"]),
    "nodes_added": reply["nodes_added"],
    "nodes_removed": reply["nodes_removed"],
    "claims_expired": reply["claims_expired"],
    "duration": duration
}).info("Scheduler capacity reconcile pass complete")
```

(`shakenfist/daemons/cluster/scheduled_tasks.py:1057-1063`)

A pass that silently rewrites a `used_cpus` counter from 12 to 9 is indistinguishable in the log from one that changed nothing. Both emit the same line.

## Why this matters

"The reconciler reports zero drift across the burst" is an exit criterion for scheduler-reservations phase 3, and it was carried forward into the phase 4a soak. It cannot currently be checked from logs at all. During the phase 4a soak (2026-08-23/24, 299 reconcile passes on sfcbr) the most that could be established was that the reconciler ran healthily, saw a stable six-node cluster and expired nothing — not that the incremental counters agreed with ground truth. Confirming that needs the Prometheus gauges and an out-of-band comparison, which is a much higher bar than reading a log line, and is not available to anyone triaging after the fact.

Drift here is also the specific failure mode the guarded-UPDATE design exists to prevent, so silent correction is exactly the signal worth having. If admission and release ever start leaking, the reconciler is the component that knows first and currently says nothing.

## Suggested fix

Have the reconcile RPC return, and the pass log, the magnitude of what it changed — at minimum a count of rows whose counters differed, and ideally the summed absolute delta per dimension. Zero drift then logs as an explicit zero, which is what makes the phase 3 criterion checkable. A non-zero value is worth a WARNING and a Prometheus counter, since under a correct implementation it should be rare.

## Notes

Found while writing up the scheduler-reservations phase 4a soak; recorded as observation 4 in `docs/plans/PLAN-scheduler-reservations-phase-04a-demand-guard.md`. Not a regression introduced by phase 4a — the logging has been this way since the reconciler was added in phase 2.

PR #3888, opened 2026-08-24
Scheduler reservations: close out phases 3, 4 and 4a

### LINKED ISSUE #3813: Scheduler demand guard is arithmetically unsatisfiable on nodes with fewer than four schedulable threads

mikalstill:
Closes out the scheduler-reservations plan's phases 3, 4 and 4a. Phase 4a's
code merged separately in #3843; this is the soak readout, the close-out, and
one server fix the soak surfaced.

## The soak

Phase 4a's decision E6 made one sfcbr soak discharge three obligations: phase
3's step 9, phase 4's step 10, and 4a's own validation. Half of that worked as
planned and half did not.

**The demand guard half came from the window.** The pre-fix baseline is the
part worth keeping, because it was nearly lost — sfcbr was still running
pre-fix code when the window opened, which made it an unplanned control arm.
Over 48h, **3,904 of 3,904** capacity-guard refusals were demand-only. Not a
majority: all of them. Every refusal the guard produced in that window was
#3813. The P9 waiver was firing on 62% of placements, so the demand term was
being overridden more often than respected.

After the fix, refusals per placement fell roughly 18x and the waiver rate
fell to 4.1%.

Two results are recorded as weaker than the phase's Definition of done assumed,
rather than written up as wins:

* Capacity-proportional deviation *rose* slightly. The demand term is a
  spreader and spreading trades packing efficiency, so this is expected — but
  it makes the burst-distribution observation weaker evidence of spreading
  than the plan claimed it would be.
* Zero reconciler drift is **not provable from logs**. The reconciler
  recomputes the counters wholesale but logs only membership and timing, so a
  pass that silently corrected a counter looks identical to one that changed
  nothing. Filed as #3882.

**The claim half could not come from waiting.** The window produced zero claim
requests, zero over-limit audit events and zero expiries, because nothing on
sfcbr creates a namespace capacity claim on its own. That is an absence, not a
result, and more waiting would not have changed it. The plan should have said
so when it wrote the step.

So the pathway was exercised deliberately, with a new
`tools/exercise-namespace-claims.py`: 33 checks against sfcbr covering request
validation, create, duplicate and capacity refusals, read and cross-namespace
non-disclosure, field-masked update, drawdown against a real instance, the
`below_usage` shrink refusal, expiry and delete.

The expiry result is the one that could not be reasoned out from the code.
`coverage_state` is a stored column swept by the reconciler rather than
computed on read, so a claim stays active past its own `expires_at` until a
pass runs — observed at 210s, 280s and 321s against a five-minute interval.
Through that window the claim reads `state: created` while refusing a grow
with 409 `not_active`. That is D2's two-facts distinction visible from outside
the database, which is what the phase 4 soak was for.

## The capability string

Phase 4 routed the claim endpoints but never advertised them, so no client
could feature-detect claims at all. The phase 0 findings predicted exactly
this need and the phase plan then lost it: D7 moved *client verbs* out of
scope and the capability string went with them, even though advertising a
shipped endpoint is a server obligation. Nothing in phase 4's Definition of
done would have caught it and no test pins the list — it was found by a
passing question at close-out.

Fixed here as `auth: namespace-claims`, and recorded in phase 4's Future work
so the cause survives with the plan. This unblocks client-python#364.

## Plan changes

Phases 3, 4 and 4a set to Complete; index arithmetic to 6 of 13. Two phases
added at the end:

* **9. Client support for claims** — the `apiclient` verbs (client-python#364)
  and moving this repo's functional coverage off
  `apiclient.Client._request_url()` onto them. D7's reasoning was sound and is
  now spent. The phase depends on a client release this repo does not control;
  the plan says to leave it open rather than close the plan around it.
* **10. Push audit** — merged with the repo-wide push-audit phase that landed
  on develop meanwhile, keeping that convention's obligations plus two
  plan-specific adaptations: the audit range is the cumulative
  `ea3c9bf63..develop` rather than `develop...HEAD`, and findings are triaged
  rather than gating, because they cannot block a push that already happened.

## Testing

`pre-commit run --all-files` passes clean. The soak evidence is sfcbr itself.

Four bugs surfaced while writing the exerciser, all in it and none in the
claims code. Three were assumptions about asynchronous object lifecycle. The
fourth is worth flagging: the harness returned each check's detail string as
its result and callers used that as a pass/fail guard, so two checks with no
detail to report were silently skipped — never failed, never listed as
skipped, and green because the assertions did not execute. A passing check now
returns `True`.

Fixes #3813


mikalstill:
## Review comments addressed

All eleven items are addressed in four commits. Each finding was verified against the tree before acting on it; two turned out to need a different fix than suggested, and one uncovered a second instance the review had not flagged.

**1 — cleanup leak (FIX).** Confirmed by reproduction rather than by reading. A probe raising in the validation window against the pre-fix script died with `UnboundLocalError` at `state.get('network')` and leaked `claimsoak-1787645718` on sfcbr; the fixed script cleans up and reports HTTP 200. The leaked namespace has been removed.

**2 — nothing pinned the capability list (FIX).** Took the stronger option. `API_CAPABILITIES` is now a module-level dict rendered into the page, with `test_root.py` pinning it. The rendered output is unchanged — 11 families, 52 tokens, same order, verified against the previous literal before the swap. Deleting the `namespace-claims` token fails two of the four new tests. Worth recording that the "every declared token is served" test alone passes *vacuously* when a token is deleted, so the exhaustive snapshot is what actually carries the weight.

**3 — the impossible-claim check.** Made it a real assertion rather than renaming it. The check now runs before the claim exists, where capacity is the only thing that can refuse it, and asserts 507 specifically. Against sfcbr it reports `ok 507 on a claim larger than the cluster [HTTP 507]` — the first time the script has observed that path. The captured output of the recorded run confirms it logged `[HTTP 409]`, so the plan is corrected to say the 507 was not demonstrated.

**4 — reconciler drift.** Implemented rather than filed. As you noted, the deltas were already in the reply and only the log line dropped them, which made filing more expensive than fixing. The pass now warns per drifting node and carries `drifted_nodes` plus per-dimension `drift_*` magnitudes. Magnitudes are summed **absolute**: a cluster where one node over-counted by three cpus and another under-counted by two has drifted by five, not by one, and a signed sum would read as very nearly healthy. A quiet pass now says `drifted_nodes=0` rather than saying nothing, so zero drift and absent instrumentation stop being the same log line. Phase 3's status note now states outright that its zero-drift criterion was accepted on healthy-pass evidence.

  This turned up a second instance you did not flag: **phase 2's status note made the same claim** — "5-minute passes with no drift" — resting on the same absent instrumentation. Qualified too.

**8 — the arithmetic.** You were right that it does not close, and the fix was not the one suggested. Recomputed from the captured audit events: it was **fourteen**, not fifteen. 14/49 is 28.6%, so the 29% was right and the count was wrong, and 14 + 3 + 1 is exactly eighteen. The 62% → 4.1% conclusion is untouched.

**5, 6, 7, 9, 10, 11** — done as suggested. The capability comment no longer describes the conductor gating in the present tense; the by-hand cleanup hint names the admin namespace; `delete_namespace` treats 404 as success; the docstring carries no expiry figure at all so it cannot drift again; `--insecure` warns on stderr; and the script is now documented in `docs/operator_guide/scheduler.md` with the capacity-consumption warning.

`pre-commit run --all-files` passes clean.

One note for a human reviewer: this PR now carries three code changes with tests where it began as plans plus one capability string. The drift instrumentation in particular is a behaviour change in the cluster daemon that this review raised only as a documentation item.

Issue #3876, opened 2026-08-23
GetReferencesFrom/api is 11x GetReferencesTo: blob and artifact views re-read references


mikalstill:
`GetReferencesFrom` from the `api` daemon runs at roughly twice the
per-standing-instance coefficient the load model expects, and unlike its
sibling `GetReferencesTo` it is not explained by request volume. Split
out of database-load phase 6 (`docs/plans/PLAN-database-load-reduction-
phase-06-residual-load.md`), which measured it but did not close it.

## Measurements

From `database_requests_total`, 24h means on sfcbr:

| Day | QPS | Standing instances | Per instance | Counter coverage |
|-----|-----|-------------------|--------------|------------------|
| 2026-08-09 | 5.00 | 17.81 | 0.281 | 4 of 6 nodes |
| 2026-08-18 | 6.58 | 12.38 | 0.531 | 6 of 6 |
| 2026-08-22 | 13.04 | 21.93 | 0.595 | 6 of 6 |
| 2026-08-23 | 9.75 | 16.57 | 0.588 | 6 of 6 |

The ceiling the phase 5 ratchet defends is 0.32 per standing instance.
Note the coverage column: before #3708 the counter could not see daemons
co-located with MariaDB, so the 2026-08-09 figure is across four of six
nodes. Scaling it by the 1.5 the fixed-rate polls show gives 0.42, so
roughly half the rise is coverage and roughly half is real.

## What it is not

Not a duplicated read in the instance view. `Instance.external_view()`
issues exactly one `get_references_from` per view, inside the
`attribute_memo()` #3654 added.

## What it looks like

`GetReferencesFrom`/api is **11.6x** `GetReferencesTo`/api — 9.75/s
against 0.84/s on 2026-08-23. Every paired external view
(instance, node, blob, artifact) reads both, so a paired workload cannot
produce that ratio. Two unpaired call sites do:

**1. `Blob.external_view()` issues three `get_references_from` where one
would do** (`shakenfist/blob.py`):

```python
'depends_on': self.depends_on,   # get_references_from(BLOB, uuid, DEPENDS_ON)
'transcodes': self.transcoded,   # get_references_from(BLOB, uuid, TRANSCODE)
...
refs_from = mariadb.get_references_from(ObjectType.BLOB, self.uuid)  # unfiltered
```

The unfiltered read at the bottom already returns the DEPENDS_ON and
TRANSCODE rows the two properties go and fetch separately. Both
[...]

PR #3896, opened 2026-08-25
Let the issue fixer write its own PR description

mikalstill:
## Summary

Every PR the issue-fix workflow opened got the same body: a boilerplate
paragraph and a `git diff --stat`. Everything the model learned while
fixing the issue — what the root cause actually turned out to be, which
half of the issue it deliberately left alone, which judgement calls a
reviewer might make differently — was never asked for, and so never
reached the reviewer.

This teaches the fixer to write its own description. It is the
project's copy of shakenfist/development#53, which makes the same
change to `templates/issue-fix/issue-fix.yml`, the source this workflow
is derived from.

## What prompted it

#3894 arrived with a three-line boilerplate body against an issue
(#3876) carrying a detailed two-cause diagnosis. There turned out to be
two separate problems, only one of which was obvious:

**Systemic:** the body template. It never varied and never used the
model's prose. #3886, #3887 and #3891 all have good *titles* — so their
commit summaries were extracted fine — and identical contentless
bodies.

**Rare:** run 32825930889 also lost its commit summary. The last thing
the model emitted was

> The unit test suite (`tox -ecover`) is running in the background;
> I'll continue when it completes.

Under `claude -p` there is no "when it completes" — ending the turn
ends the session. The fix and its tests were staged and correct, but
`summary_found` was false, so #3894's commit landed under the
placeholder "Claude Code proposed a fix for this issue."

## Changes

All in `.github/workflows/issue-fix.yml`:

- The prompt asks for a `PR_DESCRIPTION_START`/`END` block alongside
  the existing commit summary, with guidance aimed at what a reviewer
  actually needs: root cause rather than reported symptom, what changed
  and why, what was deliberately *not* done, and any judgement call a
  reviewer might make differently. It is explicitly told not to emit a
  diffstat, a `Fixes #NNNN` line or a run link — the workflow appends
  those.
- A new "How this run works" section tells the model it gets exactly
  one turn: do not background a long command and end your turn meaning
  to check back, do not end a turn describing what you are about to do.
  This is the direct fix for the #3894 failure mode.
- The extract step pulls both blocks, computes its outputs into shell
  variables and writes each exactly once rather than defaulting them
  and relying on a later line overwriting the earlier one. The
  description extraction requires *both* markers, so a truncated block
  cannot run to the end of the captured output, and it strips only
  marker lines — a PR body may legitimately contain fenced code, which
  the commit-summary path's fence stripping would have eaten.
- The publish step assembles the body into a file and passes
  `--body-file`. This matters beyond tidiness: the old body was built
  with an *unquoted* heredoc, so dropping model text into it would have
  executed any `$(...)` or backticks it contained. The diffstat is now
  fenced, so it renders as a code block instead of a collapsed
  paragraph.

Both blocks are best-effort by design. A missing description falls back
to the commit message body, and a missing commit message to the old
boilerplate — a correct fix is worth publishing even when the prose is
lost.

## Verification

- `pre-commit run --all-files` passes; the actionlint hook covers this
  file.
- I extracted the rewritten extract and publish steps and ran them
  against synthetic model output for four cases: both blocks present,
  commit summary only, neither (#3894's exact failure mode), and a
  description with a `START` marker but no `END`. All four produce a
  sensible body and the right `summary_found`/`description_found` pair.
- A `$(touch /tmp/PWNED)` planted in the synthetic description does not
  execute.

## Notes for review

- The appended heading is `## Automated verification` rather than
  `## Verification`, so a model that writes its own verification
  section does not collide with it.
- Nothing here changes what the fixer *does*, only what it reports, so
  the blast radius is confined to automated PRs. The next issue-fix run
  is the real test.
- #3894 itself has had its description rewritten by hand; it is not
  affected by this change.

https://claude.ai/code/session_01GkQoWx4NdGWs94DPx3gsb1

### DIFF
--- a/.github/workflows/issue-fix.yml
+++ b/.github/workflows/issue-fix.yml
@@ -51,6 +51,12 @@
 #   test suite and ended its turn waiting for it, which in -p mode
 #   simply ends the session: the code changes survived, both blocks
 #   did not.
+# - The description is neutralised by tools/neutralise-pr-body.sh
+#   before publication: an @mention notifies a real person when the
+#   draft is created and an issue-closing keyword closes an unrelated
+#   issue on merge. The prompt forbids both, but a side effect which
+#   fires automatically and cannot be undone should not rest on the
+#   model having complied.
 # - Both blocks are pulled out of the captured output by
 #   tools/extract-model-block.sh rather than by sed inline here. That
 #   parsing has more edge cases than it looks (a model which emits a
@@ -723,6 +729,7 @@ jobs:
           # wrapper above: Claude has just edited this checkout and may
           # have touched tools/.
           cp tools/extract-model-block.sh ${{ runner.temp }}/
+          cp tools/neutralise-pr-body.sh ${{ runner.temp }}/
 
           summary_found=false
           description_found=false
@@ -739,6 +746,15 @@ jobs:
               ${{ runner.temp }}/pr-description.txt
           then
             description_found=true
+
+            # The prompt forbids @mentions and issue-closing keywords,
+            # but the description is published verbatim and both act
+            # the instant "gh pr create" runs -- before any human has
+            # looked at the draft, and a notification cannot be taken
+            # back. Prompt compliance is a weaker guarantee than code
+            # for a side effect which is not reversible.
+            ${{ runner.temp }}/neutralise-pr-body.sh \
+              ${{ runner.temp }}/pr-description.txt
           fi
 
           # Neither block is fatal: the publish step falls back to a
@@ -944,6 +960,7 @@ jobs:
             ${{ runner.temp }}/commit-message.txt
             ${{ runner.temp }}/pr-description.txt
             ${{ runner.temp }}/pr-body.md
+            ${{ runner.temp }}/summary-body.txt
             ${{ runner.temp }}/claude-logs/
           retention-days: 14
           if-no-files-found: ignore
--- a/.github/workflows/issue-fix.yml
+++ b/.github/workflows/issue-fix.yml
@@ -51,6 +51,20 @@
 #   test suite and ended its turn waiting for it, which in -p mode
 #   simply ends the session: the code changes survived, both blocks
 #   did not.
+# - Both blocks are pulled out of the captured output by
+#   tools/extract-model-block.sh rather than by sed inline here. That
+#   parsing has more edge cases than it looks (a model which emits a
+#   block twice, a block with no closing marker, a block the model
+#   wrapped in a fence anyway) and each is covered by
+#   shakenfist/tests/test_extract_model_block.py. Extend the tests, not
+#   this step, when a new shape of output turns up.
+# - The model is asked to run tox -ecover and the workflow then runs it
+#   again in "Verify fix". The duplication is deliberate: the model's
+#   run is what lets it fix what it broke while it still has a turn,
+#   and the workflow's run is the gate on publication, which cannot
+#   trust the model's report of its own testing. Paying for the suite
+#   twice is cheaper than publishing a broken fix or burning a whole
+#   run on one.
 # - The description is model output, so it is passed to gh with
 #   --body-file. Never interpolate it into a shell string.
 # - The issue-fix-<N> branches are owned by this workflow and are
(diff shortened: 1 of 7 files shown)


mikalstill:
Thanks -- all of the actionable items are addressed, mostly by one
structural change rather than seven local patches.

The marker parsing moved out of the workflow into
`tools/extract-model-block.sh`, covered by
`shakenfist/tests/test_extract_model_block.py` (21 cases, run by
`tox -ecover`). That was the review's own test-coverage suggestion, and
it also brings the step in line with the repo convention against
non-trivial shell inline in a workflow step. Both blocks now go through
it, which closes items 1, 2, 4 and 5 together:

- **Item 1** (repeated markers): the block is the text between the last
  start marker and the first end marker which follows it, buffered and
  emitted only once that end marker is seen.
- **Item 2** (unterminated `COMMIT_SUMMARY`): both paths now require a
  closing marker. Verified against your exact synthetic input -- the
  summary is rejected and the description still comes out intact.
- **Item 4** (model-fenced block): a fence is stripped only when one
  wraps the whole block. Fenced code inside a description is preserved,
  which is a separate test.
- **Item 5** (self-truncation): a marker terminates only when it is the
  entire line, so a description may name the tokens even at the start of
  a sentence. The prompt says so too.
- **Item 3**: `pr-description.txt` and `pr-body.md` are uploaded.
- **Item 6**: the commit-body tier is written to a file and tested for
  content, so a bare subject line falls through to the generic tier.
- **Item 8**: `docs/developer_guide/ci.md` now describes the two-block
  contract, the fallback tiers and the extraction rules.

**Item 7 I have not taken.** The duplicate `tox -ecover` is deliberate:
the model's run is what lets it fix what it broke while it still has a
turn, and the workflow's run is the gate on publication, which cannot
trust the model's report of its own testing. Narrowing the model's run
trades a few minutes for whole runs lost to a failure the model could
have fixed. The reasoning is now recorded in the workflow header so it
does not get re-litigated.

Two things worth flagging that came from outside this review:

1. The template this workflow is derived from
   (shakenfist/development#53) had its own review of the same finding
   and landed a different fix. Its rule was stronger in one place --
   a *repeated* start marker with no end between restarts the block,
   where taking the first complete block leaves a literal marker line
   in the published prose. I adopted it, and pushed the reverse half of
   the merge upstream so the two do not drift.
2. From that same upstream review: the description is published into a
   PR body verbatim, so the prompt now forbids issue-closing keywords
   for any issue (not just the appended one) and forbids @mentions,
   which notify real people.

Each guard has a mutation test behind it -- breaking the property in
the script makes a specific named test fail, which I ran for all eight.
`pre-commit run --all-files` passes, including actionlint (item 10).

mikalstill:
Round 2 addressed. Both defects were in code round 1 added, and both
were real.

**Item 1 (fix)** -- correct, and your diagnosis of why the existing
test did not catch it was the useful half: `test_interior_fenced_code_is_preserved`
passes only because its fixture starts with a heading. The condition
now also requires the first and last lines to be the *only* fences in
the block, so two unrelated fenced blocks keep their markers. Your
reproduction is a test, as is the odd-fence-count case, and removing
the new clause makes it fail.

**Item 4 (consider, taken)** -- you are right that the two copies are
not the same situation, and my comment asserting they were is what
made it look considered. `claude-model-fallback.sh` is copied before
the model runs because bash reads a script lazily; these are staged
after it exits, where the hazard is the opposite one. Both now come
from `git show HEAD:`, and the comment says which hazard it is
guarding.

**Item 2 (consider, taken)** -- the reorder, plus the other half: the
neutraliser now closes a fence the model left open, so the diffstat
and footer after the description still render rather than being
swallowed up to the workflow's own fence.

**Item 3 (consider)** -- already landed in `18a0537`, pushed just
after this review started, from the template's own review of the same
change. Two differences from your suggested sed worth noting. It drops
the `@` rather than inserting an HTML comment, which is what the
prompt asks the model to do itself, so the rendered text reads
normally. And it is fence-aware: a description quoting `@property` or
an email address is a normal description, and mangling one to defuse a
hazard that is not there is its own defect, so that half has as many
tests as the rewriting half.

New in this round: `shakenfist/tests/test_issue_fix_workflow_seams.py`
pins the three seams between the workflow and its two scripts, since a
tested script the workflow does not call is neither an extraction nor
a defence. Worth flagging that its ordering assertion initially passed
for the wrong reason -- its window reached back far enough to find the
commit message step's own `Fixes` line -- which mutation testing
caught. It now scopes to the body's brace group.

`pre-commit run --all-files` passes. Every guard added across both
rounds has a mutation behind it: breaking the property in the script
or the workflow makes a specific named test fail.

`tools/neutralise-pr-body.sh` has not been reviewed yet -- it landed
between the two rounds -- so it is the part of this worth a look.

Issue #3516, opened 2026-07-25
sf-sidechannel orphans agent operations in 'executing' when the executor fails to complete them


mikalstill:
## Summary

`sf-sidechannel` can leave an agent operation **orphaned in the `executing`
state forever** when the executor fails to complete it. The client only finds
out via its own timeout. Seen in CI as
`guest_ci_tests.test_agentops.TestAgentOperations.test_instance_put_and_get_blob`
timing out after 378s with the agentop stuck in `executing` (a `get-file` that
should have materialised a blob).

Root-caused from the merge-queue run for PR #3506
(`gh-readonly-queue/develop/pr-3506-...`, run 30175769629, Guests suite).

## Evidence

Instance `3c5c0987-5a01-410b-9f05-56417e01a4ab` on sf5, agentop
`c9934e6d-512a-409f-b1d6-f19a19778a65`, command `get-file /tmp/foo`:

- The agent welcomed and `execute`/`put-blob` commands **succeeded** on this
  instance ("got result for agent execute command"), so the agent, vsock
  transport and CID are healthy. (The large CID `2884833236` is expected: SF
  assigns random 32-bit CIDs; it is consistent across the libvirt cmdline,
  domain XML and the guest.)
- The `get-file` op was dispatched to `executing` at 21:51:29 and **never
  produced a result**; the client timed out at 378s.
- In the **same run**, sibling get-file tests passed: `test_get ... ok`,
  `test_get_missing_file ... ok`. Only the variant that returns a **non-empty
  file materialised as a blob** failed (the `_handle_file_chunk`
  `chunk.offset != 0` branch: `Blob.new` + `register`).
- This is a historically known area: #2240 "Agent get-file blobs are sometimes
  empty" (being reopened alongside this).

## Why a transient hiccup becomes a permanent hang

`Instance.agent_operation_next()` pops the queue entry once the op has left
`QUEUED` (i.e. as soon as the executor moves it to `EXECUTING`). So an
`EXECUTING` op is **never re-dispatched**. If the executor then exits or stalls
without moving the op to `COMPLETE`/`ERROR`, the op is orphaned in `EXECUTING`
with nothing to retry it or fail it. Concretely, in
`daemons/sidechannel/main.py`:

1. `SideChannelJob.execute()` catches `NoSuchChannel` / `socket.error` /
   `OSError` and simply returns; if a command was already in flight (op
   `EXECUTING`), the op is left orphaned.
[...]

PR #3898, opened 2026-08-25
Agent operation deadlines phase 4: enforcement

mikalstill:
Phase 4 of `PLAN-agent-operation-deadlines.md`. Phases 2 and 3 added
the columns and let a client set them; this is the phase that makes
them bite.

The plan for this phase is
[`docs/plans/PLAN-agent-operation-deadlines-phase-04-enforcement.md`](docs/plans/PLAN-agent-operation-deadlines-phase-04-enforcement.md),
committed first so the reasoning is reviewable alongside the code.

## What this does

Adds `expired`, a terminal state meaning a timing budget the caller
set was exhausted, as distinct from `error`, which means the operation
itself failed. It is reachable from all four non-terminal states, leads
only to `deleted`, and joins `FINAL_OBJECT_STATES` so the hard delete
sweep reaps it rather than leaking an `object_states` row.

Three enforcement points:

- **At dequeue.** A queued head past its deadline is expired and
  retired inside `agent_operation_next()`, so work the caller has
  already abandoned never occupies the instance's single executor slot.
- **During preflight.** Checked on entry and after each
  `Blob.ensure_local()`, which is the longest pre-queue delay in the
  system and precisely the wait a receipt-anchored deadline exists to
  count.
- **In the executor.** `AGENT_OPERATION_EXECUTION_TIMEOUT`, the fixed
  900 second backstop, is deleted. In its place are the caller's
  wall-clock deadline and a progress timeout that applies only while a
  command which can actually report progress is in flight.

Progress has to be observed explicitly: `self.last_data` is refreshed by
any socket traffic and the executor pings every two seconds, so it never
ages past two seconds whatever the agent is doing — which is exactly why
a stalled `get-file` could sit in `executing` for the full 900 seconds
(#3516). `observe_progress()` is called from the four reply handlers
instead, and persists `last_progress` at most every ten seconds so a
chunk stream does not become one attributes write per chunk.

## Decisions worth review

- **An explicitly unbounded operation stays unbounded.** Deleting the
  900 second constant means `deadline_seconds=0` plus
  `progress_timeout_seconds=0` is now genuinely unbounded. Phase 3 has
  already published those sentinels as meaning exactly that, so capping
  either here would make the published API specification false.
  Everything that does not opt out is bounded: 600 seconds by default,
  a 30 second progress window for progress-capable commands, and a
  transition-anchored fallback for rows written before phase 3.
- **A progress stall and a passed deadline are the same state**,
  distinguished only by the message. Both are numbers the caller chose,
  and it leaves the phase 6 client with one terminal outcome to learn
  rather than two.
- **A NULL deadline anchors on the current state's transition time**,
  not on `time.time()`. Such a row carries no request receipt time, and
  a deadline anchored on now recedes at every check and can never fire.

## Two pre-existing defects found while implementing

Neither is fixed here beyond what this phase already touches:

- `AgentOperation` never overrides `_db_set_attribute()`, so every
  `agentop.error = ...` write in the tree reaches nothing but a warning
  log. `fail()` therefore records the reason on the state, where it
  survives, and keeps the `self.error` write so those call sites become
  correct if the gap is closed. Recorded in the plan's Future work; it
  wants its own issue.
- `NodeAgentopOp._preflight()`'s missing-blob path assigned `aop.error`
  from the `preflight` state, which the setter refuses; the resulting
  exception was swallowed by `dispatch_task()` and the message naming
  the blob was discarded. Fixed, since it is a sixth instance of the
  same family of writes this phase guards.

## Out of scope

Retry and the node-local reaper (phase 5), `client-python` (phase 6),
and the operator guide, user guide, release note and functional CI
coverage (phase 7). `docs/developer_guide/state_machine.md` is updated
here rather than in phase 7 because it is a rendering of
`state_targets`, and leaving it would have the tree contradicting
itself for three phases.

## Testing

All seven of the plan's definition-of-done checks pass,
`pre-commit run --all-files` is green, and the full unit suite is 3414
passed / 0 failed. Five assertions were mutation tested — the NULL
anchor, the dequeue check, the post-copy preflight check, the
`reports_progress` gate, and the field mask — and every mutation was
caught by a failing test.

The phase plan records five departures from what it briefed, including
two guards extracted into methods so they could be unit tested at all.

https://claude.ai/code/session_01Gravz2NuQqWcrq13nCrVVW


mikalstill:
## Review response

All fourteen items are addressed. The work landed in `3b19ab8a7`, `2d625d561`, `83a0d2d0e` and `536e815ad`, and the reasoning is recorded in the phase plan under *Response to the automated review*. Item by item, against the current head:

| # | Item | Where |
|---|---|---|
| 1 FIX | Progress window seeded before dispatch | The dispatch block is extracted to `_dispatch_next_command()` (`shakenfist/daemons/sidechannel/main.py:859`) so it is testable at all — previously it was reachable only through a live vsock connection. The clock is now seeded at `:911`, immediately after `_send_commands_single_envelope()` returns at `:904`, in the same breath as `self.ready = False`. `ExecutorDispatchWindowTestCase` asserts both halves: a 300 second dispatch does not expire a 30 second window, and a genuinely silent agent still does. |
| 2 FIX | Published docs still say deadlines are unenforced | Both pages corrected. The definition-of-done grep was widened from `shakenfist/config.py` to `docs/` outside `docs/plans/`, which is what let the two statements survive step 4f. |
| 3 FIX | Uncached `GetState` per executor loop iteration | Rate limited to `BUDGET_CHECK_INTERVAL` (1 second) at `main.py:764-768`. One second is ample resolution for a 30 second window and bounds the cost whichever branch `effective_deadline()` takes, which is the suggestion's third option — memoising the anchor would have left the dequeue-side read in place. |
| 11 DOC | Stale in-code comments about phase 4 | `main.py:318` now names `SideChannelExecutorJob.expire_if_out_of_budget()` as the reader, and `AgentOperation.new()`'s docstring (`shakenfist/operations/agentoperation.py:160-165`) names all three enforcement sites. `retryable = True  # read in phase 5` left alone. |
| 4 | Expiry reason recorded only on the operation | `expire()` uses `add_event_multi` (`agentoperation.py:304`), restoring the plan's own decision 2, which the implementation had quietly narrowed. |
| 5 | Preflight head with a passed deadline wedges the queue | Real, and deliberately not fixed here — see below. Documented at the site (`shakenfist/instance.py:2525-2545`) rather than left implied. |
| 6 | Effective default tightens 900s → 600s with no release note | `docs/release_notes/v07-v08.md:722-753`, covering the deleted backstop, the new anchor at request receipt, the `expired` state, and `AGENT_OPERATION_DEFAULT_DEADLINE` as the knob. Phase 7 can expand it rather than introducing it after the fact. |
| 7 | `fail()`'s `self.error` write logs a spurious warning | The second write is dropped (`agentoperation.py:310-331`). Worth noting what happened next: issue #3899 — filed from this review — was fixed on `develop` as `c8ff408f0` while this branch sat in the merge queue, so the `error` setter now stores the message on the state row that `fail()` has just written. `agentop.error` reads the message back, and the drop is correct for a better reason than the one it was made for. |
| 8 | `execute()`'s `finally` block bypasses `fail()` | `main.py:512-518` now calls `fail()`, so the orphaned-executor message reaches `object_states` like every other reason. |
| 9 | `observe_progress()` layering and DB blast radius | The read-modify-write moved into `AgentOperation.record_progress()` (`agentoperation.py:385`) alongside `add_result()`. The call is wrapped (`main.py:840-846`) and the stamp moves only on success, so a failed write is retried next call rather than suppressed for an interval. |
| 10 | Sidechannel fake reimplements the guard it tests | `test_fail_from_expired_leaves_the_state_alone` deleted; the guard stays on `_FakeAgentOp` as fixture behaviour, with a pointer to where the real guard is asserted. |
| 13 | `hard_delete()` does not clear object references | `agentoperation.py:411`. Agreed it is theoretical today; the asymmetry was the trap. |
| 14 | `constants.py` comment calls the missing `error` state deliberate | Reworded to match the plan's own assessment — a known defect, not an intent (`constants.py:194-201`). |
| 12 INFO | `initial --> expired` declared but unreachable | Noted in the `state_targets` comment as permitted for completeness (`agentoperation.py:50-55`). |

### Item 5 is deferred, not done

A `PREFLIGHT` head whose deadline has passed still wedges its instance's queue if the preflight task never runs. Expiring it here races a preflight task which may still be working, and the safe form needs a grace period generous enough that it never does — which is the node-local reaper in phase 5. Fixing it here would mean picking that grace period without the reaper's context. The gap is now stated at the `break` in `agent_operation_next()`, naming phase 5 as the owner, so the docstring no longer asserts an enforcement point without noting that it only fires if the task runs.

### On item 1

Worth recording that this was the only real defect in the review, and it is the same class of mistake the phase deliberately avoided for `last_data` — measuring the wrong interval — in the other direction. The extraction of `_dispatch_next_command()` was needed to test it, which is itself the useful outcome: that path had no test coverage at all before.

mikalstill:

mikalstill:
Thanks — twelve items this round, eleven of which produced a change. The
twelfth (item 11) is a decision the plan already argues, and is left as
it stands.

| # | Kind | Disposition |
|---|------|-------------|
| 1 | FIX | `SideChannelExecutorJob._handle_command_error()` override — `main.py:520` |
| 2 | FIX | `_await_command()` bounded and fail-fast — `shakenfist_ci/base.py:760-819` |
| 8 | DOC | Filed as #3922; cited from `constants.py:198` |
| 3 | CONSIDER | `test_leaves_expired_op_untouched` — `test_daemon_sidechannel_executor.py:116` |
| 4 | CONSIDER | `_abandon_get_file_transfer()` — `main.py:839` |
| 5 | CONSIDER | `effective_deadline(state=...)` — `agentoperation.py:235`, caller at `instance.py:2511` |
| 6 | CONSIDER | Terminal guard before the executing write — `main.py:991` |
| 7 | CONSIDER | Docstring, not a widened tuple — `main.py:910` |
| 9 | CONSIDER | `{window:g}` — `main.py:835` |
| 10 | CONSIDER | Maintenance obligation named in both fake docstrings |
| 11 | INFO | No change (decision 6) |
| 12 | INFO | Recorded against phase 5's reaper in the master plan |

**On item 1.** Confirmed, and it is the one real defect. The base handler
only emits an event; that is correct for `SideChannelMonitorJob`, which
has no operation to fail, and it was survivable for the executor only
because the 900 second backstop eventually tore the connection down and
`execute()`'s finally block recorded an error. Deleting the backstop
removed the thing that made it survivable. The executor now overrides it:
same event, then `fail()` with the agent's own error text, drop the
remaining commands, and set `ready` so the socket loop takes its ordinary
goodbye-and-disconnect exit. That exit only marks an operation complete
when it is still executing, so it cannot overwrite the error. The
`GetException` path a few lines below had the identical
spin-until-a-budget shape and got the same two lines.

**On item 2.** Agreed that the phase 7 deferral stops being defensible
once this phase changes the odds, so the fail-fast half was pulled
forward for that one loop: an absolute 900 second bound, a check for
`error`/`expired`/`deleted`, and a failure that dumps the operation plus
the instance's last fifty events. The events are the part that matters —
the expiry *reason* lives on the state row, which `external_view()` does
not publish, but `expire()` audits it against the instance too, so
"deadline passed while queued" and "no progress from the agent" are
distinguishable from the test output. The event-renewed loops in the same
file are untouched and remain phase 7's.

**On item 7, where I did not take the suggestion.** `deleted` stays out
of the tuple, with the reasoning now in the docstring rather than left
implicit. Abandoning a command sequence part way through leaves the guest
in a state no caller asked for; a delete says the record is unwanted, not
that a half-applied change should be frozen in place. It is also a
behaviour change with no end-to-end test available in this phase.
Interrupting live work belongs with phase 5's reaper.

**Item 6 went slightly further than suggested.** The guard also clears
the command list before returning. `_abort_commands_if_terminal()` in the
`finally` only fires for `error` and `expired`, and `self.ready` is true
for the dispatch to have been reached at all — so a `complete` or
`deleted` operation with commands still queued would be re-dispatched on
every pass and the loop would never reach its exit.

Split across two commits: `9be799f5c` for the two FIX items, which are
the ones CI can actually exercise, and `04eaeb001` for the rest.

Verification: full unit suite green (3,656 tests), `pre-commit
run --all-files` clean. The three new behaviours were mutation tested —
removing the `fail()` call, the partial-file teardown, or the dispatch
guard each fails a specific new test rather than passing silently.

mikalstill:

PR #3921, opened 2026-08-26
Let guests hand freed memory back to the host

### LINKED ISSUE #3920: Guests cannot return freed memory: libvirt template never enables virtio free page reporting
### Summary

Shaken Fist's libvirt domain template configures the virtio balloon for
statistics only:

```xml
<memballoon model='virtio'>
  <stats period='10'/>
  ...
</memballoon>
```

It never sets `freePageReporting='on'`, so `VIRTIO_BALLOON_F_REPORTING` is
not offered to the guest. The consequence is that a guest's host memory
footprint is a **high-water mark**: any page the guest touches once stays
charged to the hypervisor for the life of the instance, even after the guest
has freed it. The host cannot tell those pages are garbage, so the only way
it can reclaim them under pressure is to swap them out.

### Evidence

Found on a production hypervisor (`sf-6`, 62.4GiB, 24 threads) on
2026-08-27. A long-lived static CI runner instance (8GB, 4 vCPU, up 4d21h)
had **4.9GB of its pages sitting in host swap**:

```
VmSize: 13630052 kB    VmRSS: 3396252 kB    VmSwap: 5155624 kB
```

Inside that same guest, at the same moment:

```
               total   used   free  buff/cache  available
Mem:            7940    735   4721        2774       7205
load average: 0.17, 0.04, 0.01      # zero CI jobs in the preceding 24h
```

The guest is idle and considers 4.7GB of its own RAM free. The host has 4.9GB
of that instance paged out. Those two numbers are the same memory: pages the
guest finished with days ago, which it has no way to hand back.

Confirming the feature is simply not offered, from inside the guest:

```
/sys/bus/virtio/devices/virtio0 driver=virtio_balloon
features=0100000000000000000000000000110010000000100000000000000000000000
         ^bit1 STATS_VQ = 1              ^bit5 REPORTING = 0
```

Bit 1 is set because the template asks for `<stats period='10'/>`. Bit 5 is
clear because the template omits `freePageReporting`. The guest kernel is
ready and waiting:

```
/sys/module/page_reporting/parameters/page_reporting_order -> 11
```

The swap did not accumulate gradually; it ratcheted, one CI burst at a time,
and never came back (hourly samples):

```
08-23 22:00  1.12 GB   +920 MB
08-24 03:00  2.70 GB  +1624 MB
08-24 07:00  4.86 GB  +2303 MB
08-26 03:00  4.93 GB   +275 MB    ... flat ever since
```

This matters because the node was simultaneously **109% committed** (68GB of
allocated guest RAM on 62.4GiB of hardware) and still being given new work.
The hypervisor was paging out an idle guest's dead pages to make room for
live ones, which is the correct thing for it to do given the information it
has, but it should not have needed to: nearly all of that memory was free
inside the guest and could simply have been returned.

### Proposed fix

Set the attribute in
`shakenfist/deploy/collection/roles/hypervisor/files/libvirt.tmpl`:

```xml
<memballoon model='virtio' freePageReporting='on'>
```

The guest then reports freed pages and the host releases them
(`MADV_DONTNEED`) instead of swapping them.

Notes on safety and scope:

- This is **not** ballooning. The balloon target is untouched, so nothing
  shrinks a guest against its will.
- The scheduler's accounting is unaffected. `memory_total_instance_actual`
  comes from `domain.info()[2]`, which is the balloon target and equals
  `maxmem` when nothing is ballooning. So enabling this does not cause the
  scheduler to pack more onto a node just because a guest handed pages back.
- Only *free* pages are returned. Guest page cache (2774MB in the example
  above) is a real working set and stays resident.
- Requirements: QEMU >= 5.1 and libvirt >= 6.9 on the host, guest kernel
  >= 5.7. The feature is negotiated, so older guests decline it and behave
  exactly as they do today. (The affected cluster runs libvirt 11.3.0 /
  QEMU 10.0.11 with Debian 12 guests on 6.1.)
- Guests only report free runs of at least `page_reporting_order` pages
  (8MB with the value above), so this reclaims bulk free memory rather than
  fragments.

I have verified the resulting domain XML against the live libvirt schema
with `virt-xml-validate`.

### Related

Adjacent to but distinct from #3636 (scheduler RAM gates race placement, and
ranking ignores RAM). #3636 is about *where* work lands; this is about how
much RAM an instance is charged for once it lands. The sf-6 case above is a
node that lost on both counts at once.

mikalstill:
Enables virtio free page reporting on the balloon device so guests can
return memory they have freed, instead of leaving it charged to the
hypervisor until something swaps it out.

## Why

The domain template configured the balloon for statistics only, so
`VIRTIO_BALLOON_F_REPORTING` was never offered. An instance's host footprint
was therefore a high-water mark, and the host's only route to reclaiming a
guest's dead pages was swap.

On `sf-6` (62.4GiB) an idle static CI runner was found holding **4.9GB of its
pages in host swap** while reporting **4.7GB free inside the guest** — the
same memory, freed days earlier, with no way to give it back. The node was
109% committed at the time and still being placed on, so it was swapping out
dead pages to make room for live ones.

The feature bits confirm the guest was ready and the host simply never asked:

```
features=0100...  bit1 STATS_VQ = 1   (template sets <stats period='10'/>)
                  bit5 REPORTING = 0  (template omitted freePageReporting)
```

Full analysis, including the hourly ratchet of the swap growth, is in #3920.

## Scope and safety

- **Not ballooning.** The balloon target is untouched; nothing shrinks a
  guest against its will.
- **Scheduler accounting is unchanged.** `memory_total_instance_actual` reads
  `domain.info()[2]`, the balloon target, which still equals `maxmem`. So
  this does not cause nodes to be packed harder because a guest behaved well.
- **Only free pages move.** Guest page cache stays resident.
- **Negotiated.** Guest kernels older than 5.7 decline it and behave exactly
  as they do today. Hosts need QEMU >= 5.1 / libvirt >= 6.9.

## Testing

The rendered domain XML was validated with `virt-xml-validate` against a live
libvirt 11.3.0 / QEMU 10.0.11 hypervisor, using a dump of a real running
instance with the new memballoon block substituted in. `pre-commit` passes.

That validation earned its keep: a first draft of the explanatory comment
contained a `--`, which is illegal inside an XML comment and would have
broken every instance start. Worth running on any future `libvirt.tmpl` edit.

Beyond that, this needs a soak more than it needs a unit test — the behaviour
it changes is only visible in host RSS over a long-lived instance's lifetime.
I would suggest watching `VmSwap`/`VmRSS` for the static CI runners on a
hypervisor for a few days after it lands.

Fixes #3920

https://claude.ai/code/session_01HamA8edRXMvmcEakW4ufSa

### DIFF
--- a/docs/operator_guide/scheduler.md
+++ b/docs/operator_guide/scheduler.md
@@ -30,7 +30,9 @@ a placement decision can be reconstructed after the fact (see
    See [CPU overcommit](#cpu-overcommit).
 4. **RAM admission** -- the node must retain its published memory
    reservation after placement, and KSM overcommit must stay
-   under `RAM_OVERCOMMIT_RATIO`.
+   under `RAM_OVERCOMMIT_RATIO`. See [Guest memory returned to the
+   host](#guest-memory-returned-to-the-host) for how much of a guest's
+   memory a node is actually charged.
 5. **Disk capacity** -- requested disk must fit while leaving the
    node's `NODE_DISK_RESERVATION_GB` free on the instances/blobs
    filesystems. The candidate node publishes its own reservation as
@@ -117,6 +119,71 @@ The published fields are `cpu_cores`, `cpu_threads`,
 `disk_reservation_gb`. On Intel hybrid CPUs the daemon also
 publishes `cpu_cores_performance` and `cpu_cores_efficiency`; these
 are informational and nothing in scheduling consumes them yet.
+
+## Guest memory returned to the host
+
+Instances are given a virtio balloon device with **free page
+reporting** enabled. When a guest frees a page, its kernel tells the
+balloon driver, and the host releases the backing memory
+(`MADV_DONTNEED`) instead of holding it. Without this a guest's host
+footprint is a high-water mark: memory a workload touched once stays
+charged to the hypervisor for the life of the instance, and the only
+way the host can reclaim it under pressure is to swap it out. On one
+production node an idle CI runner was found holding 4.9 GB of its
+pages in host swap while reporting 4.7 GB free inside the guest, on a
+node that was 109% committed and still being placed on.
+
+Things worth knowing about it:
+
+- **This is not ballooning.** The balloon target is never moved, so
+  nothing shrinks a guest against its will. Only pages the guest has
+  already decided it does not want are returned.
+- **Guest page cache is not returned**, only free pages. A guest that
+  has read a lot of data keeps that cache, and keeps the host memory
+  backing it.
+- **It is negotiated.** A guest kernel older than 5.7 does not
+  advertise `VIRTIO_BALLOON_F_REPORTING` and simply declines, behaving
+  as it always has. The host needs QEMU 5.1 or newer and libvirt 6.9
+  or newer, which every supported platform satisfies.
+- **The guest can take the memory back at any time.** A returned page
+  is re-faulted on next use. Freed memory is an opportunity, not a
+  permanent reduction in the instance's demand.
+
+### What this changes about scheduling
+
+The two RAM admission checks are affected differently, and the
+difference matters when you are debugging a memory pressure incident:
+
+- The **measured** check -- the node must retain its published memory
+  reservation after placement -- reads the `memory_available` metric,
+  which the resources daemon takes from the host's own view of free
+  memory. Returned pages show up there, so a node running
+  reporting-capable guests will admit work it previously would not.
+  That is the point of the feature, but remember the previous bullet:
+  those guests can re-fault the pages afterwards.
+- The **allocation** checks are unchanged. KSM overcommit is computed
+  from `memory_total_instance_actual`, which is the balloon
+  allocation, and the [capacity
+  counters](#admission-is-a-guarded-capacity-claim) count requested
+  memory. Neither moves because a guest behaved well, so nodes are not
+  packed harder on that basis.
+
+### Only new domains get it
+
+A libvirt domain definition is persistent, and Shaken Fist renders the
+domain template only when the hypervisor has no definition for the
+instance. An instance that already exists therefore does **not** gain
+free page reporting from a power off and power on, a hypervisor
+reboot, or a redeploy that ships the new template -- it has to be
+recreated. This is most visible on exactly the long-lived instances
+that stand to gain the most. See
+[Upgrades](upgrades.md#free-page-reporting-applies-to-new-instances).
+
+Once an instance is recreated, the `rss kb` figure in its usage events
+steps down, because the qemu process really is holding less memory.
+It is not comparable with the same instance's historical values. The
+`actual kb` figure in those events is the balloon allocation and does
+not move, so anything billing on allocation is unaffected.
 
 ## Load-aware ordering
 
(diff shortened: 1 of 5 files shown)


mikalstill:
Thanks — round 1 was 1 fix / 3 doc / 2 consider / 1 info. Pushed
4526f10 addressing all of them except one, which I'm declining below.

**#3 (fix) — no test asserts the rendered domain XML.** Covered at both
levels, because they are two different claims:

* `shakenfist/tests/test_libvirt_template.py` renders the template across
  four variants (i440fx and q35, spice and vnc, plus UEFI with secure
  boot) — the jinja conditionals mean no single render sees the whole
  file — and asserts the memballoon carries `model='virtio'`,
  `freePageReporting='on'` and a ten second stats period. It also parses
  every variant, which is the guard for an illegal `--` inside an XML
  comment. That one earned its keep immediately: rewriting the NOTE for
  this commit introduced a `--`, and the test caught it before it could
  break every instance start.
* A functional test in `cluster_ci_tests/test_events.py` reads the
  `libvirt domain XML` mutate event from a real instance and asserts the
  attribute is in the XML libvirt was actually handed. The unit test only
  proves this repository's copy of an ansible file is right; this proves
  the hypervisor rendered it, so a node running a stale `libvirt.tmpl`
  fails rather than silently producing domains without reporting. It
  rides on the existing events test class and its network, so it costs
  one instance create.

Every new assertion was mutation tested — dropping the attribute, setting
it off, `model='none'`, dropping the stats element, changing the period,
and injecting a `--` each fail with the intended message.

**#1 (doc) — accounting claim was only half true.** Confirmed, and you had
it exactly right. `_has_sufficient_ram()` pre-filters on `memory_available`,
which the resources daemon takes from `psutil.virtual_memory().available`,
so a node running reporting-capable guests really will admit work it
previously would not. The NOTE now separates that measured pre-filter from
the allocation-based checks (`memory_total_instance_actual`, the
`scheduler_node_capacity` ledgers) that genuinely do not move, and says
explicitly that the guest can re-fault those pages at any time.

**#2 (doc) — existing instances do not pick this up.** Also confirmed
against `_power_on_inner()`. Now documented in
`docs/operator_guide/upgrades.md` as its own section, and in the template
NOTE: a power cycle, a host reboot and a redeploy all leave an existing
instance on its old XML, and recreating is the supported route. The
`virsh undefine` workaround is written down with its caveats and labelled
a workaround. Re-defining on power on is a real change with its own risks
and is not going in here, as you said.

**#4 (doc) — undocumented outside the template.** New *Guest memory
returned to the host* section in `docs/operator_guide/scheduler.md`,
linked from the RAM admission stage: what it does, the guest kernel
requirement and its negotiation, that page cache is not returned, and the
two-checks split above. It also absorbs **#6** — the `rss kb` step-down in
usage events, and why `actual kb` does not move.

**#7 (consider) — NOTE placement.** Taken, it was a one-liner. The
reporting NOTE now sits above `<memballoon>` and the stats comment stays
with `<stats>`.

**#5 (consider) — a jinja variable to disable it.** Declining, and
recording why here rather than silently. Shaken Fist is deliberately
opinionated, this is the right default, and a knob nobody sets is a config
surface plus a template conditional carried forever — the cost is
permanent and the benefit is hypothetical. The trade-off you raised is a
real one, so the auto-filed issue is the right home for it: if a
latency-sensitive workload ever demonstrates the THP-shattering cost on a
real cluster, that is the evidence to add the variable on, and it will be
a small change then.

`pre-commit run --all-files` passes clean.

Issue #375, opened 2020-09-09
Ansible CI hard codes an IP, but that might already be (randomly) allocated to another instance


mikalstill:
fatal: [sf-1]: FAILED! => {"changed": false, "meta": "Failed to parse JSON:\n[[command: sf-client --json instance create cirros_netspec 1 1  -D size=8,base=cirros  -N network_uuid=cbfc5d6b-0fd4-4861-940f-c9ef33dee81c,address=192.168.242.66,macaddress=02:42:4d:0a:86:fe -p  sf-1 ]]\n[[stdout: ERROR: Resource in use: address 192.168.242.66 in use\n]]\n[[stderr: ]]", "msg": "Error manipulating instance", "params": {"cpu": "1", "disks": null, "diskspecs": ["size=8,base=cirros"], "name": "cirros_netspec", "networks": null, "networkspecs": ["network_uuid=cbfc5d6b-0fd4-4861-940f-c9ef33dee81c,address=192.168.242.66,macaddress=02:42:4d:0a:86:fe"], "placement": "sf-1", "ram": "1", "ssh_key": "", "state": "present", "user_data": "", "uuid": null}}

mikalstill:
Ansible CI has been removed for a while now.

PR #3930, opened 2026-08-27
Close out phase 00a, plan phases 4b and 4c, land 4b's server half

mikalstill:
> **Do not enqueue until shakenfist/client-python#375 is merged to that
> repository's `develop`.** The functional jobs build their client wheel
> from a `client-python` checkout at `develop`, so the rewritten test in
> this branch needs the verbs to be there first. Opening the PR now is
> safe — the `(collection)` matrix is skipped on `pull_request`.

Closes out phase 00a of `PLAN-scheduler-reservations`, plans phases 4b
and 4c, and implements the server half of 4b.

## Phase 00a is complete

Its last outstanding item was a post-deploy observation: does a real
sfcbr CI burst still concentrate placement on the network+database node?
**It does not.** Over an hour of live CI, sf-1 took 0.78× its
proportional share of new placements, and was below parity at every
sample size measured (0.69× / 0.84× / 0.73× / 0.82× / 0.78× at n = 21,
23, 26, 32, 35).

The first snapshot said the opposite — 1.82× — and the plan records why
that was wrong, because the traps are reusable: sfcbr mixes static
runners placed days ago with ephemeral ones; the ephemeral population at
any instant is small enough that one extra 4-vCPU runner reads as 1.90×;
and the instance view publishes no `created_at`, so age has to come from
`agent_system_boot_time`.

Two other 00a criteria are evidenced by the scheduler's own events
rather than by inspection: sf-1 publishes `cpu_schedulable` 6 where
same-sized nodes publish 10 (the infra-role reservation is live), and
`cpu_schedulable_from_fallback` is `false` everywhere (the phase-1 typed
columns are being read).

**One finding is recorded rather than fixed, and fed to phase 6.** The
load stage narrows to the lowest occupied bucket *before* headroom
weighting, so it can discard nodes with far more room than the ones it
keeps — 8 of 60 observed decisions did, and 34 left a single survivor.
`cpu_load_1` measures activity rather than occupancy, so a node packed
with idle CI runners ranks as the emptiest in the cluster. It is not
concentrating placement today only because the capacity filters reject
that node before ranking runs, which is a weaker guarantee than it
looks.

## Two new phases, and a renumbering

The master plan had no phase implementing D18, the conductor contract
for namespace claims — the obligation lived in one subordinate clause of
the phase 4 stub, and nothing had implemented it. That gap matters
because D16 makes the ceiling advisory for one release *so that
exceedances are observed before they are refused*, and between phase 4
merging and this branch being written, every claim that had ever existed
on sfcbr was created by a test and deleted by the same test. Phase 5
would otherwise flip `CLAIM_ENFORCEMENT_HARD` on a measurement period
that measured nothing.

* **Phase 4c** (new) — conductor claim integration, landing in
  `shakenfist/private-ci`. Its plan lives here rather than with its
  code; the reasoning is decision E1 in the plan.
* **Phase 4b** — client support, renumbered from phase 9 and moved ahead
  of the hard ceiling.
* Push audit becomes phase 9, keeping the table contiguous.
* Phase 5's gate is corrected from "one release after phase 4" to "once
  4c's observation record exists".

## Phase 4b's server half

The functional claims suite now drives every request through the client
verbs. It used to use `apiclient.Client._request_url()` because phase
4's D7 held that a test written against new `apiclient` methods could
not pass in CI until a client release existed.

**That was wrong, and had been since 2026-06-24.** Cluster CI does not
install the released client — `shakenfist/actions` checks out
`client-python` at `develop` and the collection builds a wheel from it.
`docs/developer_guide/ci.md` now says so, because the belief cost a
deliberate design compromise and was restated in four places.

The `_claim_api()` adapter stays, and the verbs go underneath it. Every
status assertion in the file is unchanged, and `shakenfist_ci/retries.py`
keeps both its `(status, body)` contract and its freedom from
`shakenfist_client` imports — which `test_ci_claims_headroom.py` checks
by loading it by path.

## Testing

The `(collection)` matrix is skipped on `pull_request`, so a green PR
would say nothing about whether the rewritten file reaches its first
assertion. It was driven against a live cluster instead: every dispatch
path, both success and refusal, including that failure messages still
name the path they asked for. The headroom-tolerant wiring assertion was
mutation-tested by routing `_create_claim` off it, which fails exactly
one test.


mikalstill:
Addressed the review. All four `fix` items and six of the `consider` items are in the two commits above.

The high-severity item is real and stands: the five verbs live in shakenfist/client-python#375, which carries all of them plus `ServiceUnavailableException` with green checks, but is **open, not merged**. That is now recorded as a "Cross-repository ordering" section in the phase 4b plan rather than only in this pull request's first paragraph, with the enqueue hold and the reason the failure would be invisible before the merge queue. The plan table now agrees with the branch: steps 4 and 5 complete, steps 1 and 2 pointing at #375, the master plan row `In progress`, and the four satisfied definition-of-done boxes ticked.

On the `consider` items: `tearDown` now runs the namespace delete in a `finally`, so a non-`APIException` cannot leak a namespace on a shared cluster; `test_ci_claims_headroom.py` grows an AST guard asserting `_claim_api()` names exactly the five verbs and that no `_request_url` **call** survives, both mutation-tested (renaming a verb and reintroducing the call each fail exactly the guard that names them); a body passed with GET or DELETE now raises instead of vanishing; the adapter docstring says its success status is structural rather than observed.

I did not take the suggestion to trim the master plan's restatement of the CI premise to a pointer -- it is a historical record of why a decision was made as much as a live document -- but the test docstring now names the `ci.md` heading exactly, so the pointer survives that file growing.

Issue #3931, opened 2026-08-27
sf-sidechannel shutdown tears down monitors instead of executors, and can KeyError


mikalstill:
`sf-sidechannel`'s shutdown path applies its executor teardown to the
monitor dictionary, so executors are never asked to exit, the wrong
thread is joined, and the daemon can die with a `KeyError` instead of
stopping cleanly.

Found while surveying the sidechannel daemon during phase 5 planning of
`docs/plans/PLAN-agent-operation-deadlines.md`. Not caused by that plan
— introduced in `69f27b7ad` (2024-12-15, "Convert to using threads to
resolve grpc segfault issues").

## The defect

`_request_thread_exit()`
(`shakenfist/daemons/sidechannel/main.py:1297-1310`) takes the thread
record to shut down as its `t` argument, but its body reads and mutates
`self.monitors` regardless of what it was passed:

```python
def _request_thread_exit(self, instance_uuid, t):
    daemon.set_abort_path(
        t['object'].abort_path, 'from _request_thread_exit')
    add_event(
        EVENT_TYPE_AUDIT, 'instance', instance_uuid,
        'side channel monitor instructed to exit')
    self.monitors[instance_uuid]['thread'].join(0.5)      # <-- always the monitor

    if not t['thread'].is_alive():
        del self.monitors[instance_uuid]                  # <-- always the monitor
        daemon.clear_abort_path(t['object'].abort_path)
        add_event(
            EVENT_TYPE_AUDIT, 'instance', instance_uuid,
            'side channel monitor finished')
```

`_request_all_threads_exit()` (`:1284-1295`) calls it twice — once per
monitor, and then once per executor:

```python
all_executors = list(self.executors.keys())
for instance_uuid in all_executors:
    self._request_thread_exit(
        instance_uuid, self.executors[instance_uuid])
```

Only the `set_abort_path` and `clear_abort_path` calls use `t`. The
join, the liveness-driven deletion and both audit events do not.

## Consequences

For an executor call:

1. **The wrong thread is joined.** `self.monitors[instance_uuid]['thread'].join(0.5)`
   waits on the instance's monitor, not its executor, so the shutdown
   never actually waits for an in-flight agent operation to wind down.
2. **The executor entry is never removed.** `del self.monitors[...]`
   leaves `self.executors` untouched, so an executor's record outlives
   the shutdown.
3. **A live monitor can be deleted on an executor's behalf.** The
   `is_alive()` test reads the executor's thread but the deletion
[...]

PR #3941, opened 2026-08-28
Agent operation deadlines phase 5: retry and the reaper

mikalstill:
Phase 5 of `docs/plans/PLAN-agent-operation-deadlines.md`. Phase 4 made
the timing budgets bite, so an operation that outlives its deadline or
stalls mid-transfer now reaches a terminal state instead of hanging.
This phase makes a *stall* survivable rather than fatal, and closes the
wedge shapes phase 4's three enforcement points cannot see.

The planning commit is included, as is the convention for this plan.

## Commits

| Commit | What |
|--------|------|
| `7e68055cb` | The phase plan, and the master plan corrections its survey turned up |
| `5747a80dc` | The `executing -> queued` edge, `record_attempt()`, `clear_results()`, `AGENT_OPERATION_MAX_ATTEMPTS` |
| `e14b6a0fa` | The retry policy, and both executor exit paths |
| `54dd0f791` | Terminal-only pop **and** the node-local reaper |
| `01e57d50e` | The pre-connection wait becomes abortable |
| `da939cc91` | The state machine page, and closeout |

## Review notes

**`54dd0f791` is deliberately one commit, not two.** Today an operation
orphaned in `executing` leaks while its queue drains, because the pop
retires an `EXECUTING` head. After terminal-only pop the same event
would wedge that instance's queue permanently, and the reaper is what
turns it back into a queue which drains. Either half alone is a
regression rather than an unfinished improvement.

**`e14b6a0fa` lands before `54dd0f791`, and its rewritten `finally`
comment is one commit ahead of the code.** The two are ordered either
way and neither order is clean: the reaper calls
`resolve_abandoned_operation()`, so pop-first would commit a
`NameError`. Between them a retried operation would orphan rather than
re-dispatch, so the pair must not be merged separately.

**Retryability is a property of the whole command list, not of the
command in flight.** A retry restarts at index 0, so an
`[execute, get-file]` operation stalling in the `get-file` would re-run
the `execute` — the side effect phase 0 decision 6 forbids. No endpoint
builds a mixed list today, so the two readings agree in practice and no
test can demonstrate the difference; this is the version that stays
correct when one does. This is the decision most likely to be argued
with, and the cheaper reading would have shipped a smaller diff.

**The reaper acts on evidence, never on suspicion.** Two of its three
cases are the absence of a thread, which this node observes directly
because the instance is placed here and the executor is a thread in this
process. The third is a wall-clock deadline, which is absolute and so
cannot be wrong about a thread still making progress. A live executor
inside its budgets is left alone. It resolves *before* aborting, because
the executor's `finally` only rewrites an operation still in
`executing`.

## Two shapes the reaper cannot see

Both found while implementing, both recorded in the phase plan's future
work rather than left to be rediscovered:

- An instance with no monitor entry is never examined. Self-healing
  within 30 seconds in the common case; not self-healing for an
  instance that has permanently lost its agent channel, which cannot
  run agent operations anyway. The operation leaked identically before
  this phase — the pop retired the *entry*, never the *state* — so what
  is new is only that the queue stops draining. Fixing it needs an
  instance-placement list the daemon does not keep.
- An operation created with `deadline_seconds=0` whose live executor
  wedges before connecting cannot be reaped, because the reaper's only
  evidence for a live executor is `deadline_passed()` and there is no
  deadline to pass. Decision 8 forbids the alternative. Same family as
  phase 4's unbounded-operation gap, and it belongs beside it in phase
  7's release note.

## Verification

- All seven of the phase plan's Definition of done checks pass.
- Full unit suite: 3,747 tests, 3,630 passed, 117 skipped, 0 failed.
- `pre-commit run --all-files` green before every commit.
- Fifteen mutations across the four code steps, each failing a specific
  named test. Two exposed real gaps rather than confirming coverage:
  deleting the `record_attempt()` call passed every test (so nothing
  proved the attempt cap could ever fire), and an empty command list was
  vacuously retryable because `all()` over nothing is true. Both fixed,
  with tests.
- The state machine diagram was checked against `state_targets`
  programmatically in both directions: 21 edges, none in the code
  missing from the page and none on the page absent from the code.

## Out of scope

`client-python` is phase 6. The release note, operator and user guides,
and functional CI coverage are phase 7, which writes the timing story
once. `docs/developer_guide/state_machine.md` is the single
documentation exception here, because it is a rendering of
`state_targets` and would otherwise contradict the code.

Issue #3931 was filed from this phase's survey and is deliberately not
fixed here: `sf-sidechannel`'s shutdown applies its executor teardown to
the monitor dictionary, which predates this plan and wants its own
change and its own test.

*(Implemented with Claude Code, using the sub-agents the phase plan's
step table specifies.)*


mikalstill:

Commit `e8999d881` addresses all eleven items from the previous review.
Three of them changed something the phase plan had already decided, so
those corrections are made at their source in
`docs/plans/PLAN-agent-operation-deadlines-phase-05-retry.md` as well as
in the code, and the plan now carries a *Review follow-ups* table
recording the outcome of every item.

**Item 1 (the shared abort path)** was real, and the diagnosis was
exactly right. The fix is to the derivation rather than the symptom:
`SideChannelJob.__init__()` now takes an `abort_name`, so the path is
decided once instead of being assigned by the base class and overwritten
by a subclass afterwards. Assigning it in `SideChannelExecutorJob`
after `super().__init__()` would have left a second bug in place — the
base constructor *clears* whatever path it derives, so building an
executor while the daemon was shutting down cleared the monitor's abort
file and un-stopped it. There is a test for that specifically. The
executor's file is keyed by instance rather than by operation, matching
the one-executor-per-instance invariant the dispatcher enforces, and
`reap_instance_executors()` clears it once the thread is confirmed dead
so a reaper-set file cannot outlive the executor it stopped.

**Item 2 (attempts burned before reaching the agent)** is now counted on
the transition into `executing`, guarded so it fires once per attempt
rather than once per command in the list. That also dissolves the
ordering concern the old comment justified itself with: everything which
reads the counter requires the operation to already be `executing`, so
the increment for an attempt always precedes its own comparison against
the cap.

**Item 3 (the config description)** took the second of the two options
offered. Enforcing the cap at dispatch would have invented a bound on a
failure mode this phase does not handle — an operation whose executor
never connects — and the plan's own *Future work* section already
records that gap. The description now says the cap bounds retries of an
operation which reached the agent, and says explicitly that it is not a
bound on dispatch.

**Item 11** is answered rather than changed, in agreement with the
reasoning given: one thread writes that column, the field mask is
correct so there is no cross-attribute lost update, and moving the
increment to the executing transition does not add a writer.

**Item 10** stays deferred to phase 7, which the review agreed is
consistent with the agreed scope, and now has a *Future work* entry
naming the three reaper cases and the two shapes it cannot see so it is
not lost.

Items 4, 5, 6, 7, 8 and 9 are fixed as suggested.

## Verification

Each fix was mutation tested, and each mutation is caught by a specific
named test:

| Mutation | Caught by |
| --- | --- |
| Executor shares the monitor's abort path again | `SideChannelAbortPathTestCase` (3 tests) + both wait tests |
| Count an attempt per command rather than per attempt | `test_a_second_command_is_the_same_attempt` |
| Never count an attempt | `DispatchRecordsAnAttemptTestCase` (3 tests) |
| Executor exit leaves the `.partial` file | `test_an_executor_exit_removes_it_too` |
| Register the executor after the thread starts | `test_the_entry_exists_before_the_thread_is_started` |
| Sweep does not clear the abort path | `test_sweeping_a_dead_executor_clears_its_abort_path` |
| `reaper_attempts` is never pruned | `test_reaping_a_monitor_drops_its_rate_limiter_entry` |
| No lower bound on the attempt cap | `AgentOperationAttemptCapTestCase` (2 tests) |

Full suite 3,762 tests, 3,645 passed, 117 skipped, 0 failed (up from
3,758). All seven of the plan's Definition of done checks still pass.
`pre-commit run --all-files` exits 0.

## One note for the reviewer

This branch predates [#3933](https://github.com/shakenfist/shakenfist/pull/3933),
which fixed `_request_thread_exit()` on `develop` (issue
[#3931](https://github.com/shakenfist/shakenfist/issues/3931), filed
from this phase's survey). That function is deliberately untouched here
so the merge takes `develop`'s version cleanly. The two changes are
complementary: `develop` makes the shutdown path operate on the right
dictionary, and this commit gives each thread its own abort file to
operate on.

mikalstill:

Commit `4309fedf6` addresses all twelve items from the second round.
Two of them were consequences of the first round's fixes, and both are
recorded that way in the plan's *Review follow-ups* section, which now
carries a table per round.

**Item 1** was real and self-inflicted. `Thread.is_alive()` is False for
a thread which has never been started, exactly as it is for a finished
one, and `join()` raises `RuntimeError` there — both confirmed against
the interpreter rather than taken on the description. The sweep now
skips a thread whose `ident` is None. The second half of the suggestion
matters more than it first appears and is also taken: a `start()` which
raises removes the entry before re-raising, because an unstarted entry
is never collected *and* makes the instance look permanently busy to the
reaper, so an operation left executing on it would never be resolved
again for the daemon's lifetime.

**Item 2** is the interesting one. Rather than pulling #3931's fix
forward, `_request_all_threads_exit()` now signals every thread before
joining any. #3931 is already fixed on `develop` by
[#3933](https://github.com/shakenfist/shakenfist/pull/3933), so
duplicating it here would conflict on merge for no benefit; signalling
up front is correct whichever version of `_request_thread_exit()` is in
the tree, and is better on its own terms, since signalling inside the
loop made each thread's notice wait on the previous thread's half second
join.

**Item 4** changed a decision, so it is recorded at its source. The
reaper's no-executor case now errors. Decision 3 splits the terminal
outcomes on what went wrong rather than on who noticed, and this is an
executor which went away — the only difference from the exit handled in
`execute()`'s `finally` is that the process died before the `finally`
could run. The `FINAL_OBJECT_STATES` consequence was checked and is what
makes it more than cosmetic: expiring there swept the daemon-restart
case for hard deletion while the identical `finally` case persisted.

Items 5, 6, 7, 9, 11 and 12 are fixed as suggested — including item 6,
which was rated not worth doing on its own; it is one line and one extra
peek per executor death, against up to 30 seconds of blocked queue.
Items 3 and 8 stay deferred to phase 7 and the *Future work* entry now
names the three reaper cases, the two shapes it cannot see, the
`object_states.message` strings an operator will grep for, and the
not-a-bound-on-dispatch rule. Item 10 needed no change.

## Verification

Five mutations, each caught by a specific named test:

| Mutation | Caught by |
| --- | --- |
| Sweep joins an unstarted thread | `test_an_unstarted_executor_is_not_joined` |
| A failed `start()` leaves the entry behind | `test_a_failed_start_does_not_leave_the_entry_behind` |
| No up-front shutdown signal | both `RequestAllThreadsExitTestCase` tests |
| Case one expires instead of erroring | `test_a_daemon_restart_drains_the_queue`, `test_a_dead_thread_is_swept_and_resolved_in_one_pass` |
| `clear_results()` emits no event | `test_clear_results_emits_an_event` |

One of those caught a weak test written in this same change, which is
worth flagging: `test_every_thread_is_signalled` initially passed
against the bug, because Mock threads report alive, so the monitors were
never deleted and #3931's `KeyError` never fired. The fixture now makes
the monitor threads stop promptly, which is what arms it.

Full suite 3,767 tests, 3,650 passed, 117 skipped, 0 failed. All seven
of the plan's Definition of done checks still pass. `pre-commit run
--all-files` exits 0.

PR #3957, opened 2026-08-29
Plan phase 6, and dispose of issue 3565

### LINKED ISSUE #3565: test_affinity flake: soft affinity loses to resource filters under suite concurrency
## Occurrence

Merge queue run for PR #3563 on 2026-07-30:
https://github.com/shakenfist/shakenfist/actions/runs/30512428320
(job "Debian 12 tier (collection) / Smoke tests (collection)")

```
shakenfist_ci.cluster_ci_tests.test_scheduler.TestAffinity.test_affinity

testtools.matchers._impl.MismatchError: !=:
reference = '27057ba6-2ae2-41ad-a8db-b54d16157fb4'
actual    = '38438f0a-cf3f-4297-8a2f-5e2c5728a290'
: Instances eb781a6d-1f8c-454b-8eb2-79a7d77b279e and
  73d3081a-1b33-4782-b5ea-423cbe8c63e6 should be on the same node but are not
```

inst1 (tagged `first-node`) was on node `27057ba6`; inst2 (affinity
`{'first-node': 100}`) landed on `38438f0a` instead. Notably inst3
(affinity `{'first-node': -100}`) *also* landed on `38438f0a`, so the
scheduler put the attract and repel instances on the same node.

## History: this is not the old RMW flake

This test flaked before and was root-caused to the `node_attributes`
full-row read-modify-write race (`observe_this_node` reverting
`Node.add_instance`, so the tagged neighbour became invisible and every
candidate scored 0). That cause was fixed in 394336212 ("Add attribute
field masks and placement refs."): `observe_this_node` now writes with a
field mask, and placement lives in `object_references`
(`INSTANCE_LOCATION` rows) written synchronously by
`Instance.place_instance()`. This recurrence is post-fix, so it is a
different (or residual) mechanism, and it had no tracking issue until
now.

## Likely mechanism: soft affinity loses to resource filters under load

In `scheduler.py`, the headroom filters (`sufficient_idle_cpu`,
`sufficient_idle_memory`, `sufficient_free_disk`,
`sufficient_idle_disk`) all run **before** affinity scoring, and
affinity is only ever a preference among the survivors. If the node
hosting inst1 is dropped at a headroom stage while inst2 is being
scheduled, every surviving candidate scores affinity 0 and placement
falls through to the CPU-load bucket ordering — inst2 lands wherever is
least loaded.

Circumstantial support from this run: in the *same* test worker,
`test_network_plumbing_lifecycle` failed with 507
`No nodes remaining at scheduling stage sufficient_idle_cpu` at 04:24
(#3498), and `test_affinity` failed at 04:30. The cluster was
CPU-starved at exactly the window inst2 was scheduled, so inst1's node
being dropped (or barely losing the load-bucket tiebreak) is very
plausible. This would make this flake a second face of the same
suite-concurrency headroom race as #3498.

## How to confirm on the next occurrence

The scheduler now emits per-stage audit events including the `dropped`
map for each filter stage and a `schedule have highest affinity` event
with the full per-candidate `affinity_detail` breakdown. Pull inst2's
events from the collected CI logs and check:

- If inst1's node appears in a headroom stage's `dropped` map →
  confirms the mechanism above.
- If inst1's node survived filtering but scored 0 with inst1 listed as
  `skipped: instance row not found` or absent from `considered` →
  placement visibility is still broken somewhere and the RMW-class bug
  has a residual path.

It may also be worth teaching the CI failure-collection step to dump the
scheduler audit events for instances named in failed assertions, so this
evidence is captured automatically instead of being lost with the
ephemeral cluster.

## Candidate fixes (pick a direction)

1. **Define affinity-vs-headroom semantics in the scheduler.** Today a
   positive affinity silently degrades to "anywhere with headroom".
   Options: treat a strong positive affinity like requested placement
   (507 if the target node lacks headroom), or fold the affinity score
   into the final load-bucket ordering instead of applying it strictly
   before.
2. **Make the test tolerant of what soft affinity actually promises**:
   assert from the audit events that the scheduler scored inst1's node
   highest (or legitimately dropped it), rather than asserting final
   co-location that the product does not guarantee under load.
3. **Reduce the race in CI**: serialize the scheduler tests (or lower
   suite concurrency for the tier topology) so headroom starvation from
   neighbouring tests stops perturbing placement assertions. Cheapest,
   but hides the undefined semantics rather than fixing them.

Related: #3498 (same headroom starvation, different test), #3496
(requested placement ignored).

### LINKED ISSUE #3813: Scheduler demand guard is arithmetically unsatisfiable on nodes with fewer than four schedulable threads

mikalstill:
Documentation only, one commit: the phase 6 plan, its Execution-table row,
and the two corrections at source the plan commits to.

## The finding that reshaped the phase

Phase 6's stub told this phase to rule out a competing explanation for #3565
before spending its decision budget. That instruction is discharged, and the
answer is neither of the two mechanisms the plan documents argue about.

A fully traced occurrence was posted to #3565 on **2026-08-26** — after phase
4a fixed #3813 on 2026-08-24, so from a cluster where the spreader works. The
candidate set had already collapsed to **one node** before affinity was
scored. inst3 requested anti-affinity, correctly scored that node `-100`, and
placed there anyway, because a scorer given one candidate has nothing to do.
The issue's own words:

> here it did not lose a tiebreak, it was **never consulted** in any
> meaningful sense

And the half of the assertion that *passed* was the same forced choice. The
run says nothing about affinity in either direction.

It is not lost spreading, and it is not affinity losing a tiebreak to load
ordering.

## The decision most likely to be argued with

**F2 closes #3565 on a test change, not a scheduler change.** That is the
shape of an excuse and should be pushed on. The defence is the trace and only
the trace. Against a single-candidate set:

- D6 position 1 (hard require only) turns the flake into a 507 — which the
  issue itself calls "the same ejection with a different traceback".
- D6 position 2 (soften the ceiling) has nothing to soften: the affinity
  target *was* the surviving candidate, and the instance was placed on it.
- D6 position 3 is the issue's candidate fix 2 under another name, and is the
  one the evidence supports.

The honest framing: #3565 was mostly a specification bug. The product never
promised co-location under contention; the test assumed it did, and every
investigation since has hunted a scheduler defect behind an assertion that was
too strong.

**F7 declines** D6's ceiling-softening rather than deferring it. A preference
that can raise a ceiling puts a ranking input inside the transaction whose
correctness argument depends on it being arithmetic — which is exactly what
phase 3 spent its effort establishing.

**F3 keeps this from being flake suppression**: the rewritten test *skips
rather than passes* on a degenerate candidate set. A single-candidate run
carries no information, and a green there is the same false pass the trace
describes.

## Other survey findings

- **The binary affinity model does not exist at all.** None of
  `require_with_tag` / `require_without_tag` / `prefer_with_tag` /
  `prefer_without_tag` appear anywhere in the tree. This is a first
  implementation, not a migration.
- **`test_affinity` is itself a user of the weighted form it deprecates**
  (`{'first-node': 100}` / `{-100}`), so the transition mapping is exercised
  from the moment it exists and the weighted form cannot be removed in the
  same change.
- **The binding stage has moved from CPU to memory** — a fix aimed only at
  `sufficient_idle_cpu` would now miss.
- **The audit events already carry what the test needs**, and
  `_add_scheduler_detail()` already fetches them, which makes the
  expensive-looking half of the phase cheap.
- **D6's line references have all moved** (`:473-481` → `:487-501`, `:611-631`
  → `:529-600`, `:260` → `:187-256`).

## Explicitly not adopted

The `507 ... sufficient_idle_cpu` family (#3772). It comes from an admission
stage that runs *before* ranking; when it refuses every candidate no ranking
model would have helped. `PLAN-ci-cloud-sizing` owns it and has the better
analysis. Written into the plan as F6 because it is the loudest
scheduler-shaped signal in CI and is repeatedly mistaken for evidence about
ranking.

## Corrections at source

The master plan's phase 6 stub gains the traced mechanism, keeping the
superseded reasoning beneath it. D6 gains current line references and the
CPU→memory note. D6's three positions are left as written — a decision record
should show what was on offer.

## Checks

The definition-of-done checks were **run at planning time, not merely
written**. One was wrong: `grep "inst\['node'\]"` matches nothing, because the
assertions name `inst1`/`inst2`/`inst3`. Corrected to the grep that works,
with the lines it actually returns (125, 129). Finding 7's line references
corrected from 79/98 to 80/99.

`pre-commit run` passes: documentation links and anchors resolve, plan
statuses and index arithmetic agree. (The status vocabulary check caught
`Planned` as invalid and it is now `Not started`, matching phase 4c.)


mikalstill:

mikalstill:
All eight items addressed in `23fdc6111`. Each claim was checked against the
code before acting on it, per this repository's convention of verifying review
claims rather than trusting them — **all eight held**, including one that
inverted the intent of a decision the plan had just made.

**1 — F3's skip gate read the wrong event.** Confirmed and fixed. `schedule
final candidates` is published after `candidates = narrowed` (`scheduler.py:658`)
has cut the list to the winning affinity tier, so on a healthy cluster where
inst2's affinity works exactly as intended the tier is one node, the event holds
one candidate, and the gate would have skipped every run where affinity was
*correct* — permanently and silently, since a skip is green. F3 exists to stop
the suite claiming coverage it does not have, and as written it produced exactly
that. The count now comes from `len(affinity_detail)` in `schedule have highest
affinity`. All three sites corrected, and step 1 carries a note on why the
obvious event is the wrong one.

**5 — the validator hole is real and live.** Confirmed at the interpreter:
`int(['a'])`, `int({'x': 1})` and `int(None)` all raise `TypeError`, which the
existing `except ValueError` does not catch — so a malformed `affinity` value
returns **500 today**, from instance create (`:797`) and both metadata endpoints
(`:1375`, `:1425`). The binary shape is list-valued and would have hit it
immediately. Step 3 now widens the except clause as part of adding the shape,
step 2 files it separately since it is a public API fault worth tracking
regardless of this phase, and the DoD gained a criterion. Also recorded that
`int()` accepts floats and numeric strings, so "values must be integers"
describes intent rather than coercion.

**4 — the rationale argued against its own mechanism.** Correct, and the sharpest
catch here. F4 discarded caller magnitudes because they are summed over an
unbounded neighbour count, then adopted ±1 *per matching instance* — summed over
the same unbounded count. Settled explicitly in the plan: the scoring is
**count-proportional**, not set-membership, because a node carrying five members
of an affinity group really is more "with the group" than one carrying a single
member, and set-membership discards exactly the signal a pack-or-spread request
is about. F4's rationale is corrected to the real one — the *scaling*, not the
summation: choosing `100` over `1` picks a multiplier on a neighbour count the
caller cannot predict.

**3 — DoD contradicted F4.** Narrowed to same-sign, single-tag specs, with the
mixed-magnitude divergence now an explicit assertion rather than an unstated
exception. Your `{'a': 100, 'b': 1}` example is quoted in the criterion.

**7 — the grep banned a legitimate use.** Restated as "no assertion compares
`inst2['node']` or `inst3['node']`", which is what F2 actually guarantees and
leaves `inst1['node']` available as the lookup value step 1 needs. Replacement
grep run: returns 125 and 129 today, as expected.

**2 — broken table row.** Confirmed: a literal newline after "whose". Rejoined;
all six rows now parse with six cells.

**6 — dedup scope.** Now stated as **once per instance**, with per-process
(resets on restart) and per-schedule (runs on every create *and* reschedule)
explicitly ruled out, plus an instruction to confirm against
`shakenfist/data/database_load_budget.yaml`.

**8 — `affinity_detail` shapes.** Step 1 is now told to count entries without
indexing inside them, since the `node row not found` shape (`:548-551`) has no
`considered` key and would turn a transient failure into an unreadable
`KeyError`.

Item 1 is the one worth dwelling on: the plan made a decision specifically to
avoid false green, and then operationalised it in a way that would have produced
false green on every correct run. Thank you for catching it.

mikalstill:

mikalstill:
All eight actionable items addressed in `5df69605a`. Verified against the code
first; all eight held.

**1 — the same defect, through a different door.** Worth stating plainly: last
round's fix reproduced the failure it removed. `find_candidates()` runs more
than once per instance on the happy path — unforced at
`external_api/instance.py:867`, forced against the assigned node alone at
`operations/node_inst_netdesc_op.py:156`, and forced again at `:869` when the
instance is already placed. Every call publishes its own `schedule have highest
affinity` against the same instance, and a forced call yields an
`affinity_detail` of exactly one entry. A test taking the most recent — what a
descending fetch hands back first — would have skipped unconditionally on every
healthy create.

The gate now names the **unforced pass**, located through the discriminator you
pointed at: `schedule inputs`'s `forced_candidates` (`scheduler.py:438`). Both
wrong sources are recorded in F3 so neither returns, and the DoD gained your
suggested check — **the test must not skip on a healthy three-node run** —
which is the criterion that would have caught both attempts.

**2 — and the read compounds it.** Confirmed: `limit=100`
(`external_api/instance.py:1203`) over timestamp-descending rows
(`mariadb.py:6139-6146`), so the fetch returns the newest hundred while the
create-path scheduling events are the *oldest* the instance has. Step 1 now
requires an explicit limit, and `self.fail()` rather than `skipTest` when the
unforced pair is absent — a missing event means the read was wrong, not that the
run was degenerate.

**3 — a genuine inversion, one structural level up.** Confirmed: an `affinity`
value that is a list is already refused by `isinstance(value, dict)`
(`:1399-1401`), and `None` by `if not value` (`:1387-1388`). The 500 is only in
the per-tag coercion. Corrected in both the DoD and step 2's issue brief, with
`{'affinity': {'first-node': ['a']}}` versus `{'affinity': ['a']}` spelled out
so the level cannot be misread again.

**4 — the examples taught a feature that does not exist.** `ssd` and `gpu` read
as node capability tags; the mechanism matches the `tags` metadata of instances
already placed on a candidate node (`instance.py:783-784`). Replaced with
instance-shaped tags, with a sentence saying so explicitly, and the risk
reworded away from "a tag no node carries".

**5 — decided rather than left to step 3.** The scope is **inherited**:
`require_without_tag` is within-namespace. Crossing it would let a caller learn
which tags exist in namespaces they cannot read by watching which creates get
refused — a trust-boundary change, and far larger than an affinity model should
make in passing. If cross-tenant isolation is ever wanted it should be built as
isolation, with its own threat model. Step 5 must document the constraint in as
many words, and there is a DoD bullet for it, because you are right that the
first operator to reach for it will read it as isolation.

**6 — the marker problem dissolves if the event moves.** Rather than choosing
between an attribute write and an event-history read on the scheduling hot path,
step 4 now emits **where the spec is accepted** — in `_validate_instance_metadata`,
the same function step 3 widens, which sees each weighted spec exactly once. No
durable marker, nothing on the hot path, and the warning reaches the caller at
the moment they submit the deprecated form rather than at some later reschedule.

**7, 8 — editorial.** F3's orphaned "in that case" restored to "on a
single-candidate run" and the paragraph rewrapped; the phase 00 correction's
over-width line split. (Confirmed line 330's 86 characters is a pre-existing URL,
not from this change.)

On **9 and 10**: thank you for re-verifying every line reference — that survey
finding was that stale references are how the previous corrections decayed, so
independent confirmation is worth more than my own. And `pre-commit run` is green
on the branch, including `check-plan-status`, which also caught `Planned` as an
invalid status token during planning.

Two rounds, and the same gate was wrong both times in the same way. The DoD check
you proposed converts that from something review has to catch into something one
run against sfcbr catches.

mikalstill:

mikalstill:
Third review addressed in 3cd14c85d. Nine items, seven actionable; every claim was checked against the tree before being acted on, and every one held.

**1 — the deprecation event site.** Correct, and it is the second round's fix failing the way the round before it did. Round two moved the event to `_validate_instance_metadata` on the argument that acceptance time beats consumption time. Acceptance time is still right; that function is not a place an event can come from. It is module level with the signature `(key, value)` (`external_api/instance.py:1384`), so there is no object to call `add_event()` on, and on the create path it runs at `:793-799` — before `Instance.new()` at `:810`. The instance does not exist yet. The dominant path would have emitted nothing at all, which is a deprecation warning that reaches nobody. Step 5 now splits an `_affinity_spec_is_weighted()` predicate out of the validator and emits at the three sites where an instance is in scope: the create path's metadata-initialisation loop at `:834-837`, and both metadata endpoints beside their existing audit events. The definition of done gains a criterion stated against the create path specifically, since an implementation covering only the two easy endpoints would have passed a looser one.

**2 — the 507.** Correct, and it was a decision I left for the implementer to trip over. New decision **F9** settles it before step 4 is written: the stage raises `AffinityConstraintUnsatisfiable`, **a subclass of `LowResourceException`** so preflight's redirect (`node_inst_netdesc_op.py:159-180`) keeps working untouched — redirecting is right for a constraint another node may satisfy — with an `except` clause ahead of the existing one on the create path answering **409**, naming the constraint and the stage. `except` clauses match in order, so that ordering is the whole mechanism, and F9 says so. The done bullet now names the code, and records that its previous phrasing was satisfied by the wrong answer.

**3 — the event read.** Both gaps taken, both with the precedents you cited. `event_type='audit', limit=1000`, with the 1000 ceiling (`:1193-1195`) given as the reason the audit filter is required rather than tidy — the brief also mandates `self.fail()` on a missing pair, so a create emitting more than 1000 events would otherwise have become a test failure blamed on affinity. And `request_id` is now the pairing key (`schema/event.py:84`, populated in `eventlog.py:82-86`, returned by `external_api/base.py:1479-1484`), with adjacency demoted to a fallback. The observation that the preflight call carries no `request_id` at all is the part that makes it unambiguous rather than merely better.

**4 — finding 6.** You are right and I could not reproduce it either. The create path's second walk re-walks the same list, but only when a candidate was refused on demand alone, and it waives the demand guard when it does (`:924-940`) — deliberate, can succeed, and the refusal did buy something. The preflight redirect does the opposite (`node_inst_netdesc_op.py:172-180`). The only site forcing a node an earlier pass already chose is preflight's opening call at `:156`. Finding 6 is rewritten as a candidate mechanism rather than a fact, with the superseded description kept so it is not quietly dropped, and F8 and step 2 now require the trace to be re-read against those three sites before filing — and an observation that says so if none matches. A wrong issue is worse than none.

**5 — splitting the 500.** Done. Widening the `except` clause depends on none of this phase's decisions and should not wait behind an API surface, a new scheduler stage and a back-brief gate. It is step 3 on its own; the model is steps 4 to 6.

**6 and 7 — cosmetic, and both real.** 47 spaced em dashes normalised to the spaced double hyphen every sibling plan uses exclusively, and the three drifted references corrected to 574-580, 596-600 and 1398-1400. A document whose own finding 2 is about stale line references gets read against that bar.

**8 — noted in F3.** Ruling out `schedule final candidates` as skipping "unconditionally" is exact for the inst2 assertion only; inst3's `{'first-node': -100}` makes the winning tier every node *not* carrying the tag, which is two of three. The conclusion is unchanged, because a source wrong for one of two assertions is unusable either way — F3 now says which half is which rather than presenting both wrong sources as failing identically.

**9** needed nothing; `check-plan-status.py` still passes, as does `pre-commit run --all-files`.

Three rounds, and the pattern across them is worth naming: each round's fix was correct about the *principle* and wrong about the *site*. Post-narrowing event, then right event but wrong pass, then right moment but a function with nothing to emit from. What has changed is that the plan now carries falsifiable criteria at the level where the mistakes actually happen — "does not skip on a healthy three-node run", "emits on the create path", "answers 409" — rather than criteria that verify the reasoning.

mikalstill:

mikalstill:
Fourth review addressed in 9f3a200e5. Seven items; six held, one was half right, and I've said which half below. Rebased onto develop first, because two of the reference-drift claims turned out to be the branch aging rather than the references being wrong when written.

**1 — the skip guard misses the case the phase declines to fix.** This is the best catch across all four rounds, and it is a gap rather than a repeat. F3 guarded on candidate *count*, but finding 5's mechanism ejects the affinity target *specifically*: inst1's node survived `sufficient_idle_cpu` and was dropped at `sufficient_idle_memory`. With two or three other candidates remaining the count guard never fires, and the inst2 assertion fails because inst1's node is not in `affinity_detail` at all — so the rewritten test would have gone on failing for exactly the mechanism F2 holds is not a scheduler defect and F7 declines to soften, presenting as an affinity bug. F2's prose already allowed for it ("or legitimately had no choice") and I never turned that clause into a check. There is now a second skip keyed on inst1's node being absent from the unforced pass's `affinity_detail` keys, with a distinct message so CI tells the two apart, plus your point about the tier topology: step 6 must record which skip is expected where, so a permanently-skipping test cannot pass for green.

**2 — a third wrong count source, inside the event I chose.** Confirmed at `scheduler.py:604,609`: `schedule have highest affinity` publishes `'candidates': preferred`, and `preferred` is the post-scoring winning tier. The key name is the whole trap, since `schedule final candidates` and `schedule forced candidates` both use `candidates` to mean a candidate list — so an implementer sent to the right event finds the wrong number under the obvious key, and for inst2 on a healthy cluster it is 1. F3, step 1 and the definition of done now name three wrong sources and state that the count is `len(affinity_detail)` and never `len(extra['candidates'])`.

**3 — F1 contradicted its own step rows.** Mine, introduced last round: F1 said step 3 was gated by nothing, step 3 said it could land once step 2 had filed its issue, and step 2 opens "once step 1 has merged" — so a live public-API 500 was transitively gated on a cluster-CI rewrite that has to be driven against sfcbr by hand. Step 3 now files and closes its own issue, step 2 no longer files it, and F1 states the reason rather than leaving the two claims in conflict.

**4 — preflight reports a constraint failure as capacity.** Correct in substance; the mechanism is slightly different from your description, and the difference makes it worse rather than better. The redirect exhausts its candidates and raises `AbortInstanceStart(self, 'Unable to find suitable node')` at `node_inst_netdesc_op.py:276` on the *first* cycle; `'Too many start attempts'` (`:166-167`) comes on a later one after three placement attempts. Both messages are capacity-shaped, which is the same fault F9 exists to prevent one path over — the create path escapes only because the instance is deleted before preflight runs. Step 4's brief now binds the exception in those `except` clauses and requires a message naming the constraint and the stage, leaving the redirect behaviour alone.

**5 — accept-time warns nobody who already uses the form.** Taken. Step 5 now records the limit explicitly and step 6 owes a discovery recipe, since F4 commits to removing the weighted form and the removal release is exactly where that gap would surface as unplanned work.

**6 — line-reference drift, and two of the four claims do not hold.** `instance.py`'s `affinity` property is at `:739` and `tags` at `:786-787` on current develop; both corrected, and both had drifted since the branch point at `5af155827` rather than being wrong when written. But `node_inst_netdesc_op.py`'s forced `find_candidates(inst, candidates=[config.NODE_UUID])` really is at `:156`, and its `except LowResourceException` really does begin at `:159` — on `5af155827`, on `21bd725d4`, and on the rebased branch. Those two stand. I have taken the symbol-plus-line suggestion for the references I touched, which is the more useful half of the item: it is what would have made D6's staleness self-correcting, and it is the durable fix for a class of drift that four rounds of review have now surfaced twice.

**7 — the equivalence precondition.** Right, and the correct condition is equal *magnitude*, not one tag per spec: positive scaling is order-preserving, so a two-tag equal-weight spec is a legitimate test the old phrasing appeared to forbid.

`pre-commit run --all-files` and `check-plan-status.py` pass.

Worth noting what changed in the shape of this round. The first three reviews all found the same defect wearing different clothes — a gate reading the wrong thing. This one found a gate that reads the right thing and is *incomplete*, which is a different and more interesting failure, and it came from asking what the phase had explicitly decided not to fix and then checking whether the test could still fail that way. That question is now written into the plan as a criterion rather than left as a property of a good reviewer.

mikalstill:

mikalstill:
Fifth review addressed in the commit above. All ten items verified against
the tree; nine taken, one taken with a correction to the suggested fix.

**1. `OverflowError` — taken, and it is the sharpest item here.** Confirmed:
`int(float('inf'))` raises `OverflowError`, and flask's
`req.get_json(force=True, silent=True)` (`external_api/base.py:129`) uses
`json.loads` defaults, which accept the bare `Infinity` literal. So
`{'affinity': {'a': Infinity}}` 500s exactly like the other three, and a
two-exception fix would have shipped behind a ticked box. What makes it
worth more than one exception name is that this codebase already knows the
trap — `base.py:308-320` refuses non-finite durations with a comment saying
in as many words that json.loads hands the bare literals through. The
knowledge existed thirty lines from the defect and did not reach it. Step 3
now widens to `(TypeError, ValueError, OverflowError)`, `Infinity` is in
the unit cases and in the DoD bullet, and the bullet records *why* it is
listed separately so the next reader cannot trim it back to three.
`NaN` confirmed fine: `int(float('nan'))` raises `ValueError`.

**5. Booleans — taken, with the compatibility break stated rather than
smuggled.** `int(True)` is 1, so `{'affinity': {'a': true}}` is accepted
today, and under F4's mapping it would quietly become
`prefer_with_tag: ['a']`. Refusing it is right, and `base.py:299-301` is
the precedent. But unlike the `Infinity` half it turns a request that
succeeds today into a 400, so step 3 and the DoD now say that in those
words, and the step 3 issue and commit message must too.

**Where I did not follow the suggestion:** replacing the coercion with
`isinstance(dv, int) and not isinstance(dv, bool)`. `int('3')` succeeds
today, so a shape test also refuses string-encoded integers — a second,
unannounced compatibility break inside a step whose whole justification is
that it is a one-line fix for a live 500 that should not wait behind
anything. Step 3 now names that alternative and says why it is declined,
so the reasoning is on the record rather than looking like an oversight.

**2. The memo — taken.** Confirmed `placements` is a local created inside
the affinity block (`scheduler.py:541`) and `_placed_instances()` is one
`Node.from_db()` plus one `Instance.from_db()` per placed instance
(`:169-185`). The part I had not thought through is the ordering: a
`require_*` stage sitting with the admission filters runs *before* the CPU,
RAM and disk stages have pruned anything, so the naive implementation does
not merely double the reads, it pays the larger of the two against the full
candidate set. Step 4 now requires the memo hoisted so filter and scorer
share one read, and carries the `database_load_budget.yaml` check that
step 5 already had for a far smaller addition. Step 5 having that
requirement while step 4 did not was inconsistent on the face of it.

**3. The third abort — taken, confirmed exact.**
`AbortInstanceStart(self, 'Requested node lacks resources')` is at
`node_inst_netdesc_op.py:169-171`, between the two F9 already named, and
reached when `inst.requested_placement` is set and the forced call at
`:156` raised. An operator who pinned a node would be told it lacks
resources when what it lacks is a matching tag. F9 and step 4 now name all
three, and note that since the exception is already bound at `:159` they
are one change, not three.

**4. Naming the right source — taken.** Fair point about the method: this
document rules out wrong turns by name, and having banned
`extra['candidates']` three times for the count it never said it is the
correct source for tier membership, which is the other thing step 1 reads.
An implementer would plausibly have rebuilt the tier from `by_affinity` or
the `affinity_detail` scores — more code, and code that can disagree with
`sorted(by_affinity, reverse=True)[0]`. F3 and step 1 now say: tier from
`extra['candidates']`, count from `affinity_detail`.

**6. Cross-tag summation — taken.** The example is right and the framing is
the useful part: `prefer_without_tag` reads as a soft veto and is a term in
a sum, so a node with three `a` and one `b` beats a node with one `a` and
no `b`. Recorded in F4 as a consequence of count-proportional scoring and
in step 6 as something to document, with the worked example.

**7. Line reference — taken, confirmed off by two.** Line 487 is
`dropped=dropped)`, the tail of the `cpu_max_per_instance` stage; the
idle-CPU block runs 489-502. Corrected. Worth noting the item is right
about why it matters: that correction exists *because* stale line numbers
made D6 hard to check.

**8 and 9 — noted, and one line taken from 8.** Thanks for tracing the
`request_id` plumbing end to end; the `mariadb.py:2558,2562`, `:5804`,
`database.proto:637,673` and empty-string-round-trip legs are ones I
asserted rather than checked. The genuinely useful part is the last
sentence: `inst1['node']` and the `affinity_detail` keys are both node
UUIDs from the `get_active_node_metrics()` keyspace, and if that were false
step 1's lookup would never match and the test would skip permanently and
silently. That is the fourth distinct way this test could have passed for
green, so it is now stated in step 1 rather than left implicit.

**10. The recipe — taken.** Correct that it was the one load-bearing
deliverable with no falsifiable criterion, in a plan that is otherwise
strict about that. Step 6 must now run it once against a real cluster and
record the output, and there is a DoD bullet: a recipe that has never been
executed is not a mitigation.

Verified: `tools/check-plan-status.py` and `pre-commit run --all-files`
both pass.

mikalstill:

mikalstill:
Sixth review addressed in the commit above. All ten items verified
against the tree: eight taken, one taken with its premise corrected, one
declined.

**1. Taken, and it is a defect I wrote into this plan last round.** F9
and step 4 both claimed the three capacity-shaped preflight aborts were
"the same one-line change, because the exception is already bound by the
`except` clause at `:159`". Confirmed wrong. The
`except LowResourceException as e:` suite is `:159-162`; `:164` onwards
is dedented back to the `try` level, so `:166-167` and `:169-171` sit
outside it, and PEP 3110 deletes the `as` target when the suite exits.
At the interpreter: `NameError: cannot access local variable 'e' where
it is not associated with a value`. Only `:276` is genuinely inside an
`except` suite.

What makes this the most valuable item in six rounds is where it would
have surfaced. The reschedule path runs only under cluster CI in the
merge queue (`coding_rules.md:341-352`), so an implementer following the
brief literally writes a `NameError` that passes the PR and fails after
merge. Both passages now prescribe a flag captured inside the suite
(`affinity_failure = isinstance(e, AffinityConstraintUnsatisfiable)`)
and tested at the two outside sites, and the "one-line fix" claim is
gone. You are right that this is the plan's own house rule about guards
sitting where the exception is raised, turned on the plan.

**2. Taken.** `_log_and_raise_on_error()` raises
`exceptions.LowResourceException` unconditionally at `scheduler.py:391`
with a message it builds itself, so "publish the dropped map through the
helper like every other filter" and "raise
`AffinityConstraintUnsatisfiable`" could not both be true of the code as
it stands — two requirements in the same plan that did not compose. The
helper now grows a defaulted `exception_class` argument, and your second
point is the one that would have bitten later: the message is built in
the helper too, so step 4 must decide *in-step* how the constraint
detail reaches the 409 body, rather than discovering it while writing
the response.

**5. Taken, and it goes further than the item says.** Correct that
hoisting removes the doubling but not the pre-prune increase — my own
argument against the naive implementation applies to the hoist, once
instead of twice. The require stage now must return before touching the
memo when no `require_*` was requested. Your observation that the
affinity loop calls `_placed_instances()` unconditionally today
(`:542-543`, no short-circuit on empty `inst.affinity`) is the useful
half: it means skipping that walk when no affinity is requested
*reduces* measured load rather than holding it flat, which is a better
outcome against the budget gate than a wash. That is now in the brief as
the adjacent win.

**3. Taken.** The master plan does still carry the superseded
lost-spreading instruction, deliberately, under a note discharging it —
so two statements in this PR described the tree incorrectly, one of them
a definition-of-done bullet that was false in the same commit that
asserted it. Both reworded to the D6 treatment: the instruction is
retained and marked discharged, and the criterion is now "no document
still tells phase 6 to rule it out *without saying, at that point, that
it is already ruled out*".

**4. Taken, both halves.** Finding 2 — whose entire subject is stale
line references — had two of its own wrong. `:487-501` is `:489-502`.
And `:611-631` is D6's citation for PR 3722's *ranking precedence*, now
the `narrowed` block at `:637-658`, not for affinity scoring at
`:529-600`; I had mapped it to the nearest plausible block rather than
to what the citing sentence actually described. Finding 2 now records
both, because that second mistake is the instructive one.

**6. The premise does not hold, but the guard is worth having, and
checking it produced something the plan needed.** `FLASK_REQUEST_ID` is
always set on the create path: `RequestID(app)` (`external_api/app.py:63`)
is WSGI middleware that does `environ["FLASK_REQUEST_ID"] = req_id` on
every request, generating a uuid4 when the client sends no
`X-Request-ID`. So the create-path events cannot carry `None`. But that
is precisely the guarantee this plan had been asserting the pairing on
without ever naming — nothing in the repo *writes* that key, every
reference is a read, so the mechanism lives in a dependency. That is now
stated. The `.get(..., 'none')` defaults you cite are defensive logging
rather than evidence of absence. I have added the truthiness guard and
the fail-not-skip fallback anyway: a null-on-both-sides join would be a
fourth route to the silent permanent skip, and the guard costs one
clause.

**8. Taken as an explicit accept, using your own observation to decline
the fix.** You are right that the skip is all-or-nothing and that
inst3's coverage is lost on every `slim-tier` run. But the reason given
for letting inst3's assertion run — that it would trivially hold — is
exactly why it must not: an assertion that cannot fail is not coverage,
and a green from a trivially-true assertion is the same false pass
finding 3 describes, one assertion over. Splitting into two methods
would recover it honestly, but each needs the three-instance setup, and
doubling a cluster-CI instance create is a real cost against a runner
fleet that is already the binding constraint on merge throughput. F3 now
records the loss and the reasoning; step 6 already has to publish which
skip is expected on which topology, which is what keeps it visible.

**7. Declined.** `mariadb.py:6139-6146` spans `get_object_events()`'s
signature — including the `limit: int = 100` default that step 1 also
cites — through the docstring sentence "Returns a list of
``EventReadRow`` ordered by timestamp descending". Both claims the step
makes are inside the cited range. The suggested replacement,
`:6215-6217`, is inside `_direct_reserve_address()`, an IPAM reservation
function.

**9 and 10.** Noted. Step 3 now requires the boolean break in the
release notes as well as the issue and commit message, since the
validator is shared across three entry points. Nothing to do for 10;
thanks for checking the arithmetic.

Four new definition-of-done bullets, each falsifiable at the level the
mistakes actually happen: no exception variable read outside its suite,
the helper's defaulted argument with call sites unchanged, no extra
`_placed_instances()` reads for a create requesting no constraint, and
the guarded pairing.

Verified: `pre-commit run --all-files` passes.

mikalstill:

mikalstill:
## Review round 7 addressed

Seventh review of this plan. The context that matters for reading the
replies: the phase is now *implemented*, on two branches stacked behind
this one, so every finding below could be checked against real code
rather than against the plan's description of it. Four of the findings
turned out to describe traps the implementation had already avoided —
which is not a reason to leave the plan saying the wrong thing, since
the plan is what a later reader consults, so those are corrected at
source too. Two were real gaps in the implementation and are now fixed.

**1 — the short-circuit's `IndexError`. Fixed in the plan; the code
already did it.** The finding is exactly right about the mechanism:
`by_affinity` is a `defaultdict(list)`, the walk is what populates it,
and `sorted(by_affinity, reverse=True)[0]` over an empty one is an
`IndexError` on the majority path. The implementation seeds
`by_affinity[0] = list(candidates)` when it skips, so downstream sees
one tier at score zero. Mutation-tested: deleting that one line fails
**36** of the scheduler suite's tests, so the guard is real and not
notional. Step 4's brief now states the contract, and there is a new
definition-of-done bullet for it. One refinement on the suggestion: the
skip is keyed on whether there are any tags to *score*, not on
`inst.affinity` being empty — a specification of nothing but `require_*`
constraints would otherwise still walk.

**2 — the event payload on the skipped path. Real gap, now closed in
three places.** The event *is* still published on that path, with
`candidates`, `highest_affinity` and `by_affinity` populated and
`affinity_detail: {}`; but nothing said so, and you are right that the
recipe step 6 owes operators goes blank on precisely the creates they
are most likely to be reading. Step 4's brief now specifies the payload,
step 6's brief requires the recipe to cover both shapes,
`docs/operator_guide/scheduler.md` documents the second shape and leads
the diagnostic checklist with "is `affinity_detail` empty?", and there
is a new unit test asserting the event is published with that exact
shape when the walk is skipped.

**3 — validator ordering. Fixed in the plan; the code already had it
right.** The binary branch discriminates on the reserved keys above the
coercion loop and `return`s, so lists never reach `int()`. Mutation-
tested by deleting that `return` so the branch falls through, which is
the precise failure the finding predicts: **3** validator tests fail.
Step 4's brief now states the ordering requirement in as many words, and
a new DoD bullet spans both steps — this is a defect that only exists in
the composition of two individually-correct steps, which is exactly the
kind a per-step criterion cannot catch.

**4 — the 409 and the abort messages were untested. Both now tested.**
This was the substantive finding of the round. There were scheduler-level
tests asserting `AffinityConstraintUnsatisfiable` is raised and what its
message says, and you are right that those cannot see which `except`
clause caught it — so the ordering that F9 calls "the entire mechanism"
had no test at all. Added:

- `ExternalApiAffinityRefusalTestCase` in `shakenfist/tests/test_external_api.py`,
  building on the existing create-path harness: an unsatisfiable
  constraint answers 409 with the constraint and stage in the body, a
  genuine capacity refusal still answers 507, and the refused instance is
  still deleted. Mutation-tested by swapping the two `except` clauses:
  **2** of them fail, and the 507 one keeps passing — which is why it
  takes the pair to pin the behaviour and not either one alone.
- Five tests in `shakenfist/tests/test_node_inst_netdesc_op_preflight.py`
  over the abort path: both guards name affinity when the flag is set,
  both still say "attempts" / "lacks resources" for a real capacity
  failure, and the redirect itself still redirects — which is the whole
  reason the exception is a subclass.

F9 and step 4's brief now say the tests are owed and why the scheduler
suite cannot carry them, with two new DoD bullets.

**5 — the bootstrapping dead end. Real, and it was undocumented.**
Confirmed against `_satisfies_hard_affinity()`: constraints match tags on
instances already placed, so the first member of a `require_with_tag`
group ejects every candidate and 409s, including the attempt to create
the instance that would satisfy it. Now a Risks entry, a step 6
documentation requirement, a DoD bullet, and a warning admonition in
`docs/user_guide/affinity.md` stating the workaround in the same place —
seed the group without the constraint, or use `prefer_with_tag`.

Taking the second half of the suggestion — distinguishing "carried by
nobody in the namespace" from "carried, but not on a surviving
candidate" — I have **declined**, and it is worth saying why rather than
leaving it silent. The scheduler only ever sees the candidate set, and by
the time affinity runs the admission filters have already pruned it. So
"no candidate carries this tag" is knowable and "nothing in the namespace
carries this tag" is not, without a query the create path does not
otherwise make. A message that asserted the second from evidence for the
first would be wrong in exactly the bootstrapping case it exists to
explain — an operator whose tagged instance sits on a node that failed
the CPU filter. Documentation covers it without a claim the code cannot
support.

**6 — initialise `affinity_failure` before the `try`.** Taken, and the
code already does it. The reasoning is recorded in F9 rather than only in
the step brief, because it is the same reasoning as the PEP 3110 trap one
paragraph above and belongs beside it: today's safety is a property of
the control flow and not of the guards, and the failure mode is one this
plan has already paid to discover once.

**7 — the fetch widening belongs to the helper's contract.** Taken, and
the code already does it: `_scheduler_events()` takes the instance uuid
and nothing else, with `event_type='audit', limit=1000` in its body.
Step 1's brief now says so, with your reason — the truncation was
reintroducible precisely because the widening lived at the call site.

**8 — one table for the four constraints.** Taken. `affinity.md` now
leads the constraint section with a four-row table carrying the name, the
hard/soft strength and the effect, under a sentence stating the
already-placed and same-namespace scope for all four at once, followed by
a pointer to the two caveats. Step 6's brief requires the table shape
rather than prose, with your reasoning about the shared suffix family
carrying two separate counterintuitive readings.

**9 — the Execution table's density. Declined, with reservations.** The
concern is fair and the table really is a wall. But the briefs-in-the-
table shape is mandated by `PLAN-TEMPLATE.md`, and the `### Step N`
sections in the sibling phase plans of this same series are *post-hoc
implementation notes*, not relocated briefs — so the suggested structure
would be a new convention rather than an existing one. It would also be
churn without a reader: the briefs have now been executed, and step 7's
close-out will add exactly those `### Step N` sections in the sibling
style, which is where a later reader gets the readable version. If the
convention should change, that is worth raising against
`PLAN-TEMPLATE.md` rather than diverging in one phase plan.

**10 — D6's line numbers will drift again.** Taken. The correction now
names every range by symbol as well as by number — `:489-502` is the
`sufficient_idle_cpu` stage inside `find_candidates()`, `:187-256` is
`_has_sufficient_cpu()` where `hard_max_cpus` is applied, and the two
comment-anchored blocks are quoted — with an instruction to re-derive
from the names when the numbers next move. Verified `hard_max_cpus` is at
`scheduler.py:218`, inside that range.

**11 and 12 — noted with thanks.** The independent confirmation of the
`Infinity`/boolean reasoning in particular is useful: that step ships a
deliberate compatibility break on three entry points and it is better for
having been checked by someone else.

`pre-commit run --all-files` passes on both branches; the scheduler,
preflight, validator and create-path API suites all pass.

mikalstill:

mikalstill:
## Review round 8 addressed

Eleven findings taken. The twelfth is answered by the implementation
rather than by an edit. As last round, the phase is implemented on the
two branches stacked behind this one, so every claim could be checked
against the code the plan produced rather than against the plan's
account of it.

**1 — the short-circuit disables weighted affinity if it lands without
the mapping. Taken; the code avoided it by accident of ordering, which
is not a defence.** The mechanism is exactly right: `scoring_tags` is
computed from the `prefer_*` lists, which are empty for a weighted
specification until the mapping fills them. I checked what shipped —
at step 4's commit (`fc8a8349c`) there is no short-circuit at all; it
arrives in step 5's commit together with the mapping, so the hazard
never existed in the tree. But nothing in the plan said it had to be
that way, and the plan is what a reader follows. Step 4's brief now
makes step 5 a stated hard prerequisite and tells an implementer to
leave the walk unconditional. There is a new DoD bullet and a new unit
test asserting a weighted spec still produces a populated
`affinity_detail` — the one assertion that fails if the two are ever
separated, and the reason they now cannot be quietly separated.

I did not take the alternative predicate ("no `prefer_*` **and** not a
weighted spec"). It is worse code written to survive an ordering
nothing needs, and it would have to be deleted by the release that
removes the weighted form.

**2 — "the unforced pass" is two passes.** Taken, and this is the one I
am most glad you caught. inst2 and inst3 are separate creates against
candidate sets that need not match, and the traced #3565 run is
precisely a case where one had collapsed and the other had not. Guards
computed from inst2 alone would leave the inst3 assertion running
against a pass that may have had one candidate — failing for exactly
the mechanism F2 holds is not a scheduler defect, which is the failure
this whole finding exists to prevent, reached through a singular noun.
F3 now states per-instance evaluation, and the DoD bullet is falsifiable
by reading the two call sites. The implementation already does it (each
assertion is passed that instance's own `events[name]`), but again,
by construction rather than by instruction.

**3 — affinity.md contradicts F2. Real, and the most valuable finding
of the round.** `docs/user_guide/affinity.md` said, in an admonition,
that Shaken Fist "filters possible candidate hypervisors based on the
affinity coefficients specified" — and this phase had added a section
saying the opposite three screens below it, so the page disagreed with
itself on its central claim. Fixed on the implementation branch: the
admonition now says weights **rank** and constraints **filter**, and
the -100..100 recommendation carries the deprecation where a reader
meets it rather than only in a section further down. Step 6's brief
names both pages and says why, and there is a DoD bullet. Your point
that the word is not simply wrong any more is the right nuance and is
recorded — the fix is to say which half does which, not to delete it.

**4 — the require stage's position. Taken, and the plan was describing
a design that was not built.** The stage shipped as the last admission
filter, after `sufficient_free_disk` and immediately before the
affinity block — which is your suggestion — so the memo moves two
statements over an already-pruned set and the pre-prune argument never
applied. The three paragraphs of mitigation are gone, replaced by the
reasoning for the position: a hard filter is order-independent for
correctness, only the stage name in the refusal event changes, so the
choice is free and the cheap end is the right one.

One thing stays: the early return when no `require_*` is requested. Not
for the reason the plan gave, but because the stage still runs ahead of
the load-shedding filters, so without it every create walks candidates
it has no constraint to test.

**5 — the third abort had no instruction, no criterion and no test.**
Taken in all three. F9 now instructs it: inside the `except` suite,
branch on the exception and raise `'No node satisfies the requested
affinity constraints: %s'` with a matching audit event. The DoD bullet
covers all three sites. Two tests added on the implementation branch,
in both directions. Your observation that the site with the *easiest*
fix was the one left unspecified is the useful part — it is inside an
`except` suite, so it needs no carried flag, and that is precisely why
an earlier draft could describe it and move on.

**6 — the adjacency fallback.** Taken as "define it", not "drop it". It
is now specified exactly: sort ascending by `timestamp`, find the
chosen `schedule inputs`, take the first affinity event strictly after
it, `self.fail()` if none — which is what the implementation does. On
reachability I think the premise is slightly off: `RequestID(app)`
guarantees the id on a *current* cluster, but `request_id` is a recent
addition to `EventReadRow` and this test runs against whatever sf-api
the CI cluster has. A mixed-version or older cluster is the reachable
path. That is now written down, so it is a documented degradation
rather than the latent flake you correctly object to.

**7 — the discovery recipe has no server-side capability.** Correct,
and I confirmed it: `get_object_metadata()` is per-object and no
`sf-ctl` subcommand exposes a query. It is specified as a client-side
N+1 loop and the cost is accepted in as many words, with the reason —
it is a one-off pre-upgrade run, not a monitoring query. Scoping a new
query capability into a medium-effort documentation step is exactly the
failure mode you name. Two traps are now in the brief because they cost
me a wrong first version: `instance show` renders metadata as Python
literals, not JSON, so a `json.loads` recipe reports a clean cluster
and finds nothing.

**8 — no cluster-CI coverage for the binary model.** Taken. Two cases
added to `cluster_ci_tests/test_scheduler.py`: a `prefer_with_tag` tier
assertion through step 1's helper with the same two skips, and a
`require_with_tag` refusal asserted as a 409. Your note that the
refusal case needs no successful create is what makes it cheap — it
also means it runs on every topology including single-node, so it is
recorded in `ci.md` that a skip there is a bug in the test. Step 5's
"decide whether test_affinity moves" is now "add the binary case
beside it", not left open.

**9 — two of my own line references are off by one.** Confirmed both
(`:603` not `:602`; `:544` not `:542-543`) and corrected. Fair hit in a
plan whose finding 2 is about exactly this.

**10 — the exception's definition site.** Taken:
`shakenfist/exceptions.py`, named in F9 with your reason — an operation
module importing the scheduler for an exception type is the outcome
worth preventing.

**11 — the break bundled into a bugfix commit. Taken, and the branch is
rebuilt.** `scheduler-reservations-phase-06-validator` is now two
commits: `Refuse malformed affinity values with a 400.` (the 500 fix,
`Fixes #3967`, changing no request that succeeds today) and `Refuse
boolean affinity values: a break.` The split is byte-exact — the second
commit restores the carved-out hunks verbatim and the resulting tree
matches the pre-split tree. The implementation branch was rebased onto
the new tip. You are right that a break riding under a subject line
saying "bugfix" is easy to wave through, and the release notes were
carrying weight the commit history was not.

**12 — the Execution table. Taken, having declined it last round.**
Being asked twice was the signal, and the measurement settled it: the
sibling phase plans in this series carry briefs of roughly 300
characters; steps 1 and 4 here were 8,743 and 11,136 — twenty times the
house norm — so this plan was not following the convention I cited in
declining, it was stretching it past breaking. Five briefs now live in
`### Step N brief` sections, with a real one-or-two-sentence brief and
a link left in each cell; every row is now 245–1690 characters. Nothing
was shortened: a word-multiset diff against the previous commit shows
the only words missing are ones this round's other findings rewrote.
The deviation from `PLAN-TEMPLATE.md` is recorded above the table so it
is not read as an oversight — and if this shape is right, it belongs in
the template rather than in one plan.

**13 — noted, thank you.**

`pre-commit run --all-files` passes on all three branches. The two new
cluster-CI cases run only on `merge_group`, as `test_affinity` does.

mikalstill:

mikalstill:
## Review round 9 addressed

Seven findings taken, one declined. The plan branch is pushed; the
implementation commit is written and verified but **not pushed yet** --
see the note at the end.

**1 — `require_*` re-runs on every reschedule. Taken, and this is the
finding of the round.** You are right on the mechanism and right that
nothing covered it. `find_candidates()` is not a create-time function,
preflight calls it on every restart, so an instance placed under
`require_with_tag` whose tagged neighbour is later deleted can be
refused when it next starts.

The plan now carries a decision, **F10**, rather than only a risk,
because "is this enforced later?" is a question with two defensible
answers and the plan should be seen to have picked one. It enforces:
a constraint which applies at create and then stops is a create-time
hint with a misleading name, and `require_without_tag` is the clearer
case — an instance which must not share a hypervisor with `batch` work
has not stopped needing that on its second boot. Suppressing the stage
for forced and preflight calls would also make the constraint depend on
*how* an instance came to be scheduled, which a user cannot observe.
Three things make it survivable and all three were already in the plan
for other reasons: the preflight redirect tries every other node first
(which is why the subclass matters), F9's abort messages name the
constraint instead of reporting a full cluster, and `require_*` is
opt-in.

**Your pointer to the admonition uncovered something older.** Both
halves of that sentence are wrong, and the second was wrong before this
phase: `Instance.affinity` reads `self.metadata` live
(`instance.py:792-793`) and metadata is read from MariaDB on each
access rather than snapshotted at create (`baseobject.py:660-666`), so
"if you change the tags or affinity of an instance after instance
creation it will not affect that instance in any way" has never been
true. What this phase changes is the consequence — a re-read could only
re-rank before, and can now refuse — which moves the sentence from
misleading to dangerous. Fixed on the implementation branch, with the
restart trade also stated in the constraint section where someone
choosing between `require_*` and `prefer_*` will meet it. New DoD
bullet.

**2 — the pipeline descriptions.** Taken in the plan. The
implementation had in fact updated both (`scheduler.md` carries the
hard affinity stage as step 6 of 10, and `ARCHITECTURE.md:170` reads
"and hard affinity constraints), scores soft affinity"), but step 6's
brief said only "put the recipes there" and guessed `ARCHITECTURE.md`
was "unlikely to need touching" — which, as you say, is a gap in this
plan's own standard rather than a new requirement. Both are now named,
with a DoD bullet: a pipeline described in two places is wrong in two
places.

**3 — a list value reaching the scorer's `int()`.** Taken as a
hardening of the brief, and I want to be precise about the status: the
implementation already avoids it. `_binary_affinity_spec()` dispatches
on the reserved keys before anything coerces, and
`map_weighted_affinity()` wraps its `int()` in
`except (TypeError, ValueError, OverflowError): continue`. So there is
no live 500. But nothing said to do either, and your reading of the
consequence is exactly right — the create path catches only
`LowResourceException` and `CandidateNodeNotFoundException`, so a
`TypeError` there is a 500 on create, which is step 3's failure class
reintroduced by the step that fixes it.

Step 4 now states **two** rules, because one is not enough. Dispatch on
shape before coercing, so a binary spec never enters the weighted path.
And have the weighted path skip a value it cannot coerce rather than
raise — everything stored before step 3 landed was never validated at
all, and an instance whose metadata predates a validation rule should
place somewhere rather than become permanently unschedulable, on create
*and* on every later reschedule. Two unit tests added, and the second
is mutation-tested: making the mapping raise instead of skipping fails
it.

**4 — no automated guard against permanent skipping.** Taken as "make
the choice visible", which is what you asked for. The suggested
automation — fail rather than skip when `get_nodes()` returns three or
more — would fail every `slim-tier` run, and `slim-tier` is expected to
skip until `PLAN-ci-cloud-sizing` lands, so it cannot be adopted today.
Both the plan and `docs/developer_guide/ci.md` now say the expectation
is documentation rather than enforcement, why, and what changes it:
once that plan lands and neither topology is expected to skip, the
guard becomes a fail on any cluster with three or more hypervisors. You
are right that leaving it implied was uncomfortable in a phase whose
stated fear is that a silent skip reads as a pass.

**5 — name the release notes file.** Taken: `docs/release_notes/v07-v08.md`,
with the instruction to name whichever is current at implementation
time rather than leaving "the release notes" to be located.

**6 — restructure the briefs as numbered lists. Declined.** The
readability argument is sound and I took its structural half last round
(the briefs are out of the table cells now). This further step I think
is wrong for a reason specific to where this plan is: the steps are
implemented, so the briefs have stopped being prompts and are now the
record of what was specified. Rewriting a record so it reads better as
a prompt loses fidelity — the retrospective passages you name ("an
earlier draft said X, which is wrong") are the most load-bearing
sentences in the document, since each one marks a defect that survived
a review round, and outside the instruction they annotate they lose the
thing they attach to. The "what happened, in order" reading is what
step 7's implementation notes are for, in the shape the sibling phase
plans already use. If briefs of this size recur on a *future* phase,
the numbered-list shape is the right answer and belongs in
`PLAN-TEMPLATE.md` rather than in one plan.

**7 — GOALS.md:38.** Taken, added to step 7's close-out. "I am not
confident that the affinity rules for instances work as expected" is
the project-level form of exactly this phase's question, and it is the
one artifact that should not still say it afterwards.

**8 — the mid-sentence parentheticals.** Taken, both. The line-reference
aside now follows the paragraph it annotates instead of splitting it
from "Ranking cannot rank a candidate that is no longer in the
candidate list", and the disposal note sits before "Three positions,
for phase 6 to choose between:" so that sentence reaches its list
intact. Also fixed a ` --- ` that should have been ` -- ` in the same
edit.

---

**On pushing the implementation branch.** The push was refused:

```
- A pull request for this branch has been added to a merge queue. Branches that
  are queued for merging cannot be updated. To modify this branch, dequeue the
  associated pull request.
```

#3971 is queued. The commit for findings 1, 3 and 4 is written, and
`pre-commit run --all-files` passes on it, but pushing needs the PR
dequeued and that is not my call to make on a merge in flight. Finding
1 is the one worth weighing here: the plan and the user guide currently
disagree about whether a hard constraint outlives the create, and the
implementation commit is what settles it.

mikalstill:

mikalstill:
## Review round 10 addressed

Eleven findings taken, one recorded rather than actioned in the way suggested. Plan changes are in the commit above (`f3b0e6f84`); the code changes are in `ad4faa73e` on the implementation branch, which the merge queue is currently blocking from pushing (see the note at the end).

**1. F10 cites the wrong lines for `Instance.affinity`** — correct, and mine. `:792-793` is inside `agent_state`; the property is at `:738-740`, which is what the Design section already said. Fixed, and `baseobject.py:660-666` tightened to `:659-667` while I was in the sentence.

**2. The `require_*` stage is told to copy only the namespace skip** — correct about the plan, and the most useful finding in this round. The brief said "exactly as the scorer already does" and then cited `:574-580`, which is one of four skips. Two things follow.

The implementation had in fact handled all four — `i is None`, `i.uuid == inst.uuid`, `i.tags or []`, namespace — so neither predicted failure occurs today. But it handled them by *restating* them, which is the thing `coding_rules.md` opens by warning against, and the review is right that a filter and a scorer which disagreed about what counts as a neighbour would place an instance somewhere its own audit trail says it should not be. `_affinity_neighbour_skip()` is now the single predicate, returning the audit record naming the skip or `None`; the scorer appends what it returns to `considered`, the require stage ignores the value and reads the tags of anything it does not reject.

The plan brief now cites the whole predicate (`:555-580`, and the four sub-ranges are `:556-561`, `:562-567`, `:568-573`, `:574-580` — the review's were each off by one), names each skip, and requires one helper rather than a restatement.

The test-coverage note attached to this was also right, and the gap was on *both* sides of the predicate. Deleting the self skip failed nothing before this round: the scorer never had a test for it either. There are now two:

- a **placed** instance carrying tag `x` can still be rescheduled under `require_without_tag: ['x']` (the reschedule case F10 describes: preflight calls `find_candidates()` on every restart, so by then the instance is one of its own node's neighbours)
- a tagged instance in **another namespace** neither satisfies `require_with_tag` nor trips `require_without_tag`

Mutation-tested: deleting the self skip now fails one test, deleting the namespace skip fails two.

**3. `forced_candidates` is `bool(candidates)`** — correct, and the consequence is real even though it is not reachable in the topology the test runs on. The redirect builds its list by excluding the current node, so on a single-node cluster that list is empty, `bool([])` is false, and a second event reads unforced. `test_affinity` skipping below three nodes is what makes the pass unique, which is a property of the node count and not of the flag.

Taken both ways. `_unforced_affinity_event()` now sorts ascending and takes the oldest — `_scheduler_events()` returns the API's newest-first order, so `inputs[0]` was in fact the *newest*, which is the wrong end of exactly this problem. F3 and the step 1 brief now say to take the oldest, and why.

**4. A fourth candidate site for finding 6** — taken, with one adjustment: it is a fourth *mechanism* at a site step 2 already names, rather than a fourth site. `find_candidates()` tests `if candidates:` (`:400`, `:441`), so an empty forced list reverts to the whole cluster including the node the redirect built the list to exclude. As the review says, that is literally a retry against an already-refused node, and it is the only one of the four whose code path demonstrably re-offers it. Step 2 now checks the trace against it and, failing a match, records it as its own observation: the redirect means "every node except this one" and silently means "every node" whenever that set comes out empty.

**5. `ARCHITECTURE.md`'s pipeline sentence is already wrong** — correct, and the insert-only edit had already landed on the implementation branch, exactly as predicted. Both `queue_state` (`:624`) and `sufficient_idle_disk` (`:634`) run after the scoring block ends at `:603`. The sentence now reads pre-filters → soft affinity → load shedding → load ranking, with one clause saying why affinity sits between the two, and there is a done criterion for it separate from the insertion bullet.

**6. The step-4 done criterion depends on a step-5 change** — correct. Reworded to gate on step 4 alone (a wash: the require stage returns before touching the memo) with a separate step-5 bullet for the reduction the short-circuit delivers, and a note that the two steps promise different things.

**7. Step 3 bundles a break into a bugfix commit** — agreed, and already done: the validator branch has been two commits since round 9, `Refuse malformed affinity values with a 400.` then `Refuse boolean affinity values: a break.` The brief said one, so the plan is now what shipped rather than what was first written.

**10. `mkdocs.yml` nav line is off by two** — correct that it was wrong, though the suggested `:456` is the number on `develop`; on this branch it is `:450`. That is the argument for the alternative you offered, so the line number is gone and the nav key is cited instead, which is what finding 2 of this plan asks for everywhere else.

**11. No done criterion asserts the namespace scoping in a test** — taken, in both directions, plus the self-skip criterion from the test-coverage note. The tests are described under finding 2 above.

**12. Numbered checklists at the end of the briefs** — taken. I declined the same suggestion last round on the grounds that the steps are implemented, so a brief is now a record rather than a prompt. Asked again in a narrower form — keep the prose, add a self-check — that reasoning does not hold: the checklist is for whoever checks the diff against the brief, and this plan's own stated risk is silent omission of one requirement out of a dozen. Steps 1 and 4 now end in nine and eleven numbered items respectively, carrying nothing the prose above them does not already argue.

**8 and 9, recorded in Future work rather than actioned.** Both are real and both are bigger than this phase.

The `value` declaration (9) is `('value', 'body', 'string', ...)` in **fourteen** places across seven modules — `instance.py`, `interface.py`, `auth.py`, `artifact.py`, `node.py`, `blob.py`, `network.py`. Correcting the instance one alone leaves seven sibling endpoints disagreeing about the same parameter, and `ARGTYPES` has a `dict` token but no any-JSON one, while metadata values genuinely are any JSON. The fix needs a token before it needs a patch. Out of scope because it is pre-existing and repo-wide, not because it is not real.

On mypy (8): I checked, and neither `scheduler.py` nor `external_api/instance.py` is in the rollout (`tox.ini:69-91`). Adding either is a change to that module entire, not to what this phase adds, and annotating four new functions inside a module nothing type-checks buys the appearance of coverage and no enforcement. Recorded with that reasoning, and with the observation that `scheduler.py` is the better of the two candidates when someone does it.

`pre-commit run --all-files` passes on both branches; the full unit suite is 4,072 tests green.

**One note on the implementation commit.** `ad4faa73e` is written and verified but not pushed: PR #3971 is in the merge queue, and `GH006` refuses updates to a queued branch. Dequeuing a merge in flight is not my call, so it waits. Nothing in it is a bug fix — the shipped filter was already correct on all four skips — so if #3971 merges first, the shared predicate, the two tests and the `ARCHITECTURE.md` correction land as a follow-up rather than being lost.

mikalstill:

Issue #3696, opened 2026-08-10
Merge CI: self-hosted runners lose communication mid-test when concurrent merge groups saturate the under-cloud


mikalstill:
## Symptom

Two cluster jobs in the same merge queue run died mid-test with no failing step and no logs at all. GitHub's check annotation on both is:

> The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

- Run: https://github.com/shakenfist/shakenfist/actions/runs/31343862518 (merge group for PR #3685, 2026-08-10)
- Job "Ansible modules (collection)", runner `sfcbr-rYNsa75PjPKVME4z`, died 01:41:37 during step 10 "Run ansible module tests"
- Job "Debian 12 cluster (collection)", runner `sfcbr-Uc2A99WRVw90tXs1`, died 01:44:43 during step 9 "Run functional tests"

Both jobs had completed "Build the smoke cluster" successfully. Every step from the test-running step onward has a `null` conclusion, and `GET /actions/jobs/<id>/logs` returns `BlobNotFound` — the runner never uploaded anything, so there is no test-level evidence to look at at all.

Three sibling jobs in the same run ("Debian 12 tier", "Ubuntu 24.04 cluster", "Guests") passed.

## Why this is systemic

PR #3685 changes OpenAPI parameter declarations and documentation. Nothing in it executes at runtime on a cluster node, and it cannot terminate a runner process.

The starvation reading is well supported by the surrounding evidence. At the time, **two merge groups were running concurrently against the same under-cloud** — runs 31343841332 and 31343862518, both formed for PR #3685 within 30 seconds of each other — each launching five nested clusters, alongside a `pull_request` run for `issue-fix-3654`. That is roughly eleven nested clusters on sfcbr at once. In the same window, three other attempts by the same PR failed with `507 sufficient_idle_cpu` scheduling errors (#3498, #3670), which is the softer form of the same shortage.

## Why the duplicate merge group is worth a look

[...]

PR #3958, opened 2026-08-29
Size the node lifecycle runner.

mikalstill:
Adds the size element to `node_lifecycle_collection`'s `runs-on`.

Like the cluster jobs that reach their runner through
`shakenfist/actions`' reusable `smoke-cluster.yml`, this one named no size,
so the conductor had nothing to match and fell back to the first entry in
`CI_SIZES` — **`xs`, one vCPU and 2048 MB**. The flavour was never chosen;
it was the alphabetically first thing in a dict.

Node lifecycle deploys a cluster with ansible from the runner. The
conductor's per-instance cost samples put its unconstrained demand at
**2537 – 2719 MB** across the 16 runs that landed on a larger runner by
accident. On `xs` it is clamped to ~1.85 GB of the 1927 MB the guest can
address, with ~112 MB minimum free across 100 runs and **no swap device**
to absorb anything it cannot reclaim in time.

The companion PR shakenfist/actions#44 carries the full measurements, the
reason the sizing recommender cannot see this (both its upsize triggers are
swap-based, and there is no swap), and the cost analysis. It also covers
`smoke_collection`, the four-way `functional_matrix_merge_collection` and
`ansible_modules_collection`, which take their runner from that reusable
workflow rather than from here.

**Merge shakenfist/actions#44 first.** `functional-tests.yml` pins
`smoke-cluster.yml@main`, so this change is inert on its own.

## Deliberately left on `xs`

`Schema ENUM widening` installs MariaDB and runs tests on the runner, but
it builds no cluster and drives no deploy, and there is no measurement
saying it needs more. `pin-indirect-dependencies` likewise. Worth noting
`release.yml:90` was already asking for `s`, so the convention existed —
these jobs just never picked it up.

## Relationship to #3696

Refs #3696. The lost-runner failures in that issue are the conductor
reaping busy runners on a flapping GitHub `offline` reading, root-caused
and fixed in shakenfist/private-ci#18; this is the secondary finding from
the same investigation. Not closing the issue — see
[the analysis comment](https://github.com/shakenfist/shakenfist/issues/3696#issuecomment-5460824570)
for what is fixed, what is still open, and why the saturation reading in
the issue title was wrong.

https://claude.ai/code/session_017hketun8nDHtGsPx4Medtf

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -525,7 +525,12 @@ jobs:
       (github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch')
       && needs.check_paths.outputs.code_changed != 'false'
     needs: [check_paths]
-    runs-on: [self-hosted, vm, debian-12]
+    # Carries a size element for the same reason smoke-cluster.yml does:
+    # without one the conductor falls back to xs (1 vCPU / 2048 MB), and
+    # this job drives a full ansible deploy from the runner. Measured
+    # demand is ~2.5-2.7 GB against 1927 MB usable and no swap. See
+    # shakenfist/shakenfist#3696.
+    runs-on: [self-hosted, vm, debian-12, s]
     # The deploy and test steps below are capped at 90 and 60 minutes;
     # this is the backstop for the setup and teardown around them, and
     # is still far short of GitHub's six hour default.


mikalstill:
The prerequisite is satisfied: **shakenfist/actions#44 merged** (as
`d4aba31` on `main`), so `smoke-cluster.yml@main` now carries
`runs-on: [self-hosted, vm, debian-12, s]` and this change is no longer
inert. Nothing else gates it — checks are green (10 success, 6 skipped)
and `mergeStateStatus` is `CLEAN`.

Two things from the review of #44 that bear on this PR:

**`s` is two vCPU and 4096 MB.** Neither PR body stated it, and
`CI_SIZES` lives in the conductor where a reader here cannot check it.
Against this job's 2719 MB ceiling that leaves about a third of the
runner free. It is also the recommender's own answer rather than a
guess — the stray `xl` landings all produce downsize recommendations,
and they all say `s`.

**This job is the high-water mark of the whole set, not a straggler.**
At 2719 MB it is above every job routed through the reusable workflow
(the ceiling there is 2714 MB, `Ansible modules`). So until this merges
the worst case in the fleet is still the one running on `xs`.

The review also asked whether the multi-node topologies sit at the top
of the measured range, and suggested deferring a single size until that
was checked. The premise was inverted: `slim-tier` is two nodes and
`slim-primary` is five (`ansible/ci-topology-*.yml`), and the samples
agree — slim-primary peaks at 2678–2714 MB while slim-tier sits at
2618 MB, below some single-node runs. That question is answered rather
than deferred; the reasoning is recorded in the workflow comment in
#44.

Also worth noting for after this merges: #44 added the rule to
`actions`' `AGENTS.md` — a `vm` `runs-on` must name a size, because an
omitted one is a silent downgrade to `xs` rather than an unconstrained
job. That is the opposite of the existing static-runner trap, and it is
the convention that would have caught this. This repository's own
`AGENTS.md` does not carry it yet.

One consequence to watch on the first run: naming a size narrows the
job to `s` runners only, so it can no longer opportunistically take an
idle larger one. If `s` capacity is contended that surfaces as a queued
job rather than a failure, and the conductor's queue-time samples are
where it would show.

PR #3970, opened 2026-08-30
Stop the sidechannel executor draining its operation

mikalstill:
Phase 5 of the agent operation deadlines plan merged as #3941 with a
third automated review round unaddressed. This is its one confirmed
defect, fixed on its own rather than folded into phase 6.

## The bug

`SideChannelExecutorJob.__init__` took its working command queue
straight from `agentop.commands`. `AgentOperation.commands` returns
the operation's own list by reference (`agentoperation.py:220`), so
`self.commands.pop(0)` in `_dispatch_next_command()`
(`main.py:1132`) mutated the operation as it dispatched.

That list reaches further than the operation. `_db_get()` referenced
`data.commands` from the `AgentOperationData` which
`mariadb.get_agent_operation()` caches, and pydantic's
`frozen=True` on that model prevents attribute assignment rather
than mutation of a list field's contents, so the pop reached the
process-wide object cache and every `AgentOperation` later built
from a hit on it.

Reproduced against the merged tree before fixing:

```
retryable before dispatch: True
retryable after dispatch:  False
cached model commands:     []
fresh from_db commands:    []
```

## Why it matters

`operation_is_retryable()` opens by refusing an empty command list,
so a single-command `get-file` -- the operation shape #3516 is
actually about -- called itself unretryable the moment its only
command went out. Both executor exit paths and the reaper took the
terminal branch instead of requeueing, which is the inverse of the
decision phase 5 exists to make. A retry which did happen would also
have redispatched with nothing left to send, since the reaper's own
`AgentOperation.from_db()` reads through the same poisoned cache
entry.

## The fix

Both ends, because each is wrong on its own terms:

- `SideChannelExecutorJob.__init__` copies the list. The executor
  owns its work queue; the operation's record of what it was asked
  to do is not that queue.
- `_db_get()` copies rather than handing every reader the cache's
  interior.

Either alone would have prevented one half. `_db_get()` alone leaves
the retry decision inverted, because the executor and
`operation_is_retryable()` read the same object; the executor fix
alone leaves the sharp edge in place for the next caller.

## Tests

Five new tests, all of which fail against the unfixed code
(verified by reverting each change in turn):

- Three in `test_daemon_sidechannel_executor.py` building a real
  executor through its real constructor. The existing dispatch tests
  hand-set `job.commands`, which is exactly why they could not see
  this -- the alias was made in `__init__`.
- Two in `test_agent_operation_retry.py` covering the cache
  containment.

Full suite: 4009 tests, 118 skipped, 0 failures. `pre-commit
run --all-files` clean.

### DIFF
--- a/docs/developer_guide/coding_rules.md
+++ b/docs/developer_guide/coding_rules.md
@@ -478,6 +478,55 @@ against a mocked engine cannot do that, and neither can a test that
 rewrites the query to suit sqlite -- register the missing SQL functions
 instead, so the statement under test is the statement that ships.
 
+## A frozen model is not a deep frozen model
+
+`mariadb.get_<type>()` returns a Pydantic model with
+`model_config = ConfigDict(frozen=True)`, and that model is the one
+entry the process-wide static object value cache holds for its whole
+TTL. Every reader in the process gets the same object. `frozen=True`
+stops attribute *assignment* on it; it does not stop mutation of the
+*contents* of a list or dict field. So `data.commands.pop(0)` succeeds,
+and it edits what every other reader on the node is about to see.
+
+Two obligations follow, and they are two different obligations:
+
+* **Copy the container you hand out.** A `_db_get()` or
+  `_static_values_to_dict()` copying `list(data.commands)` is the
+  containment boundary. Without it, every object built from one cache
+  hit shares one list, so a mutation anywhere is a mutation everywhere
+  until the entry ages out.
+* **Do not mutate an element of a container you were handed.** The copy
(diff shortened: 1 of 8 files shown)


sfconductor:

The conductor will merge this at `403862e91eed` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `403862e91eed`, bypassing the merge queue.

PR #3971, opened 2026-08-30
Add binary affinity constraints, and stop asserting co-location

mikalstill:
Phase 6 of the scheduler-reservations series, implementing
[PLAN-scheduler-reservations-phase-06-affinity.md](https://github.com/shakenfist/shakenfist/pull/3957)
(steps 1, 4, 5 and 6).

**Stacked on #3972.** That PR is a standalone fix for a live 500 and
should merge first; its single commit appears in this diff until it
does. Everything from `tests: assert what soft affinity actually
promises.` onwards belongs to this PR.

## What this does

**Affinity gains a binary value shape.** The `affinity` metadata key
accepts a second form using four reserved names, alongside the existing
tag weights:

```json
{
    "require_with_tag": ["database"],
    "require_without_tag": ["batch"],
    "prefer_with_tag": ["cache"],
    "prefer_without_tag": ["webserver"]
}
```

`require_*` are **hard**: they run as a filter stage beside the CPU,
memory and disk admission filters, and a create no candidate satisfies
fails with a **409**, naming the constraint and the stage.
`prefer_*` are **soft** and rank the survivors, contributing +/-1 per
matching co-located instance.

**The weighted form is deprecated but keeps working**, mapped onto the
binary one where the scheduler reads it -- positive to
`prefer_with_tag`, negative to `prefer_without_tag`, zero to nothing.
The magnitude is discarded, which is deliberate and documented: a weight
was a multiplier on a neighbour count the caller could not predict, so
two weights were never comparable. Ordering is unchanged wherever every
weight in a spec shares a magnitude, which covers every single-tag spec.
Setting a weighted spec records a deprecation event at accept time.

**`test_affinity` no longer asserts co-location.** Per the plan's F2,
that is a guarantee the product does not make: soft affinity is
consulted only when the scheduler has a choice, and issue #3565's traced
failure was a single-candidate run where affinity was never consulted at
all. The test now reads the scheduler's own audit events, pairs the
unforced scheduling pass by `request_id`, and **skips rather than
passes** on a degenerate run -- with two distinct messages, so
"only N candidates" and "affine node not a candidate" are separable in
CI output.

**A load reduction, not just a wash.** The affinity scorer called
`_placed_instances()` unconditionally -- one `Node.from_db()` plus one
`Instance.from_db()` per placed instance per candidate -- for every
create, including the majority which request no affinity at all. It is
now skipped when there is nothing to score, with `by_affinity[0]` seeded
so everything downstream sees what it saw before.
`database_load_budget.yaml` is unchanged: nothing added here runs on a
poll loop.

## Notes for the reviewer

- `AffinityConstraintUnsatisfiable` **subclasses**
  `LowResourceException` deliberately, so preflight's redirect keeps
  working -- another node may satisfy a constraint this one does not. The
  consequence is that the create path's `except` clauses are
  order-sensitive, and that ordering is the whole mechanism behind the
  409. `ExternalApiAffinityRefusalTestCase` pins it: reversing the two
  clauses fails two of its tests while the 507 case keeps passing.
- The three preflight aborts carry a flag out of the `except` suite
  rather than reading the exception. Python deletes the `as` target when
  the suite exits (PEP 3110), and two of the three sites are dedented
  back to the `try` level -- so reading `e` there is a `NameError` on a
  path that runs only under cluster CI, which would pass this PR and
  fail in the merge queue.
- `require_with_tag` has a **bootstrapping dead end**: the first member
  of a group cannot be created under the constraint that defines the
  group. That is the constraint working, and it is documented with its
  workaround in `docs/user_guide/affinity.md`.
- Cluster CI is the only thing that exercises the rewritten
  `test_affinity`, and it runs on `merge_group` rather than
  `pull_request`.

*(Authored with Claude Code)*

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -166,14 +166,18 @@ HTTP push; otherwise it logs to the local systemd journal. See
 
 The scheduler (`shakenfist/scheduler.py`) is in-process in each `sf-api`
 worker; there is no scheduler daemon. It filters candidate hypervisors
-against the `node_metrics` table (hard pre-filters: hypervisor role, queue
-health, CPU/RAM/disk headroom, disk bandwidth, and hard affinity
-constraints), scores soft affinity, then
-ranks by **load per schedulable thread** in coarse buckets with
+against the `node_metrics` table (hard pre-filters: hypervisor role,
+CPU/RAM/disk headroom, and hard affinity constraints), scores soft
+affinity, sheds load (queue health, disk bandwidth), then ranks by
+**load per schedulable thread** in coarse buckets with
 headroom-weighted selection so differently sized machines share work
-proportionally. These pre-filters order and prune the candidate list from
-a metrics snapshot up to a minute stale; they are not themselves the
-admission decision.
+proportionally. Affinity sits between the pre-filters and load
+shedding deliberately: a node that cannot fit the instance is never
+scored, while a busy node is still one the user asked for, so load
+shedding may narrow the winning affinity group but never moves
+placement out of it. These stages order and prune the candidate list
+from a metrics snapshot up to a minute stale; they are not themselves
+the admission decision.
 
 Admission is a separate, atomic step. Once the pipeline has picked a
 candidate, `Instance.place_instance()` makes one guarded capacity claim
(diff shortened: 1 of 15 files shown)


sfconductor:

sfconductor:

The conductor will merge this at `ad4faa73e3aa` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `ad4faa73e3aa`, bypassing the merge queue.

PR #3972, opened 2026-08-30
Refuse malformed affinity values with a 400.

### LINKED ISSUE #3967: Affinity metadata values return 500 instead of 400 for lists, dicts, null and Infinity
`_validate_instance_metadata()` coerces affinity dictionary *values*
with `int(dv)` inside a `try` that catches only `ValueError`
(`shakenfist/external_api/instance.py:1403-1406`). `int()` raises
other exceptions for perfectly ordinary JSON, and those escape the
handler and reach the caller as a **500**.

Four inner values reproduce it:

| Value | Exception `int()` raises |
|---|---|
| `{'affinity': {'first-node': ['a']}}` | `TypeError` |
| `{'affinity': {'first-node': {'x': 1}}}` | `TypeError` |
| `{'affinity': {'first-node': null}}` | `TypeError` |
| `{'affinity': {'first-node': Infinity}}` | `OverflowError` |

The `Infinity` case is reachable without a client doing anything
unusual: `req.get_json(force=True, silent=True)`
(`external_api/base.py:129`) uses `json.loads` defaults, which accept
the bare `Infinity` and `NaN` literals. The same trap is already
documented and handled thirty lines away, for durations, at
`external_api/base.py:308-320`. (`NaN` is fine —
`int(float('nan'))` raises `ValueError`.)

The level matters. A malformed **outer** `affinity` value is already
refused correctly — a list by the `isinstance(value, dict)` check at
`:1398-1400`, and `None` by `if not value` at `:1387-1388`. Only the
inner per-tag coercion leaks.

`_validate_instance_metadata()` is shared, so this is three public
entry points, not one:

- instance create (`external_api/instance.py:797`)
- `PUT /instances/<ref>/metadata/<key>` (`:1375`)
- `POST /instances/<ref>/metadata` (`:1425`)

The list-valued case is the one most likely to be hit in practice:
it is the value shape the binary affinity model proposed in
`docs/plans/PLAN-scheduler-reservations-phase-06-affinity.md` uses,
so any caller who guesses that syntax against today's server gets a
500 rather than a 400 telling them the server does not support it
yet.

### Included behaviour change

The fix also refuses **booleans**. `isinstance(True, int)` is true in
Python, so `int(True)` is 1 and `{'affinity': {'first-node': true}}`
is accepted today as a weight of one. That is accidental rather than
intended, and the binary affinity model would give `true` a meaning
nobody asked for. This turns a request that currently succeeds into a
400, on all three entry points above, and mirrors the existing
boolean refusal at `external_api/base.py:299-301`.

Numeric strings are deliberately left working: `int('3')` succeeds
today, and tightening the check to a type test would refuse them as
well. That is a separate decision and not part of a 500 fix.

*(Filed with Claude Code assistance)*

mikalstill:
_validate_instance_metadata() coerced affinity dictionary values with int(dv) inside a try that caught only ValueError. int() raises TypeError for a list, a dict or None, and OverflowError for infinity, so all four escaped the handler and reached the caller as a 500 -- from instance create and from both instance metadata endpoints, since the validator is shared by all three.

The Infinity case needs no unusual client. req.get_json(force=True, silent=True) uses json.loads defaults, which accept the bare Infinity and NaN literals, so flask hands them straight through. The same trap is already documented and handled thirty lines away for durations, at external_api/base.py:308-320. NaN needed nothing, since int(float('nan')) raises ValueError, but it is asserted anyway so a later narrowing of the except clause cannot drop it silently.

The level is deliberate. A malformed outer affinity value is already refused correctly -- a list by the isinstance(value, dict) check, and None by the falsiness check at the top of the function. Only the inner per-tag coercion leaked.

This also refuses booleans, which is a behaviour change rather than a 500 being fixed: isinstance(True, int) is true in Python, so int(True) is 1 and {'affinity': {'first-node': true}} has been accepted as a weight of one. Nobody writing true means "weight 1", and the binary affinity model in PLAN-scheduler-reservations-phase-06-affinity.md would give it a meaning nobody asked for. It is recorded in the release notes as well as here, because the shared validator makes it three public entry points. It mirrors the existing boolean refusal at external_api/base.py:299-301.

Numeric strings are left working. int('3') succeeds today, so tightening the check to a type test would refuse them too; that is a separate decision and does not belong in a fix for a 500.

The five proof-of-fix tests were mutation tested against the unfixed validator and all five fail there; the seven remaining cases are regression guards which pass either way, including one asserting that numeric strings stay accepted.

Fixes #3967

Prompt: Mikal asked for phase 6 of the scheduler reservations plan to be implemented. Steps 1 and 4 are gated on his agreement per the plan's back brief, but step 3 is explicitly ungated -- it is a live 500 on a public API which the plan deliberately separated so it would not wait behind an API surface, a new scheduler stage and a back-brief gate -- so it is landed first and on its own.

Assisted-By: Claude Code

### DIFF
--- a/docs/release_notes/v07-v08.md
+++ b/docs/release_notes/v07-v08.md
@@ -152,6 +152,11 @@
   entry -- that is, `{"affinity": {"first-node": ["a"]}}` -- previously
   escaped the validator and produced a server error on instance create
   and on both instance metadata endpoints.
+* **Behaviour change:** a boolean affinity value is now refused with a
+  400. `{"affinity": {"first-node": true}}` was previously accepted as a
+  weight of one, because Python's `int(True)` is 1. If you relied on
+  that, send `1` instead. Numeric strings such as `"3"` are still
+  accepted, unchanged.
 
 ## Supported distributions
 
--- a/docs/release_notes/v07-v08.md
+++ b/docs/release_notes/v07-v08.md
@@ -146,6 +146,12 @@
   instruction to base64 encode on the caller's behalf — check that a
   regenerated client does not double encode content you already encode
   yourself, as the API has always required.
+* Malformed values inside the reserved `affinity` instance metadata key
+  are now refused with a 400 instead of returning a 500. A list, a
+  dictionary, `null` or `Infinity` used as the *value* of an affinity
+  entry -- that is, `{"affinity": {"first-node": ["a"]}}` -- previously
+  escaped the validator and produced a server error on instance create
+  and on both instance metadata endpoints.
 
 ## Supported distributions
 
--- a/shakenfist/external_api/instance.py
+++ b/shakenfist/external_api/instance.py
@@ -1400,6 +1400,14 @@ def _validate_instance_metadata(key, value):
                 400, 'value for "affinity" key should be a valid JSON dictionary')
 
         for key_type, dv in value.items():
+            # isinstance(True, int) is true in Python, so int(True) is 1 and a
+            # JSON true would be accepted as a weight of one. Nobody writing
+            # true means "weight 1", and the binary affinity model gives it a
+            # meaning nobody asked for, so refuse it explicitly.
+            if isinstance(dv, bool):
+                return sf_api.error(
+                    400, 'affinity dictionary values should be integers, not booleans')
+
             # int() raises TypeError -- not ValueError -- for a list, a dict or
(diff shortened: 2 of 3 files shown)


sfconductor:
Merged at `5267b3f45814`, bypassing the merge queue.

PR #3973, opened 2026-08-30
Phase 10: where the pre-execution time goes

mikalstill:
Phase 10 of the queue performance plan, planned and implemented. It set
out to explain a 15-17 second latency population and ended by
withdrawing it, which is the main thing to know before reading the diff.

## What this phase was, and what it turned into

The master plan's phase 10 was titled "the 15 second dependency wait"
and pointed at #3863. That was fixed by #3916, so the phase was
re-scoped around the question phase 9 left open: given that a first
deferral now costs 0.1 s, roughly 400 of 823 first deferrals were still
sitting at 15-17 s, and nobody knew why.

They were #3863 itself, still in the sample. Phase 9's window label was
ten hours out -- this cluster stamps local time with a `Z` suffix, so a
window read off the log records reads ten hours later than the window
Loki was actually asked for -- and about 40% of it predates the
redeploy which brought #3916 in.

That is confirmed independently of the window boundaries, by the
message text. The pre-fix code passed no delay and took an `int`
default, emitting `Execution deferred for 15 seconds`; the ladder
computes a float and emits `15.0`. Counted separately:

| | integer form (pre-fix) | float form |
|---|---|---|
| in phase 9's window | **411** | 1 |
| 2026-08-26 / 08-27 | 772 / 632 | -- |
| 2026-08-28 onwards | **0** | -- |

411 against phase 9's "roughly 400", falling to zero the day after the
redeploy and staying there. The claim is withdrawn and corrected where
it was made.

## What is actually left

Over 90% of the operations waiting 15 seconds or more never deferred
once. The deep tail is queue sit, not deferral -- mostly blob transfers
on a lane saturated with their own work, which is diagnosable and
expected. One `networknode`/`background` incident is not: 12 operations
created within a second, drained serially 31 minutes later, while the
same dispatcher served 85 other operations on that queue including 10
on the background lane itself.

That one cannot be diagnosed with the events we retain, because an
operation which never defers emits nothing between creation and
execution. It is written up as a classification rather than a
measurement, and the two timestamps which would close the gap are
named. It carries forward as a successor issue.

Decision 5 of the phase plan held: no change to dispatch, concurrency,
pool sizing or fairness was made or is proposed.

## Changes

* `tools/operation-timeline.py` -- joins the defer and execution event
  streams per operation uuid and decomposes `wait_seconds` into queue
  sit, summed defer delay and residual. Pages around Loki's 5000 line
  cap and subdivides any chunk which comes back at the ceiling, because
  a silently truncated window looks exactly like a complete one.
* The defer event now carries `delay` in `extra`. The message prose is
  deliberately unchanged -- the retained history this phase measures is
  matched by that exact text.
* `node_blob_op`'s `BlobAlreadyBeingTransferred` path used
  `defer_with_backoff()` instead of a bare `defer()`, which was the last
  caller in the tree taking the flat 15 second default.

## Errors this phase made, and corrected

Recorded candidly in the Results section rather than quietly fixed:

* The phase plan's own decomposition formula was wrong. `created_at` is
  `event_ts - seconds - wait_seconds`; the execution event is emitted
  after `execute()` returns, so omitting the execution time yields
  negative intervals on long-running operations. A test pins it.
* The survey attributed the window's single 15.0 s defer to
  `node_blob_op`'s bare `defer()`. It is the ladder reaching its own
  `MAX_DEFER_DELAY` cap; all 32 such events in the retained span come
  from `node_inst_netdesc_op`, and `node_blob_op` did not defer at all
  in either window. The code change is still right, the evidence for it
  was not.
* The phase was re-scoped around a population which turned out not to
  exist.

## Verification

`pre-commit run --all-files` clean on the committed tree, all ten hooks.
The measurement was run against two 42 hour `sfcbr` windows, both
fetch-verified against `count_over_time` with no chunk at the line
ceiling. Phase 9's figures were independently reproduced (1,568 defers
and 823 first deferrals, to the digit) before its claim was withdrawn.

https://claude.ai/code/session_01XNs91691riJggzstkGi2wc


mikalstill:
Addressed in 484564884. Eight of the ten items taken, one declined on the
review's own advice, one folded into another. Each claim was checked against
the tree before acting, which changed the answer on two of them.

## Fixed

**1, ZeroDivisionError when every joined operation is incomplete.** Real, and
reproducible exactly as described. `main()` now returns 0 with an explanation
naming `--lookback-minutes`, matching the empty-window path.

**2, `clipped` compares the log clock against the Loki clock.** Correct, and
this is the same class of error the phase was written to correct, one layer
down. Worth being precise about the blast radius: it does not move any
published number. Every interval in every table is a difference of two
log-clock stamps, where the offset cancels; only a comparison against the
*window* needs the skew, and nothing in the results rests on the clipped
count. Each `Operation` now carries its own skew and `created_at_ingested`
does the comparison, rather than a median over `join.skews` -- per-event is
exact and costs nothing here.

**3, blob contention retry budget.** Agreed that 105s is too short, but the
suggested schedule was arrived at differently. `ensure_local()` discards a
partial file which has gone five minutes without progress and restarts the
transfer (`blob.py:680-693`), so the ladder wants sizing against that rule
rather than against a blob size: anything shorter gives up while the transfer
it is waiting on is still healthy. `_TRANSFER_CONTENTION_DELAYS` is ten rungs,
long enough to outlast a full stall-and-restart cycle, with that reasoning in
the comment. Bounded rather than unbounded is kept deliberately -- the cluster
maintainer re-requests under-replicated blobs every five minutes, so spinning
forever buys nothing a sweep would not.

**4, defers joined to every execution event for a uuid.** Correct, and step
10a made `node_blob_op` a third operation type with that shape, so this change
made the bug more reachable. Defers are now filtered to
`timestamp <= started_at`. The `multi_execution_uuids` note now says what is
actually happening instead of attributing it to lost events.

**5, neither node_blob_op test exercises the real defer_with_backoff.** Fair
-- the mocked tests cover the branch, not the property the change depends on.
Added a test which runs the real `defer_with_backoff` with `enqueue_work_item`
mocked and asserts the op is left `STATE_QUEUED`, so a change to that contract
fails here. Mutation-tested by deleting the `STATE_QUEUED` assignment in
`defer()`: the test fails, as it must.

**6, tool documented only under networking.** Cross-reference added from the
work-queues section of `database_internals.md`, with the passage left where it
is.

**7, `--csv` help text.** Extended to say it exports the operations whose join
is complete.

**8, quadratic membership test.** Now an id set.

**10, the delay test is partly self-referential.** Pinned against the literal
as well as against the field.

## Declined

**9, the count_over_time boundary sample.** The recommendation not to change
it is right and it is taken as such: the check is a tripwire for large losses,
and describing a delta of two as a match would suppress a real small loss to
remove a cosmetic one. The `differed by 2 of 23,179` in the plan is already
annotated as the window edge.

## Verification

Every fix which claims to prove something was mutation-tested -- each of the
four behavioural changes was reverted in turn and the corresponding test
confirmed to fail. New tests cover the all-incomplete window, the clipped
count under a non-zero skew, and a uuid with two execution events and an
interleaved defer. 4,037 unit tests pass; `pre-commit run --all-files` clean
on all ten hooks.

The phase plan's Results section records what the review changed, including
the clock bug, since that section is the record of what this phase got wrong
as well as what it found.

sfconductor:

The conductor will merge this at `484564884bac` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `484564884bac`, bypassing the merge queue.

Issue #3976, opened 2026-08-30
GetNetworkAttributes from sf-net runs ~1.0/s against a 0.248/s budget on the slim-tier topology in merge CI


mikalstill:
Found during merge CI triage of PR #3968. This is a *different* assertion in
the same test as #3975 and needs a different fix, so it is filed separately.

## Symptom

`cluster_ci_tests.test_database_tier.TestDatabaseTier.test_no_unbudgeted_fixed_rate_database_polling`
failed in the **Debian 12 tier (collection)** job (`slim-tier` topology) of
PR #3968's merge group, on the *over-budget* assertion rather than the
unbudgeted-poll one:

- Run: https://github.com/shakenfist/shakenfist/actions/runs/33300049893
- Job: `Debian 12 tier (collection) / Smoke tests (collection)` (99226389487)

```
reference = []
actual    = [(('GetNetworkAttributes', 'net'), 0.9994231306536527, 0.248, 0.996)]
: These pairs are budgeted and ran steadily well above what the model predicts
for a cluster of this shape. Do not raise the budget to make this pass: either
the load is a regression worth fixing, or the model has changed and that change
belongs in a commit which says so.
summary={"activity_qps_per_window": [84.46, 64.98, 101.757, 197.153],
         "activity_spread": 3.034, "activity_spread_required": 1.8, ...}
```

So `GetNetworkAttributes` from the `net` daemon ran at ~1.0/s against a
modelled 0.248/s — about 4x budget — and held that rate (r2 0.996) while the
rest of the suite's activity varied by 3x around it.

## Notes

- The measured rate is suspiciously close to exactly 1.0/s, which suggests a
  per-second loop in `sf-net` reading network attributes once per network,
  rather than load that scales with test activity.
- This is the `slim-tier` topology (2 hypervisors, separate network node).
  The budget model may simply have been fitted on `slim-primary` and not
  hold for this shape — but per the assertion's own message, that would be a
  deliberate model change, not a budget bump.
- The polling test passed on this same job in the merge groups for #3940
  ([run 33273340426](https://github.com/shakenfist/shakenfist/actions/runs/33273340426))
  and #3961
  ([run 33289851537](https://github.com/shakenfist/shakenfist/actions/runs/33289851537)),
  so this is newly observed rather than long-standing — though those runs may
  have skipped or not reached this assertion, so "newly *introduced*" is not
  established.
- `develop` between those two bases (`2fcfe8af..52b4a944`) touched
[...]

PR #3985, opened 2026-08-30
ci: exempt the CI headroom probe's own polling

### LINKED ISSUE #3975: test_no_unbudgeted_fixed_rate_database_polling fails in merge CI: the ci_headroom_probe is itself an unbudgeted 0.4/s (GetNodeAttributes|GetAllNodeDaemonStates|GetNodeMetrics, api) poll
Found during merge CI triage of PR #3968.

## Symptom

`cluster_ci_tests.test_database_tier.TestDatabaseTier.test_no_unbudgeted_fixed_rate_database_polling`
failed in the **Ubuntu 24.04 cluster (collection)** job of PR #3968's merge
group:

- Run: https://github.com/shakenfist/shakenfist/actions/runs/33300049893
- Job: `Ubuntu 24.04 cluster (collection) / Smoke tests (collection)` (99226389481)

```
reference = []
actual    = [(('GetAllNodeDaemonStates', 'api'), 0.39972509425618863),
 (('GetNodeAttributes', 'api'), 0.39972509425618863),
 (('GetNodeMetrics', 'api'), 0.39972509425618863)]
: These (operation, caller_daemon) pairs are not in
shakenfist/data/database_load_budget.yaml, and ran above 0.30/s at the same
rate in every measurement window without following the rest of the suite,
which is what a new fixed-rate poll looks like and what a busy test does not.
```

## The check is right: this is a real fixed-rate poll, and the CI harness is what produces it

The poller is `tools/ci_headroom_probe.py`, added by #3940 and started for
every cluster deploy by `tools/ci_headroom_launch.sh` in `shakenfist/actions`
(`smoke-cluster.yml`, step "Start the cluster headroom probe").

From the same job's log:

```
Copying /srv/ci/_work/shakenfist/shakenfist/shakenfist/tools/ci_headroom_probe.py to 10.0.0.10.
    interval 15s, cap 3900s
    series /srv/ci/traces/headroom.jsonl
```

Each sample calls `GET /nodes` (`client.get_nodes()`) and
`GET /admin/resources` (`client.get_cluster_resources()`). `GET /nodes` runs
`Node.external_view()` per node, which issues `GetNodeAttributes` and
`GetAllNodeDaemonStates`; `/admin/resources` reads `GetNodeMetrics`. So on a
cluster of N nodes at a 15s interval each of those three RPCs runs at N/15
per second.

That cluster is `primary` + `sf1`..`sf5` = 6 nodes. 6/15 = **0.4/s**, against
a measured **0.39972509425618863/s** for all three pairs. The rate is set by
configuration and not by the suite's activity, which is precisely the shape
the check is designed to catch.

## Why it only surfaced now

- `ci_headroom_launch.sh` skips silently when the probe file is absent ("that
  is expected on a component ref predating the headroom probe"), so the poll
  did not exist before #3940 merged into `develop` (`0787725`).
- The earlier merge groups for #3960 and #3961 ran against base `2fcfe8af`,
  which predates #3940. The polling test **passed** there
  ([run 33289851537](https://github.com/shakenfist/shakenfist/actions/runs/33289851537)).
- In #3940's own merge group
  ([run 33273340426](https://github.com/shakenfist/shakenfist/actions/runs/33273340426))
  the probe was running on the Ubuntu cluster job, but the test *skipped*
  there ("the tier ran at a level rate throughout the measurement"), so it
  never got to assert.
- PR CI never sees it: `pull_request` runs only the single-node
  `Smoke tests (collection)` job, where 1/15 = 0.067/s sits under the 0.30/s
  threshold. Only the merge-group cluster jobs are large enough to trip it.

So any PR entering the merge queue on a base that includes #3940 can hit
this. It is not specific to #3968, whose diff only touches
`shakenfist/network/bridged_vxlan_network.py`.

## Fix direction (needs a decision, not a mechanical patch)

`database_tier.py` already names the intended escape hatch: traffic produced
by the suite rather than the cluster belongs in `HARNESS_DRIVEN_PAIRS` in
`shakenfist/deploy/shakenfist_ci/load_budget.py`, which today holds only
`('GetObjectEvents', 'api')`.

The blunt fix — adding `('GetNodeAttributes', 'api')`,
`('GetAllNodeDaemonStates', 'api')` and `('GetNodeMetrics', 'api')` — would
work, but it permanently blinds the detector to a *genuine* future fixed-rate
poll of node state from `sf-api`, which is exactly the class of regression
this check exists to find. Worth considering instead:

- Have the probe authenticate as (or be attributed to) a distinct caller
  identity so harness traffic is separable from `api` traffic generally.
- Have the test window exclude the probe's contribution, e.g. by having
  `ci_headroom_collect.sh` publish the probe's sample count/interval so the
  expected N/interval can be subtracted.
- Stop the probe for the duration of the measurement window.

Labelled `automated-fix-attempted` at filing time to keep the autofixer off
it, because the obvious three-line patch is the one that should *not* land.

mikalstill:
Fixes #3975

## What was wrong

`tools/ci_headroom_probe.py`, added by ci-cloud-sizing phase 1 (#3940) and
started for every cluster deploy by `ci_headroom_launch.sh` in
`shakenfist/actions`, samples `GET /nodes` and `GET /admin/resources` on a
15s timer for the whole of the functional test step. Reading the roster runs
`Node.external_view()` per node, so each sample is one `GetNodeAttributes`
and one `GetAllNodeDaemonStates` per node, and `/admin/resources` is one
`GetNodeMetrics` per node.

On a cluster of N nodes that is N/15 per second for all three — paced by a
constant and indifferent to what the suite is doing, which is exactly the
signature `test_no_unbudgeted_fixed_rate_database_polling` is built to call a
new polling loop. On the six-node merge queue cluster all three read
0.3997/s against a 0.30/s ceiling and failed the build:

```
reference = []
actual    = [(('GetAllNodeDaemonStates', 'api'), 0.39972509425618863),
 (('GetNodeAttributes', 'api'), 0.39972509425618863),
 (('GetNodeMetrics', 'api'), 0.39972509425618863)]
```

It only surfaced once #3940 was on the merge base: `ci_headroom_launch.sh`
skips silently when the probe file is absent, and PR CI runs a single-node
cluster where 1/15 = 0.067/s sits under the threshold. Only the merge-group
cluster jobs are large enough to trip it, so it was reachable by any PR
entering the queue, not by the one that hit it.

## The change

The check is right, and the probe is the measuring instrument rather than the
thing measured — so the exemption goes in `HARNESS_DRIVEN_PAIRS` beside the
events reads the await helpers make, not in the budget.

Rejected alternatives, both recorded in the comment beside the entries:

- **A budget entry.** It would have to carry a `per_node` term to cover this,
  raising every real cluster's ceiling in proportion to its size to model load
  that exists nowhere outside a CI job.
- **Lengthening the probe interval** until it sits under the ceiling. That is
  hand-editing a level to make a check pass, which the budget file's own
  header forbids.

## What it costs

Said plainly in the comment, because a silent exemption reads as coverage: CI
can no longer see a new fixed-rate poll of node state made through sf-api.
That blind spot is CI's alone — none of the three pairs is budgeted for the
`api` caller, so `ShakenFistUnbudgetedDatabasePolling` goes on watching all
three at the unbudgeted ceiling on every real cluster. The pairs are also
still reported in the run detail's `harness_driven` list rather than dropped.

## Two tests hold the exemption to its justification

Phase 1 instrumentation is meant to come out once the sizing question it
answers is closed, so the exemption must not outlive the probe:

- `test_the_suite_still_probes_cluster_headroom` derives the argument from
  `tools/ci_headroom_probe.py` by AST, and fails when it stops reading those
  two endpoints, stops sampling on a timer, or goes away entirely.
- `test_the_headroom_probe_pairs_are_exempt` is the other half of the
  contract: it fails if the pairs are dropped while the probe remains.

## Verification

- `pre-commit run --all-files` passes (`tox -e py3`, `flake8`, `mypy` and the
  five custom checks), each hook confirmed with its own exit code.
- All five new assertions were mutation-tested — probe file removed, `sleep`
  removed, `get_nodes()` removed, `get_cluster_resources()` removed, and a
  pair dropped from the set. **Every one fails**; the baseline passes.
- The exact failing measurement (0.3997/s, 6 nodes, 0.30/s ceiling) replayed
  through the real `load_budget` classifier now lands in `harness_driven`,
  and `assertEqual([], unbudgeted)` passes.

## Deliberately not done

- **#3976 is untouched.** The same merge group's Debian 12 tier job failed
  the same test on a different assertion (`GetNetworkAttributes`/`net` at
  ~4x budget on `slim-tier`). That is an unrelated defect with its own fix,
  so this PR does not clear that job.
- **No change to the probe itself.** Its `/admin/resources` call constructs a
  Scheduler and refreshes node metrics, so it is not free on exactly the
  busiest clusters — worth watching against the #3772 family, but a
  behaviour change to phase 1 instrumentation is not this fix.

---

https://claude.ai/code/session_01514djiwuwDUi56vC48PWNc

### DIFF
--- a/docs/developer_guide/ci.md
+++ b/docs/developer_guide/ci.md
@@ -84,10 +84,13 @@ Phase 1 of `docs/plans/PLAN-ci-cloud-sizing.md` (see
 decisions behind it) added two data-gathering instruments to every
 functional cluster job, so that later phases can size CI's clouds from
 a distribution instead of the handful of hand-collected numbers the
-plan started from. Neither instrument gates anything -- see "Nothing
-here is a quality gate" below -- so a reader troubleshooting a CI
-failure can skip this section; it matters when reading a job's bundle
-afterwards, or when working on the sizing plan itself.
+plan started from. Neither instrument gates anything itself -- see
+"Nothing here is a quality gate" below -- but the poller's own traffic
+does interact with a check that gates, which is the one reason a
+reader troubleshooting a CI failure might need this section; see "The
+probe's traffic is exempted from the idle-load check". Otherwise it
+matters when reading a job's bundle afterwards, or when working on the
+sizing plan itself.
 
 ### Two instruments, not one
 
@@ -182,7 +185,58 @@ it prints (committed vCPU as a fraction of the admission ledger,
 against bounds of 0.35 and 0.70) is explicitly labelled PROVISIONAL:
 phase 0 set those bounds with no distribution to check them against,
 phase 2 replaces or defends them, and any enforcement is phase 5's to
-add. Nothing in CI today can fail because of this instrumentation.
+add. No verdict this instrumentation prints can fail a job.
+
+That is not the same as the instrumentation being invisible to the
+checks that do gate, which is what this section used to say and what
+issue 3975 disproved -- see below.
+
+### The probe's traffic is exempted from the idle-load check
+
+The probe polls, and `test_no_unbudgeted_fixed_rate_database_polling`
+in the functional suite exists to notice polling. Each sample's `GET
+/nodes` runs `Node.external_view()` per node, which is one
+`GetNodeAttributes` and one `GetAllNodeDaemonStates` each, and its
+`GET /admin/resources` reads node metrics per node, which is one
+`GetNodeMetrics` each. On an N node cluster at the default 15 second
+interval that is N/15 per second for all three, from the `api` caller,
+flat across windows and indifferent to what the suite is doing --
+which is exactly the signature of the server-side polling loop the
+check is built to catch. It clears the unbudgeted ceiling, `max(0.25,
+0.05N)` per second, from four nodes upwards; it was found on the six
+node merge queue cluster, where all three read 0.3997/s against
+0.30/s and failed the build (issue 3975).
+
+Those three `(operation, caller)` pairs are therefore listed in
+`HARNESS_DRIVEN_PAIRS` in
+`shakenfist/deploy/shakenfist_ci/load_budget.py`, alongside the events
+reads the suite's own await helpers make. The budget file is
+deliberately not the home for them: no deployed cluster runs the
+probe, so a budget entry would model load that exists nowhere outside
+a CI job, and it would have to be a per-node term, raising every real
+cluster's ceiling in proportion to its size.
+
+Two consequences worth knowing:
+
+* **The exemption costs coverage.** CI can no longer see a *new*
+  fixed-rate poll of node state made through sf-api, whatever its
+  rate. The blind spot is CI's alone -- none of the three pairs is
+  budgeted for the `api` caller, so the
+  `ShakenFistUnbudgetedDatabasePolling` alert, which reads its
+  exclusions from the budget file rather than from that set, still
+  watches all three at the unbudgeted ceiling on every real cluster.
+  The pairs also stay visible in a run's `harness_driven` list rather
+  than being dropped from the report.
+* **Retiring the probe means trimming the exemption.** It outlives
+  its justification otherwise, and no test can catch that on its own:
+  the launcher (`ci_headroom_launch.sh`) and the workflow steps live
+  in `shakenfist/actions`, so a decommission done there stops the
+  probe without touching anything in this repository.
+  `test_the_suite_still_probes_cluster_headroom` covers what it can --
+  it fails if the probe is deleted here, stops sampling on a timer,
+  stops reading one of the two endpoints, or loses the note in its
+  docstring that records this obligation -- but a probe that is simply
+  never launched again looks identical to a running one from here.
 
 An absent census is not zero refusals. When the Loki query failed or
 log shipping was unhealthy, the report says so explicitly rather than
--- a/docs/plans/PLAN-ci-cloud-sizing.md
+++ b/docs/plans/PLAN-ci-cloud-sizing.md
@@ -806,6 +806,22 @@ sure none of them is closed by accident:
   cloud may quieten it, may not, and either way decides nothing
   about whether the scheduler was right. It needs a disposition in
   phase 0 before phase 4.
+- **#3975** (closed) -- the phase 1 headroom probe failed a merge
+  queue build. Its per-sample `GET /nodes` and `GET
+  /admin/resources` read node state once per node, so at the default
+  15s interval it produces N/15 per second of `GetNodeAttributes`,
+  `GetAllNodeDaemonStates` and `GetNodeMetrics` from the `api`
+  caller, which clears the idle-load check's unbudgeted ceiling from
+  four nodes upwards. D15 said nothing in this phase can fail a
+  build; that was true of the phase's own verdicts and false of its
+  traffic. Fixed by exempting the three pairs in
+  `HARNESS_DRIVEN_PAIRS`, not by lengthening the interval or
+  widening the budget. **This carries an obligation into whichever
+  phase retires the probe: trim those three pairs from
+  `shakenfist/deploy/shakenfist_ci/load_budget.py` at the same
+  time.** Nothing enforces it, because the launcher and the workflow
+  steps are in `shakenfist/actions` and a decommission done there
+  leaves this repository untouched.
 - **#3882** (open) -- reconciler drift is not provable from logs.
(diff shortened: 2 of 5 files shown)


mikalstill:
Cross-reference, so this is not merged twice by accident:
[#3991](https://github.com/shakenfist/shakenfist/pull/3991) fixes the same
issue a different way. It was written during merge CI triage of
[#3970](https://github.com/shakenfist/shakenfist/pull/3970)'s merge group
(run [33346719231](https://github.com/shakenfist/shakenfist/actions/runs/33346719231),
which hit this on the Debian 12 cluster job) without knowing this PR was
already open.

Both carry `Fixes #3975` and both rewrite the same comment block in
`load_budget.py`, so they conflict and **only one should land**.

The difference is the exemption's shape. This PR adds the three pairs to
`HARNESS_DRIVEN_PAIRS`, which drops them at any rate. #3991 puts them in a
separate `HEADROOM_PROBE_PAIRS` with a computed allowance of
`N / interval x tolerance`, subtracts that, and reports the residual — so
a pair running above the probe's own rate still fails the build. That is
the alternative the issue body asked for when it argued a blanket entry
"permanently blinds the detector to a *genuine* future fixed-rate poll of
node state from sf-api".

Both unblock the merge queue; I have replayed the measurements from both
recorded occurrences (runs 33300049893 and 33346719231) through each. The
trade is simplicity here against a bounded blind spot there — 0.64/s on a
six-node cluster instead of unbounded. `GetNodeMetrics/api` is the pair
where it matters most, since real API traffic produces it too: instance
creation constructs a Scheduler and reads node metrics.

No opinion pushed on which is right — that is a judgement about how much
machinery the blind spot is worth, and it is yours. Closing whichever loses
is all that is needed.

*(Triage assisted by Claude Code)*

mikalstill:
Addressed the review in 65d98ee, which takes the fix item, the document
item and four of the five considers.

**Item 1 (fix)** -- correct, and the fix is in three places rather than
one. The comment now claims only what
`test_the_suite_still_probes_cluster_headroom` can actually see, and says
plainly that a decommission done in `shakenfist/actions` stops the probe
without touching anything here. The obligation to trim the three pairs is
written in `ci_headroom_probe.py`'s docstring (the file a retirer has
open), in `docs/developer_guide/ci.md`, and against #3975 in the master
plan's bug list. A new assertion holds the docstring note itself, so it
cannot be lost to a rewrite of the file it guards.

**Item 2 (document)** -- new "The probe's traffic is exempted from the
idle-load check" section in `ci.md`, and both "nothing can fail" claims
corrected: no verdict this instrumentation *prints* can fail a job, but
its traffic reads to the idle-load check as a poll. Same correction
applied to D15 in the plan.

**Item 3** -- taken. N/15 clears max(0.25, 0.05N) from four nodes upwards;
the message now says four and shows the arithmetic, with six noted as
where it was observed.

**Item 4** -- taken, the comment half. `Node.external_view()` and
`scheduler.get_active_node_metrics()` are named as the producers, with a
note that the other per-node reads carry api budget entries and so reach
the budgeted branch. I left the AST derivation out as more machinery than
the risk warrants, which matches your own note.

**Items 5 and 6** -- both taken. `PROBE_OPERATIONS` is bound once and both
assertions share one loop; the root walk and the sleeping-loop match are
now shared helpers used by both derivation tests.

**Item 7** -- not taken, for the reason in your own footnote. A
rate-bounded exemption needs the probe's `--interval`, which this suite
does not know because the actions-side launcher derives it from
`test_timeout_minutes`, so it would mean a second constant that can
silently disagree with the launcher.

Verified with `pre-commit run --all-files` (all hooks) and by mutation:
probe deleted, sleep removed, `get_nodes()` removed,
`get_cluster_resources()` removed, docstring obligation removed, one pair
dropped, all three dropped, and `shakenfist_ci/base.py`'s sleeps removed
to check the shared helper still guards the events derivation. Every one
fails with the right test; baseline is green.

sfconductor:

The conductor will merge this at `65d98ee83df7` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `65d98ee83df7`, bypassing the merge queue.

PR #3991, opened 2026-08-31
ci: bound the headroom probe's load exemption to what it explains

### LINKED ISSUE #3975: test_no_unbudgeted_fixed_rate_database_polling fails in merge CI: the ci_headroom_probe is itself an unbudgeted 0.4/s (GetNodeAttributes|GetAllNodeDaemonStates|GetNodeMetrics, api) poll

mikalstill:
## ⚠️ This overlaps #3985 — only one of the two should land

[#3985](https://github.com/shakenfist/shakenfist/pull/3985) is already open
against this issue and fixes the same failure a different way. Both carry
`Fixes #3975`, and both edit the same comment block in `load_budget.py`, so
they will conflict. **Pick one and close the other.** I did not know #3985
existed when this branch was written; it is offered as the alternative the
issue body asked for, not as a replacement decided on.

| | #3985 | this PR |
|---|---|---|
| Mechanism | adds the three pairs to `HARNESS_DRIVEN_PAIRS` | adds them to `HEADROOM_PROBE_PAIRS` with a computed allowance |
| Unblocks the merge queue | yes | yes |
| Blind spot afterwards | unbounded for those pairs | above ~0.64/s on a six-node cluster |
| New machinery | none | `harness_allowance()`, two constants, a summary field |

The issue body argued against the blanket form: it "permanently blinds the
detector to a *genuine* future fixed-rate poll of node state from sf-api,
which is exactly the class of regression this check exists to find", and
listed subtracting the probe's contribution as an alternative. This PR is
that alternative. #3985 is simpler, already written up, and already green —
if the wider blind spot is acceptable, it is the cheaper landing.

## What was wrong

`tools/ci_headroom_probe.py`, added by ci-cloud-sizing phase 1 (#3940) and
started for every cluster deploy by `ci_headroom_launch.sh` in
`shakenfist/actions`, samples `GET /nodes` and `GET /admin/resources` on a
15s timer for the whole functional test step. Reading the roster runs
`Node.external_view()` per node — one `GetNodeAttributes` and one
`GetAllNodeDaemonStates` each — and `/admin/resources` is one
`GetNodeMetrics` per node.

On an N-node cluster that is N/15 per second for all three, paced by a
constant and indifferent to what the suite is doing, which is exactly the
signature `test_no_unbudgeted_fixed_rate_database_polling` is built to call
a new polling loop. PR CI never sees it: its single-node cluster runs the
same probe at 1/15 = 0.067/s, under the 0.30/s ceiling. Only merge-group
cluster jobs are big enough to trip it, so it was reachable by any PR
entering the queue.

## The change

Exempt the three pairs, but as an **allowance** rather than a blanket skip:
subtract the rate the probe is known to produce and report the residual, so
a pair running above the probe's own rate still fails.

The distinction is the point. `GetNodeMetrics/api` is not only the probe —
instance creation constructs a Scheduler and reads node metrics too, and in
the merge group for #3970 that pair did not even reach the unbudgeted branch
because the suite's own activity moved it. Skipping it outright would hide a
real signal in order to silence a synthetic one.

The budget is the wrong home on the same argument as `GetObjectEvents`: no
deployed cluster runs a headroom probe, so an entry would model load that
exists nowhere outside a CI job, and it would have to be a `per_node` term
that raises every real cluster's ceiling in proportion to its size.

## What it still costs

Said plainly: a new fixed-rate poll of node state from sf-api is invisible
below the probe's own rate — 0.64/s on a six-node cluster. That floor is
CI's alone. None of the three pairs is budgeted for the `api` caller, so
`ShakenFistUnbudgetedDatabasePolling` goes on watching all three at the
unbudgeted ceiling on every real cluster.

## Verification

- `pre-commit run --all-files` passes, exit 0, all ten hooks.
- Full unit suite: 4012 tests, 3894 passed, 118 skipped, **0 failed**.
- The real measurements from **both** failing runs replayed through the real
  `load_budget` classifier now land in `harness_driven` with
  `unbudgeted == []`:
  - run 33346719231 (`pr-3970`, 6 nodes, 0.39981792260485516/s)
  - run 33300049893 (`pr-3968`, 6 nodes, 0.39972509425618863/s, all three pairs)
- Regression controls: the same pair at 1.7x, 2x and 3.5x the probe's rate is
  still **caught**; `GetNodeMetrics/queues` at 0.9/s is untouched;
  `GetObjectEvents/api` keeps its blanket exemption.
- Guards mutation-tested individually — `--interval` default changed,
  `get_nodes()` removed, the loop's `sleep` removed. Each fails; the baseline
  passes; the probe is byte-identical to HEAD afterwards.

## Known limitation

`HEADROOM_PROBE_INTERVAL_SECONDS` mirrors a default that the workflow in
`shakenfist/actions` overrides explicitly with `--interval 15`, and no test
here can see that repository. It fails safe: a workflow sampling *faster*
than this leaves a residual above the allowance, which is a visible failure
naming the measured rate rather than a silent hole.

The "probe file deleted entirely" branch in the two guards is belt-and-braces
rather than load-bearing — `test_ci_headroom_probe.py` imports the probe at
module scope, so removing it fails test *discovery* before either guard runs.

## Also in this PR

`docs/developer_guide/ci.md` claimed "Nothing in CI today can fail because of
this instrumentation." #3975 is precisely that happening, so the claim is
corrected and the probe's own load and its allowance are documented.

---

https://claude.ai/code/session_011pjfk5vQtwP1NNHY55VCgv

### DIFF
--- a/docs/developer_guide/ci.md
+++ b/docs/developer_guide/ci.md
@@ -182,7 +182,52 @@ it prints (committed vCPU as a fraction of the admission ledger,
 against bounds of 0.35 and 0.70) is explicitly labelled PROVISIONAL:
 phase 0 set those bounds with no distribution to check them against,
 phase 2 replaces or defends them, and any enforcement is phase 5's to
-add. Nothing in CI today can fail because of this instrumentation.
+add. No verdict this instrumentation *prints* can fail a build.
+
+That is not the same as the instrumentation being invisible, and
+issue #3975 is the difference. The probe is a poll against the API,
+and the API reads the database, so the probe adds database load like
+anything else does -- and the idle-load check in
+`cluster_ci_tests/test_database_tier.py` is built to notice exactly
+that shape of traffic. It duly noticed the probe and failed every
+merge-queue cluster job until the allowance below was added. An
+instrument that perturbs what another check measures is a real cost of
+having it, so weigh it before adding the next one.
+
+### The probe's own load, and the allowance for it
+
+Each sample makes two API calls, and both fan out per node:
+`GET /nodes` reads every node's attributes row and daemon-state rows,
+and `GET /admin/resources` reads every node's metrics. So on an
+N-node cluster the probe produces `GetNodeAttributes`,
+`GetAllNodeDaemonStates` and `GetNodeMetrics` from `api` at N/15 per
+second each -- 0.4/s on a six-node merge-queue cluster, against the
+0.30/s ceiling the idle-load check applies to traffic that has no
+budget entry. PR CI never sees it, because its single-node cluster
+runs the same probe at 1/15 = 0.067/s.
+
+Those three pairs are therefore exempted, in
+`HEADROOM_PROBE_PAIRS` in `shakenfist_ci/load_budget.py` -- but as an
+*allowance* rather than a blanket skip. The check subtracts what the
+probe is known to produce and reports the residual, so a pair running
+above the probe's own rate still fails the build. That distinction is
+the point: `GetNodeMetrics` from `api` is not only the probe, since
+instance creation constructs a Scheduler and reads node metrics too,
+and skipping the pair outright would have hidden a genuine regression
+in order to silence a synthetic one. The blind spot that remains is
+bounded, CI-only, and named in the comment on those constants; the
+same pairs are watched at the same ceiling on every real cluster,
+because `ShakenFistUnbudgetedDatabasePolling` takes its exclusions
+from `shakenfist/data/database_load_budget.yaml` and not from the CI
+suite.
+
+The allowance divides by the probe's sampling interval, which is a
+constant in the CI suite copied from the probe's own `--interval`
+default. `test_database_tier_harness.py` reads that default back out
+of `tools/ci_headroom_probe.py` and fails if the two drift, and
+asserts the probe still samples both endpoints on a timer -- because
+the day it stops, the allowance stops being an explanation and
+becomes a hole.
 
 An absent census is not zero refusals. When the Loki query failed or
 log shipping was unhealthy, the report says so explicitly rather than
--- a/shakenfist/deploy/shakenfist_ci/database_tier.py
+++ b/shakenfist/deploy/shakenfist_ci/database_tier.py
@@ -17,6 +17,7 @@ slim-tier topology has, so it stays in the cluster suite.
 """
 
 import json
+import math
 import time
 
 import requests
@@ -87,6 +88,21 @@ def scrape_database_counters(mesh_ip):
         except ValueError:
             continue
     return counters
+
+
+def _allowance_json(allowance):
+    """Render a harness allowance for the run summary.
+
+    math.inf is what lb.harness_allowance() returns for a pair the
+    harness drives at a rate it cannot predict. json.dumps() will write
+    that as a bare Infinity, which no strict JSON reader accepts, and
+    this summary is meant to be read by tools as well as by people. None
+    means the same thing here as it does in the allowance itself -- no
+    number to compare against -- and survives a round trip.
+    """
+    if allowance is None or allowance == math.inf:
+        return None
+    return round(allowance, 3)
 
 
 def scrape_operation_requests(mesh_ip, operation, caller_daemon):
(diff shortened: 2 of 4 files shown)


sfconductor:

The conductor will merge this at `74ab44cd2c37` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
No longer armed to merge: GitHub refused the merge. Nothing was merged, and the arming has been dropped rather than retried.

mikalstill:
Closing in favour of #3985, which merged as 422bc4956.

The two PRs disagreed about mechanism and only one could land. #3985 exempts
the three pairs in `HARNESS_DRIVEN_PAIRS`; this PR subtracted a computed
allowance instead. Everything else this branch carried either landed there or
only existed to serve the mechanism that lost:

* The `docs/developer_guide/ci.md` correction — that "Nothing in CI today can
  fail because of this instrumentation" was false — landed in #3985, more
  thoroughly than here.
* The probe-still-samples guard landed there as
  `test_the_suite_still_probes_cluster_headroom`.
* `test_headroom_probe_interval_matches_the_probe` is dead without the
  allowance: nothing divides by `--interval` under a blanket skip.
* `_allowance_json()` and the summary `allowance` field are likewise only
  meaningful when there is an allowance to report. The run summary already
  listed harness-driven pairs before either PR.

#3985 also did two things this branch did not: it named the *producer* behind
each pair, so "if a producer grows a new per-node read with no `api` budget
entry, extend this set" is written down; and it recorded the decommission
obligation in three places — the probe's own docstring, the CI headroom
section of `docs/developer_guide/ci.md`, and against #3975 in
`docs/plans/PLAN-ci-cloud-sizing.md`.

## The one thing worth carrying forward

`GetNodeMetrics/api` is not exclusively probe traffic. `Scheduler.__init__`
calls `get_active_node_metrics()` (`shakenfist/scheduler.py:183`), so every
instance creation produces that pair from the `api` caller as well. The
blanket exemption therefore hides a genuine regression class and not only the
synthetic one, and hides it at any rate rather than above the probe's own.
#3985's documentation concedes exactly this: "CI can no longer see a *new*
fixed-rate poll of node state made through sf-api, whatever its rate."

That is recorded here rather than reopened as work. The probe is phase 1
instrumentation with a written decommission obligation — see the #3975 entry
in `docs/plans/PLAN-ci-cloud-sizing.md`, which requires whoever retires the
probe to trim those three pairs at the same time — so the blind spot is
time-limited by construction. `harness_allowance()` is a fair amount of
machinery to bound a hole that is meant to be closed by deletion. If the probe
outlives phase 1, or if the exemption is still in place when something needs
to trust `GetNodeMetrics/api` again, this branch is the design to come back
to.

Issue #4001, opened 2026-09-01
Preflight redirect silently re-offers the node it excluded when the exclusion set is empty


mikalstill:
`_instance_preflight()` handles a failed placement on the current node by
rebuilding the candidate list from every node *except* the current one, and
asking the scheduler to choose from it:

```python
# shakenfist/operations/node_inst_netdesc_op.py:193-200
# Try placing on another node
try:
    candidates = []
    for node in s.metrics.keys():
        if node != config.NODE_UUID:
            candidates.append(node)

    candidates = s.find_candidates(inst, candidates=candidates)
```

`find_candidates()` decides whether it has been forced by testing that list for
truth rather than for presence:

```python
# shakenfist/scheduler.py:583
'forced_candidates': bool(candidates),

# shakenfist/scheduler.py:586
if candidates:
    ...
else:
    candidates = []
    for n in self.metrics.keys():
        candidates.append(n)
```

So an **empty** forced list is indistinguishable from no forced list at all, and
the scheduler reverts to the whole cluster -- including `config.NODE_UUID`, the
one node the redirect built its list to exclude. The redirect means "every node
except this one" and silently means "every node" whenever that set comes out
empty.

## When the set comes out empty

Whenever `s.metrics` holds only the current node. The obvious case is a
single-node cluster, where the exclusion removes the only entry. It is also
reachable on a larger cluster whenever the metrics snapshot taken by the
`Scheduler()` constructed at `node_inst_netdesc_op.py:154` happens to carry one
row.

## What it costs

Mostly diagnostic today, which is why this is a defect report and not an
incident. The re-opened pass runs the same filter stages against the same node
that has just failed them, so the usual outcome is the same
`LowResourceException` one stage later. Two things are wrong regardless:

1. **The audit trail lies.** `schedule inputs` publishes
   `forced_candidates: false` and `schedule initial candidates` publishes the
   whole cluster, so a reader reconstructing a placement cannot tell a redirect
   from a fresh decision. Those events are the primary diagnostic surface for
   scheduling, and #3565 was closed on the strength of reading them.
2. **The exclusion is not honoured.** If a metrics refresh between the opening
   forced call and the redirect changes the answer, the redirect can place the
   instance back onto the node it was redirecting away from -- the one outcome
   the code at `:195-198` exists to prevent.

## Suggested fix

Distinguish "no list" from "an empty list":

```python
if candidates is not None:
```

An empty exclusion set means "there is nowhere else to go", which is a
[...]

PR #4006, opened 2026-09-01
Close out scheduler-reservations phase 6.

mikalstill:
Phase 6's code merged as #3957, #3971 and #3972, but its step 2 (triage) and
step 7 (close-out) never ran. Step 7 edits the plan file that step 1 merges,
so it cannot run until that merge has happened, and nothing enforced the
follow-up. The result was a phase whose status columns said `Not started`
against merged work, and an issue closed with no explanation on it.

## Step 2: the 3565 disposition

**#3565 was closed on 2026-08-31 with no comment at all** -- the exact
failure this step exists to prevent. The disposition is now posted as
[issuecomment-5489559241](https://github.com/shakenfist/shakenfist/issues/3565#issuecomment-5489559241):
the traced run's candidate set had collapsed to a single node before affinity
was scored, so there was no tiebreak to lose, the half of the assertion that
passed had the same cause as the half that failed, and D6's three positions
are disposed of by F2 and F7. It closed on a test change, not a scheduler
change.

This PR carries the documentation half: `PLAN-ci-cloud-sizing` tracked the
issue as open and as a "defect to fix", and required it to be given a
disposition before its phase 4. Both references are corrected, and its worry
about a bigger cloud masking the signature is discharged -- `test_affinity`
no longer asserts co-location, and **skips** on exactly the condition a small
cloud produces, so growing the cloud can no longer turn it green by hiding
the question.

## Step 2: what the retry finding actually was

Finding 6 asked for the retry behaviour to be pinned to a call site before
anything was filed. Pinning it showed the trace is **designed behaviour**:
the create path refused all four candidates on demand alone, waived the
demand guard as designed (`external_api/instance.py:951-957`), placed on
`4f14ab0d`, and preflight re-validated that placement 18 seconds later on
`sf5` through its *opening* forced call (`node_inst_netdesc_op.py:159`). The
create path's second walk never calls `find_candidates()` again, so it cannot
be the source of the forced event either.

**#4001** is filed for the fourth mechanism instead, which is real:
`find_candidates()` tests `if candidates:` (`scheduler.py:586`), so the
preflight redirect's "every node except this one" silently becomes "every
node" whenever that set comes out empty. The issue says in as many words that
it rests on a reading of the code rather than on the trace -- a wrong issue
is worse than none.

## Step 7: the close-out

Statuses flipped in the master plan, the phase plan and `index.md` (now 9 of
14), a phase status note recording why 3565 closed on a test change, and
`GOALS.md:38` -- "I am not confident that the affinity rules for instances
work as expected" -- replaced with what is now known and the concrete
remainder, retiring the numeric weights.

## The definition of done was checked, not ticked

Thirty-eight of thirty-nine criteria are ticked against evidence gathered at
close-out. Two are worth naming:

- **`test_affinity` does not skip on a healthy three-node run.** Run
  `33452092452`, job `99684426210`, `Debian 12 cluster`: `test_affinity`,
  `test_binary_affinity_prefers_the_tagged_node` and
  `test_unsatisfiable_require_with_tag_is_refused` all pass, 102 passed and 0
  failed. The same run's stage census shows `affinity_constraints` dropping
  five candidates for "no co-located instance carries a required tag", so the
  hard stage is demonstrably firing in cluster CI rather than merely present.
- **The two `_placed_instances()` read-count criteria were measured.** A
  create requesting no affinity performs **3** reads on `98d15817d` and **0**
  on `b54f126e5`, counted through `mock.patch.object` under the existing
  mocked-mariadb harness in both trees. Step 4 asked for a wash and step 5
  for a reduction; the reduction is total.

**One criterion is left unticked on purpose**: the mutation check on negative
affinity contributions was step 1's implementation-time check, and re-running
it needs a cluster CI cycle because the assertion lives in the cluster suite.
It is recorded as unverified rather than ticked on the strength of somebody
having probably done it.


sfconductor:
Merged at `4a2dfd12dbba`, bypassing the merge queue.

Issue #3884, opened 2026-08-24
Coalescing cannot fold per-node tasks: the key is the network, not (network, node)


mikalstill:
Found while addressing review on #3880 (queue-performance phase 8), which
fixed the join that had made coalescing inert since 2026-05-26 (#3878).

## The problem

Both coalescing paths key on `COALESCIBLE_TARGET_COLUMN`, which for
`NetOp` is `network_uuid` alone:

* `mariadb._direct_find_existing_coalescible_op` (enqueue-side dedup)
* `mariadb._direct_claim_coalescible_siblings` (worker-side fold)

Neither can filter on which *queue* a sibling is on, because
`cluster_operations` has no queue column, and `NetOp`'s model has no
`node_uuid` either -- `target` is queue routing and is never persisted on
the op row. So "same network" is the finest grain the fold can express.

That is fine for tasks which do cluster-wide work on the elected network
node, but `network_ensure_mesh` does **node-local** work:
`_apply_ensure_mesh` diffs *this* host's FDB. It is enqueued per
hypervisor from three sites:

* `shakenfist/network/network.py:980` (`Network.ensure_mesh`)
* `shakenfist/daemons/network/maintain.py:673`
* `shakenfist/daemons/network/maintain.py:728`

each with `target=<node_uuid>, family='network'`.

Meanwhile `shakenfist/network/network.py:316` and
`shakenfist/external_api/instance.py:1091` enqueue the two-task list
`[network_apply_create_network_node, network_ensure_mesh]` to the
cluster-wide `networknode` queue. With `network_ensure_mesh` declared
coalescible, that survivor's fold would have matched every hypervisor's
pending single-task mesh op on the same network -- same op type, same
network, single task, state `queued` -- and marked them `complete`
without doing their work. Each of those nodes' net-workers would then hit
the terminal-state branch, drop the item, and never update its FDB.

This is the failure described in the comment at
`shakenfist/operations/baseoperation.py:335-353`, which broke
`test_single_virtual_networks_work` on the network-facade branch. The
guard added at the time only checks the *survivor's* queue name, not the
siblings', so it does not close it.

It was inert while the join was broken. #3880 turns the join on.

## What #3880 did instead

* Removed `network_ensure_mesh` from `COALESCIBLE_TASKS`.
* Added an enqueue-time guard raising
  `exceptions.InvalidCoalescibleEnqueue` when any coalescible task is
  enqueued to a non-`networknode` target, so the convention is enforced
[...]

PR #4007, opened 2026-09-01
Phase 11: fold mesh work per node, not per network

mikalstill:
Phase 11 of the queue performance plan, planned and implemented.
`network_ensure_mesh` folds again, keyed on `(network_uuid,
node_uuid)`. Phase 8 removed it because the network alone could not
tell one hypervisor's FDB work from another's, so a survivor on node
A marked node B's mesh op complete and B's FDB went stale. #3884.

## The plan was wrong twice, and both corrections matter more than the original work

Recorded here rather than quietly fixed, because either would have
been invisible in review.

**A `None` key value had to bind `IS NULL`, not skip the fold.** Key
columns are declared per operation *class*, so widening NetOp's to
`(network_uuid, node_uuid)` widens it for the cluster-wide tasks too
-- and those carry `node_uuid = None`. The first commit's preflight
refused a `None`-valued column outright, so the widening would have
silently switched coalescing off for `network_apply_update_dnsmasq`
and `network_apply_create_network_node`, the only two tasks that
actually fold today. It would have been logged. It would not have
been noticed until the phase measured a loss. `IS NULL` is also the
correct semantics: a cluster-wide operation folds only other
cluster-wide operations, a per-node one folds only its own node's.
Both are strictly narrower than the old network-only key, which is
the property the whole phase rests on.

**A node-aware key is necessary but not sufficient.** Both guards
originally tested the key but not which dispatcher would drain the
work -- while the enqueue guard's own error message already named
that as the deciding factor. `enqueue_cluster_operation` builds the
queue name as `{target}-{family}-{priority}`, so a node uuid in
`target` only reaches `sf-net` (which hashes every operation for a
network to one worker thread) when the caller also passes
`family='network'`. The default `clusteroperation` family reaches
`sf-queues`, which starts one worker per claimed item with no
routing key at all. Both guards now test the family.

## Scope: half of #3884 is deliberately not here

`NodeNetOp.network_apply_create_hypervisor` looks like the cheap
half -- its model already carries both columns, no version bump
needed. It is excluded because it is drained by `sf-queues`. The
safety argument has four links and the fourth (one worker thread per
target inside the draining process) does not exist there: two
workers on one node can hold two operations for the same `(network,
node)`, and worker A's fold can flip B's operation to `complete` in
the window between B's `if op.state.value != STATE_QUEUED: return`
and B's own transition to `executing`, for which `state_targets` has
no edge. That race has never been exercised, because the fold has
never run on a queue without partitioning. Fixing it means
partitioning the pool or making dequeue-to-executing atomic; both
belong on their own merits.

## Verification

Against a real database driving the production statements, not
mocks:

* a node A survivor folds only node A's sibling, leaving node B's
  queued;
* cluster-wide `dnsmasq` still folds its cluster-wide sibling and
  touches neither node's mesh op;
* narrowing the key back to the network alone reproduces the phase 8
  bug exactly, folding both nodes.

Twenty six unit tests, and all five required mutations break the
tests that claim to cover them -- removing the fold guard's
queue-name condition fails exactly the three tests asserting it and
nothing else.

The functional test asserts the mesh is *correct* after a fold, not
merely that a fold happened: it reads flood FDB entries off every
participant and requires each to flood to all the others, which is
what `_apply_ensure_mesh` writes and `Network.is_mesh_okay` audits.
A ping would pass on a learned entry, which ages out, and would not
notice a missing permanent one. It hard-fails when neither a
per-node fold nor a per-node dedup is seen, because coalescing that
matches nothing is indistinguishable from coalescing that is
switched off -- that is #3878, and it survived three months. It
skips, with counts, only in the narrower case where the key
demonstrably matched rows but the fold's race did not land.

## Two bugs found on the way

* `shakenfist/tests/mock_mariadb.py`'s coalescing preflight still
  refused a `None` key value, so with the widened key it reported
  "no coalescing" for every cluster-wide NetOp and later mock-based
  assertions would have been measuring nothing.
* `NetOp` has been at `current_version = 2` since 9855ef77d with no
  `_upgrade_step_1_to_2`. `baseobject.upgrade()` resolves each step
  with a bare `getattr` and no default, so a version-1 row raised
  `AttributeError` rather than the `UpgradeException` the branch
  beneath that `getattr` exists to raise. Narrow -- operations are
  hard deleted thirty seconds after going terminal -- but reachable
  during a rolling upgrade, and this phase adds a 2-to-3 step
  directly on top of the hole. Fixed, mutation tested, and a
  structural test now fails if any operation schema bumps
  `current_version` without its step.

## Rolling upgrade

The two coalescing RPCs got new V2 names rather than new fields. A
repeated field would be wire compatible and unsafe: an old
`sf-database` would ignore the extra key columns and fold on the
network alone, which is the cross-node corruption the key exists to
prevent. A V2 name answers `UNIMPLEMENTED`, and the client treats
that as coalescing being unavailable rather than falling back -- a
loud, temporary loss of an optimisation. The V1 pair is retained
unmodified for one release.

## Not done

* **Step 11h, the `sfcbr` re-measurement**, needs the merged build
  deployed there for 24 hours. `What step 11 measured` says plainly
  that nothing has been measured rather than leaving a gap which
  reads as "no effect". The pre-change baseline is recorded: over
  six hours, 581 of 1,510 NetOps were refused by the then
  single-column guard, and 573 of 919 per-node network operations
  dequeued alongside at least one sibling.
* **The successor issue** for the `sf-queues` half. Its body is
  written into the plan's Future work section ready to file; once it
  has a number, the `InvalidCoalescibleEnqueue` message in
  `net_op.py` needs it substituted.

The phase is therefore `In progress`, not `Complete`.

## Unrelated commit

`4e9a327ce` regenerates five `*_pb2_grpc.py` files whose
`GRPC_GENERATED_VERSION` was left behind by the grpcio bump in
7aa075f55. Only the version line moves. Split out so it is not
churn inside a phase commit.

https://claude.ai/code/session_01XNs91691riJggzstkGi2wc


sfconductor:

The conductor will merge this at `8feb312b60c2` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `8feb312b60c2`, bypassing the merge queue.

PR #4015, opened 2026-09-01
Tolerate either agent operation failure exception in the CI suite

mikalstill:
Unblocks [client-python#380](https://github.com/shakenfist/client-python/pull/380), whose smoke CI is red on this repository's own test suite.

## What broke

Phase 6 of the agent operation deadlines plan teaches the client's agent-operation await loops to recognise a terminal state on the *first* poll and raise a new, more precise `AgentOperationFailed`. Previously every one of those loops spun out its entire await budget on a `!= 'complete'` test and then raised `AgentCommandError`. Three places in `shakenfist_ci` were written against the old behaviour.

**The visible one.** `test_get_missing_file`, in both `smoke_ci_tests/test_agentops.py` and `guest_ci_tests/test_agentops.py`, asserts `AgentCommandError`. Under the new client, fetching a missing file raises `AgentOperationFailed` from `_await_agentop` before `await_agent_fetch`'s own loop is ever reached, so the assertion misses:

```
FAIL: shakenfist_ci.smoke_ci_tests.test_agentops.TestAgentOperations.test_get_missing_file
shakenfist_client.apiclient.AgentOperationFailed: Agent operation
cebfc8a8-da07-4a60-9f45-28f37d767d96 entered terminal state "error"
```

**The quieter one**, which nothing has hit yet. `_await_instance_ready`'s cloud-init retry loop in `base.py` catches `AgentCommandError` twice. Under the new client, a health check whose agent operation ends in `error` escapes the retry entirely rather than being retried — turning a recoverable hiccup into a test failure. Worth fixing in the same change, since it is the same break.

## Why it accepts both exceptions rather than switching

Neither repository can switch first. This repository's CI builds the client from client-python's `develop`; client-python's CI builds the server from this repository's `develop`. A test here asserting `AgentOperationFailed` fails against today's client, which has no such attribute to reference.

So the four sites share one constant in `shakenfist_ci/base.py`:

```python
AGENT_OPERATION_FAILURES = (
    apiclient.AgentCommandError,
    getattr(apiclient, 'AgentOperationFailed', apiclient.AgentCommandError))
```

The `getattr()` is the part that keeps the module importable against an old client. This merges green now and makes client-python#380 green on a re-run.

Phase 7 narrows it to a plain `AgentOperationFailed` reference once the client change has merged and the ordering constraint is gone. The phase 7 row of `PLAN-agent-operation-deadlines.md` records that cleanup, and a new paragraph in the plan's CI-suite section records the pull-forward, following the shape of the "Partly pulled forward into phase 4" note already there.

## Verification

- `testtools.assertRaises` accepts a tuple — confirmed in the installed `MatchesException` docstring ("If a tuple is given, then as with isinstance, any of the types in the tuple matching is sufficient to match"), not assumed.
- Exercised the constant's logic with a five-case harness: old client, new client, both exception types, the degenerate same-class-twice tuple that an old client produces, and that an unrelated `ValueError` still propagates rather than being swallowed. All pass.
- `pre-commit run --all-files`: all ten hooks pass.

Not verified against a real cluster — the assertion only re-runs meaningfully in smoke CI, which is what this PR exists to unblock.

## Merge order

This lands first. Then client-python#380 re-runs and goes green.

https://claude.ai/code/session_01Th2MPjs3zYoF4ieE2Wcthi


sfconductor:

The conductor will merge this at `3ce02714e06a` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `3ce02714e06a`, bypassing the merge queue.

PR #4033, opened 2026-09-02
Address the review of the multi-column coalescing key

mikalstill:
Follow-up to #4007, which merged at its original head before this review
had been addressed. The eight items are the automated review left on that
PR (review `5077013481`); the code changes here are the response to them.

| # | Review item | Response |
|---|---|---|
| 1 | Enqueue-side dedup was priority-blind, so a `user-waiting` enqueue could adopt a pending `background` op and inherit its slower lane | `_reusable_priorities()` in `net_op.py` passes an "at least as urgent" set through the proto to an `IN` filter in `mariadb.py`. Only the enqueue side needs it: the worker-side fold runs at dispatch on an already-dequeued op, so a survivor at any priority does the work immediately |
| 2 | The schema upgrade steps were never executed by a test | `test_upgrade_from_version_1` / `_2`, plus a test that every step between the declared versions exists |
| 3 | `getattr(..., None)` silently widened the key when a class mis-declared a key column | New `missing_key_columns` guard. A column present-but-`None` still binds `IS NULL` (cluster-wide ops legitimately fold each other); a column absent from the class now skips the fold and records `key_column_missing` |
| 4 | Cross-node index contention on `cluster_operations` | Declined as a code change — the shape of the contention cannot be predicted from the pre-change baseline. Folded into step 11h's measurement brief, which now reads lock waits and deadlocks rather than duration alone |
| 5 | An UNIMPLEMENTED RPC from an older database tier was recorded as a successful fold | `exceptions.CoalescingUnavailable`, caught in `baseoperation.py` and recorded as `coalescing_unavailable` |
| 6 | The upgrade-step sweep matched function names in comments and strings | Rewritten on `ast`, with `test_a_mention_is_not_a_definition` to hold it there |
| 7 | The functional test measured a fold against wall clock | `POST_DEDUP_FOLD_GRACE`, a deadline relative to the observed first dedup |
| 8 | The mock database accepted uuids the real preflight rejects | `UUID(str(value))` coercion in `mock_mariadb.py` |

Each assertion that claims to prove a fix was mutation-tested by reverting
the fix and confirming the test fails.

Rebased onto `develop` a second time, because scheduler-reservations
phase 7 has since landed its own `database.proto` changes. The generated
stubs were therefore regenerated with `tox -e genprotos` rather than
merged by hand; against `develop` the resulting `database_pb2.pyi` diff
is this branch's `priorities` field and nothing else, so the descriptor
churn in `database_pb2.py` is purely the offset shift from carrying both
sets of fields. The only other conflict was the `docs/plans/index.md`
row for this plan.

The full unit suite and `pre-commit run --all-files` both pass on the
rebased head.


sfconductor:

The conductor will merge this at `205b49b6c160` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

The conductor will merge this at `e8336b0a3e40` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

The conductor will merge this at `61d262d4cad8` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `61d262d4cad8`, bypassing the merge queue.

PR #4036, opened 2026-09-03
Declare marshmallow and PyJWT, which we import

mikalstill:
Two runtime imports that were declared only inside the generated indirect
dependency block.

| package | imported by | arrived via |
|---|---|---|
| `marshmallow` | `external_api/validation.py` (`marshmallow`, `fields`, `validate`) | `webargs` |
| `PyJWT` | `federation.py` (`jwt`, `jwt.PyJWKClient`), `external_api/base.py` (three types from `jwt.exceptions`) | `Flask-JWT-Extended` |

Neither is a dependency we have. They are dependencies somebody else has, that
we have been reading over their shoulder. The pins exist for exactly as long as
`webargs` and `Flask-JWT-Extended` keep requiring them, and
`tools/pin-indirect-dependencies.sh` removes a pin the moment nothing resolves
to it.

This is the shape that already bit us once: `oslo_concurrency` was imported by
the CI harness on an edge that existed only because `shakenfist-utilities`
declared a dependency it never used, and it would have broken at import time
the day that line was removed.

The generated copies go in the same change, the way `pbr` is already handled --
the reconciler leaves out anything declared above the marker, so the two never
fight over a name.

### This changes nothing about what gets installed

The versions are exactly what the block already pinned, so this moves two lines.
Resolving the direct dependencies before and after yields the same 87 packages,
with no symmetric difference at all.

### Still reported after this

`jsonschema` and `packaging`, imported by `tools/render-review.py` and
`tools/build-collection.py` rather than by the package. That is a weaker case
and a separate decision about whether tool scripts should pull the runtime
dependency list around, so they are deliberately left.

### Interaction with #4035

Independent of it and mergeable on its own, but both edit the generated block
within three lines of each other, so whichever lands second needs a trivial
rebase there. #4035 is blocked on a library-utilities release and a pin bump
anyway, so this one landing first is the cheaper order.

### Testing

`pre-commit run --all-files` clean. `undeclared-direct-dependency` drops both
findings.

### How this was found

By the `undeclared-direct-dependency` criterion in shakenfist/development#84.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,11 @@ dependencies = [
     "cryptography==50.0.1",               # apache2
     "python-magic==0.4.27",               # mit
     "webargs==8.7.1",                     # mit
+
+    # Imported directly by external_api/validation.py, not merely reached
+    # through webargs, so it is declared rather than left to arrive as a
+    # transitive pin webargs could drop at any release.
+    "marshmallow==4.3.1",                 # mit
     "py-cpuinfo==9.0.0",                  # mit
     "distro==1.9.0",                      # apache2
     "pbr==7.0.3",                         # apache2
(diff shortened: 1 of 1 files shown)


sfconductor:

The conductor will merge this at `49a3ccf14445` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

The conductor will merge this at `49a3ccf14445` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `49a3ccf14445`, bypassing the merge queue.

PR #4038, opened 2026-09-03
Stop importing oslo_concurrency in the CI harness

mikalstill:
The CI harness was the last thing in this repository importing
`oslo_concurrency`, for ten calls to `processutils.execute()`. What those calls
needed was `subprocess.run()` with an acceptable-exit-code list, which is now
`shakenfist_ci/process.py`.

### What this does and does not do

**It does not remove anything from an install**, and `pyproject.toml` is
untouched. The first version of this description claimed it shed twelve
packages; that was wrong, and the automated review caught it. An import has no
say in pip's resolution, and two *declared* dependencies pull the oslo stack in
regardless:

- `shakenfist-utilities` requires it. Removed on that project's develop branch
  by shakenfist/library-utilities#44, so this is waiting on a release and a pin
  bump here.
- `clingwrap` requires **and uses** it -- `clingwrap/main.py:12` imports
  `processutils` for the same reason this file existed to serve. Until
  clingwrap makes the same change, all twelve stay: `oslo.concurrency`,
  `oslo.config`, `oslo.i18n`, `oslo.utils`, `debtcollector`, `fasteners`,
  `iso8601`, `netaddr`, `pyparsing`, `rfc3986`, `stevedore`, `wrapt`. Fifteen
  percent of the dependency closure, and 34 of the 248 dependency bumps merged
  here in the year to September 2026.

So this is one of three steps, not the whole thing.

**What it buys today** is that the harness no longer depends on a package it
does not declare, which is what `undeclared-direct-dependency` in the
consistency audit asks of it.

### Behaviour preserved

The semantics the callers relied on, kept deliberately:

- `(stdout, stderr)` as text, **nothing stripped** -- callers split on newlines
  and parse the output as JSON.
- `check_exit_code` as `True` (only 0), `False` (any), an int, or a list.
- Decoding via `os.fsdecode` rather than a strict UTF-8 decode, so a stray byte
  from `ip` or `ssh` cannot lose a test to a `UnicodeDecodeError`.
- stdin is `/dev/null`. `ssh` reads stdin, and a command inheriting the test
  runner's would block rather than fail.

### The namespace key is off the command line entirely

The command line tests logged every `sf-client` invocation to stderr, key
included, on success as well as failure -- so masking it inside the exception
bought nothing at the one call site it was written for.

`sf-client` has honoured `SHAKENFIST_KEY`, `SHAKENFIST_NAMESPACE` and
`SHAKENFIST_API_URL` since v0.2.5, so `execute()` grew an `env` argument -- laid
over the environment, not replacing it, because these commands need `PATH` --
and `_exec_client()` uses it. The key never reaches an argv.

`mask_secrets()` stays as a backstop, now applied at the logging site as well
as in the exception, and widened to catch hyphenated forms like `--ssh-key`.
It is documented as a backstop rather than as the protection: name-based
redaction is the technique `coding_rules.md` records this project abandoning in
the API layer after it leaked identity tokens, and it cannot match a short
option, a positional secret, or a flag nobody thought of. A test pins that gap
so it is not mistaken for coverage.

### Node-exec commands have a timeout

`SF_CI_NODE_EXEC_TIMEOUT`, 300 seconds by default. `ssh`'s `ConnectTimeout`
bounds the TCP connect only, and a node whose storage has hung will accept a
connection and then stop responding -- so a wedged node stalled the whole job
until the runner's own timeout killed it, naming no test.
`ProcessTimeoutError` subclasses `ProcessExecutionError`, so
`_require_node_exec()` skips on a wedged node the same way it already skips on
an unreachable one.

### Also

`_log_netns` iterated a string and wrote it one character at a time. It writes
the string. Same bytes.

### Interaction with the other open PRs

#4035 has been rebased and its `oslo.concurrency` declaration commit dropped,
so there is no longer a conflict between the two. It remains blocked on a
`shakenfist-utilities` release.

### Testing

`pre-commit run --all-files` clean, unit suite included, no skips. 25 tests in
`shakenfist/tests/test_ci_process.py`, loading the module from source the way
`test_safe_headers.py` does -- `process.py` imports nothing beyond the standard
library, so the harness stays coverable from the unit suite.

New documentation in `docs/developer_guide/ci.md` ("Running commands from the
functional suite").

### DIFF
--- a/docs/developer_guide/ci.md
+++ b/docs/developer_guide/ci.md
@@ -77,6 +77,56 @@ from another plan. If you find yourself about to do something similar, check
 this section first, and check the install path itself rather than a
 description of it.
 
+## Running commands from the functional suite
+
+`shakenfist_ci/process.py` is how the functional suite shells out. It is
+`subprocess.run()` with an acceptable-exit-code list, and it replaced
+`oslo_concurrency.processutils.execute()` -- which the harness was importing
+without declaring, and which nothing else in this repository used.
+
+Three things about it are load-bearing, and are the reason it is a module
+rather than ten inline `subprocess` calls:
+
+* **Output is returned unstripped, and decoded with `os.fsdecode`.** Callers
+  split on newlines and parse JSON, so a trailing newline is data. `ip`, `ssh`
+  and the client run against real hosts and occasionally emit a byte that is
+  not UTF-8; `fsdecode` maps it to a surrogate instead of losing the test to a
+  `UnicodeDecodeError`.
+* **stdin is `/dev/null`.** `ssh` reads stdin, and a command that inherited
+  the test runner's would block rather than fail.
+* **`ProcessTimeoutError` subclasses `ProcessExecutionError`.** A node that
+  accepts an ssh connection and then wedges is exactly as untestable as one
+  that refuses it, so `_require_node_exec()` skips on both without being
+  taught the difference.
+
+### Credentials go through the environment, not argv
+
+`_exec()` in the command-line tests writes every command it runs to stderr,
+and the harness prints it again on failure. So the namespace key reaches
+`sf-client` as `SHAKENFIST_KEY` in the child's environment -- along with
+`SHAKENFIST_NAMESPACE` and `SHAKENFIST_API_URL`, which the client has honoured
+since v0.2.5 -- and never appears on a command line at all.
+
+`process.mask_secrets()` exists as well, and rewrites the value of
+`--key`-shaped flags to `***` before a command or its output goes into an
+exception. It is a backstop and not the protection. Name-based redaction is
+the technique this repository already abandoned once in the API layer, for the
+reason recorded under "Credential-carrying routes are not logged, not
+redacted" in `coding_rules.md`: it silently starts leaking the day somebody
+adds a name it has not heard of. It cannot match a short option, a positional
+secret, or a flag nobody thought of. Keep the credential out of the string.
+
+### Node-exec commands have a timeout
+
+`_node_exec()` passes `SF_CI_NODE_EXEC_TIMEOUT` (seconds, default 300) to
+every command it runs, local or over ssh. These are introspection commands --
+`ip`, `iptables`, `virsh` -- so one of them taking minutes means the node is
+wedged, which is a state a test may well have just put it in. ssh's
+`ConnectTimeout=10` bounds the TCP connect only, and does nothing for a session
+that connects and then stops responding; without this the job would stall
+until the CI runner's own timeout killed it, with no indication of which test
+was to blame.
+
 ## CI headroom instrumentation
 
 Phase 1 of `docs/plans/PLAN-ci-cloud-sizing.md` (see
--- a/shakenfist/deploy/shakenfist_ci/base.py
+++ b/shakenfist/deploy/shakenfist_ci/base.py
@@ -12,9 +12,10 @@ import sys
 import time
 
 import testtools
-from oslo_concurrency import processutils
 from prettytable import PrettyTable
 from shakenfist_client import apiclient
+
+from shakenfist_ci import process
 
 
 logging.basicConfig(level=logging.INFO, format='%(message)s')
@@ -54,6 +55,16 @@ AGENT_OPERATION_FAILURES = (
 SF_CI_SSH_USER = os.environ.get('SF_CI_SSH_USER', 'debian')
 SF_CI_SSH_KEY = os.environ.get(
     'SF_CI_SSH_KEY', os.path.expanduser('~/.ssh/id_rsa'))
+
+# How long a node-exec command may run before it is killed. These are
+# introspection commands -- ip, iptables, virsh -- so any of them taking
+# minutes means the node is wedged, which is a thing a test may well have
+# just done to it. ssh's ConnectTimeout bounds the TCP connect only: a
+# session which connects and then stops responding, because the node's
+# storage has hung, would otherwise stall the job until the CI runner's
+# own timeout killed it with no indication of which test was to blame.
+SF_CI_NODE_EXEC_TIMEOUT = int(
+    os.environ.get('SF_CI_NODE_EXEC_TIMEOUT', '300'))
 
 
 # The delete endpoint refuses a namespace which still owns a live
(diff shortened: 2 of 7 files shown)


mikalstill:
All six items addressed in `8d6bb6f4b` (amended rather than stacked -- this
repository does not squash, so the incorrect claim in item 1 would otherwise
have landed in history verbatim).

**1. The docstring's central claim is wrong.** Confirmed and fixed. I checked
the PyPI metadata: `clingwrap==2.1` requires `oslo.concurrency`, and it is not
vestigial -- `clingwrap/main.py:12` imports `processutils` for exactly the same
purpose this file was written to replace. So the twelve packages stay
regardless of this PR. The module docstring, the commit message and the PR
description now all say so, name the two blockers, and say this is one of three
steps. Good catch; the rationale would have outlived the mistake.

**2. The key is logged in clear before the command runs.** Fixed
structurally rather than by masking the log line. `sf-client` has honoured
`SHAKENFIST_KEY`, `SHAKENFIST_NAMESPACE` and `SHAKENFIST_API_URL` since v0.2.5,
so `execute()` grew an `env` argument -- laid over `os.environ` rather than
replacing it, since these commands need `PATH` -- and `_exec_client()` in both
command line test files uses it. The key never reaches an argv, so there is
nothing at the log site to mask. The masking is applied there anyway, as a
backstop for the next command line somebody builds by hand.

**3. Name-based redaction is what the coding rules say fails.** Agreed, and
with item 2 done it is no longer the protection. Three changes: the regex is
widened to `(--?[\w-]*(?:key|password|passwd|token|secret)(?:=|\s+))(\S+)` so
`--ssh-key`, `--api-key` and `--namespace-key` are caught; the comment now
cites `coding_rules.md` and states the hole plainly rather than presenting the
regex as a fix; and there is a test, `test_masking_does_not_cover_short_options`,
which asserts that `-k SECRET` passes through unmodified, so the gap is pinned
rather than mistaken for coverage. On the `.stdout`/`.stderr` mutation: kept,
because a command echoing its own argv is a real path, but the exception's
docstring now says the attributes are masked copies and that the exception is
the wrong place to read exact output from. `test_the_error_masks_the_output_streams_too`
covers it.

**4. `check_exit_code` no longer accepts a bare int.** Fixed. The
`isinstance(check_exit_code, int)` branch goes after the `is True` / `is False`
identity checks, with a comment saying why that order matters. The ambiguous
docstring sentence now reads "Any exit code which is not acceptable raises
ProcessExecutionError". Two tests: an accepted int, and an int still rejecting
other codes.

**5. No timeout.** Added. `execute()` takes `timeout`, and `_node_exec()`
passes `SF_CI_NODE_EXEC_TIMEOUT` (300s default, env-overridable) on both the
local and the ssh path -- a hung local `ip` stalls the job just as thoroughly.
Rather than letting `subprocess.TimeoutExpired` escape past handlers that
already know how to cope, expiry raises `ProcessTimeoutError`, which
**subclasses `ProcessExecutionError`**: a node that accepts an ssh connection
and then wedges is exactly as untestable as one that refuses it, so
`_require_node_exec()` skips on both without being taught the difference.
`exit_code` is `None`, and the timeout fires regardless of `check_exit_code`
since there is no exit code to accept.

**6. `_log_netns` in the commit message.** It was already in both the commit
message and the PR description ("`_log_netns` iterated a string and wrote it a
character at a time, which emits the same bytes as writing the string; it
writes the string"), and has been kept through the rewrite.

**On the missing test scenarios**, all four are now covered or answered: the
exception's masked output streams, the bare int, the masking boundary, and --
for the fourth -- the converted call sites are still functional-suite-only,
which the stdlib-only arrangement cannot change.

Test count is 25, up from 15. `pre-commit run --all-files` is green including
the unit suite, with no skips. New documentation in
`docs/developer_guide/ci.md` under "Running commands from the functional
suite", covering the module's three load-bearing behaviours, the
credentials-through-the-environment rule, and the timeout knob.

mikalstill:
### CodeQL: four alerts dismissed as false positives

The push above turned CodeQL red with "4 new alerts including 4 high severity
security vulnerabilities", all `py/clear-text-logging-sensitive-data` on the
two `_exec()` functions. Recording the reasoning here, since the dismissal
comments only live in the security tab.

The expression CodeQL names as the sensitive data is
`process.mask_secrets(cmd)` -- the redaction call itself, at columns 51-76 of
line 16. Its Python heuristic treats a call to a function whose *name* matches
a sensitive-data pattern as a **source** of secrets, and `mask_secrets`
contains "secret". So it sees a `*secret*`-named call flowing into
`sys.stderr.write` and reports clear-text logging.

The command string reaching that write cannot contain the namespace key. It is
handed to `sf-client` through `SHAKENFIST_KEY` in the child's environment and
never appears on an argv, which is the change this PR makes in response to
item 2 of the review.

The inversion is exact, and worth knowing about this rule:

> On `develop` -- where `cmd` is built as `sf-client ... --key %s ...` and
> written to stderr on **every** invocation, pass or fail -- there are **zero**
> `py/clear-text-logging-sensitive-data` alerts.

CodeQL did not flag the real leak. It flags the fix for it, and what it points
at is the redaction. Renaming the function would silence it, but the name is
accurate and the next clearest candidates (`credential`, `token`) are on the
same heuristic's list, so that trades a correct name for a round trip through
CI.

Alerts 124-127 dismissed as false positive, following the existing practice
here -- `py/insecure-protocol` in this same `cluster_ci_tests/` directory was
dismissed the same way. CodeQL is green again, and is not a required check in
any case (`Can see status`, `Can enqueue` and `Can merge` are the three).

All checks now pass.

sfconductor:

The conductor will merge this at `8d6bb6f4be0c` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

The conductor will merge this at `8d6bb6f4be0c` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `8d6bb6f4be0c`, bypassing the merge queue.

Issue #3586, opened 2026-07-31
sf-database: per-minute events COUNT(*) thrashes the buffer pool and starves the systemd watchdog when MariaDB stalls


mikalstill:
## Symptom

`sf-database.service` hit its 60s systemd watchdog, was SIGABRT-killed and core-dumped on **both** sfcbr gateways (sf-1 and sf-2) near-simultaneously (within ~150ms of each other), three times in seven days: 2026-07-28 17:33 AEST, and 2026-07-31 05:58 + 06:01 AEST. Each event auto-recovered in ~8 seconds, but while both gateways are down together the entire cluster sees `DatabaseUnavailable` — this surfaced as a previously-unseen traceback signature in nightly infra reporting.

```
2026-07-31T05:58:18.684883+10:00 sf-2 systemd[1]: sf-database.service: Watchdog timeout (limit 1min)!
2026-07-31T05:58:18.749002+10:00 sf-1 systemd[1]: sf-database.service: Watchdog timeout (limit 1min)!
2026-07-31T05:58:20.012984+10:00 sf-1 systemd[1]: sf-database.service: Main process exited, code=dumped, status=6/ABRT
```

## Root cause

The MariaDB server logged, at exactly the moment of the first kill (05:58:12 AEST = 19:58:12 UTC):

```
2026-07-30 19:58:12 439310 [Warning] InnoDB: Could not free any blocks in the buffer pool!
                                     Consider increasing innodb_buffer_pool_size.
```

followed 8 seconds later by a flood of `Aborted connection` warnings from both gateways' IPs as the SIGABRT'd daemons' connection pools died. The server also logs chronic `InnoDB: Long wait (N seconds) for double-write buffer flush` (escalating 5s→10s→15s on 2026-07-29).

Two things combine to turn that server-side stall into a gateway coredump:

1. **The events row-count gauge is an exact `COUNT(*)` full scan of a table bigger than the buffer pool.** `_direct_get_events_count()` runs every ~60s in each gateway's main loop. The events table is currently ~6.4M rows / ~3.4 GiB data + ~1 GiB indexes, while the buffer pool was 3 GiB — so every scan evicts essentially the whole pool, twice a minute across the two gateways. The gauge is simultaneously a major *cause* of the buffer-pool churn and the main *victim* of the resulting stall.

[...]

PR #4052, opened 2026-09-03
Scheduler reservations phase 7: capacity diagnostics

mikalstill:
Phase 7 of [PLAN-scheduler-reservations](docs/plans/PLAN-scheduler-reservations.md), planned and implemented. The phase plan is [PLAN-scheduler-reservations-phase-07-diagnostics.md](docs/plans/PLAN-scheduler-reservations-phase-07-diagnostics.md).

## The survey changed what this phase is

The master plan asked phase 7 to build three things D9 named: a failure-path verbose diagnostic, success-path drawdown events, and ceiling-rejection events. Surveying the tree found phases 3, 4, 4a and 6 had already built essentially all of them — each shipping the diagnostics for its own behaviour, because a guard nobody can see the reasoning of is not reviewable. The failure-path diagnostic is *deeper* than D9 asked for, and the CI triage tooling already reads the stage events (built by `PLAN-ci-cloud-sizing`, not by this plan).

What was actually missing is narrower and was not visible from the stub: **neither namespaces nor claims had a REST events endpoint**. So the claim accounting D9 calls "the signal a workflow's declared footprint needs revision" was being written where nobody could fetch it — including phase 4c, whose observation record gates phase 5. The master plan's stub and phase name are corrected at source.

The phase is renamed from "Diagnostic-mode rejection logging" to "Capacity diagnostics": there is no mode in the tree, and a default-off switch would hide the diagnostics on exactly the clusters that need them (G1).

## What changed

- **`shortfall`** on every capacity dimension detail, computed once in the shared `_capacity_dimension()` builder. The demand dimension is uncharged, so it is `used - limit` there rather than `used + requested - limit`.
- **The release reply widened.** `instance placement released` carried only a node name while its drawdown twin carried four counters. The post-release counters did not exist anywhere, so `ReleaseInstancePlacementReply` now carries them, plus `counters_node_uuid` naming the single node they describe.
- **`instance placed` gained `disk_gb`**, and both events source their amounts from the same `_capacity_claim` triple.
- **Claim exceedance recorded against the namespace** as well as the instance, with a new `claim_uuid` on the admission reply so both events name the claim. A claim's events die with `hard_delete()`, so the namespace is the durable destination (G2) — the same reasoning `namespace_claim.py` already used for the deletion event.
- **Two new events endpoints**, `/auth/namespaces/<namespace>/events` and `/auth/namespaces/<namespace>/claims/<claim_ref>/events`, both admin-gated.
- **A degraded capacity read is now visible.** `get_scheduler_node_capacity()` swallows `DatabaseUnavailable` and returns `[]` — correct and deliberate, since #3586 bounded this read to close a watchdog window — but indistinguishable from a table the reconciler has not populated. It returns a `SchedulerNodeCapacityRead` pair a caller must unpack, and the event fires only when the read actually failed.
- **A guard-refusal census** in `tools/ci_headroom_report.py`.

Observability only throughout: no guard comparison, counter arithmetic or refusal path changes. Phase 4c is mid-soak and a behavioural change would land inside the window whose data phase 5 depends on.

## Two measurements, run rather than promised

**Event volume supports G1.** The scheduling trail is **4.85%** of `events` rows over 24h on sfcbr (3.91% over 7d, 15.7 events per schedule) — an order of magnitude below the two-thirds share that made the namespace-key event storm a real problem. Fan-out is one `events` row per event and one `event_objects` row per related object, so message-prefix share is not multiplied by object count.

**A finding about the instrument, not the cluster.** `ci_headroom_collect.sh` in `shakenfist/actions` filters Loki with `schedule (at stage|has no candidates at stage)`, which selects the stage messages and nothing below them — so **no census CI collects today can contain a capacity guard refusal**, however many occurred. The census had to be rebuilt from the bundle's journals to exercise the new code. Widening that filter is in another repository, is recorded as future work, and has no issue filed yet.

That reconstruction is worth reading: **127 refusals, every one at stage `node`, every one on the `demand` dimension alone** with cpus, memory and disk inside their limits — the #3813 shape — in a run whose stage census reads clean. Directly relevant to #3772, though it does not close it.

## Worth a reviewer's attention

Step 2's post-release counter read is a `SELECT` inside the release transaction, which cost an existing assertion: `test_the_reference_lookup_runs_outside_the_release_transaction` asserted *no* `SELECT` at all, and now asserts no read *before the first write* — the invariant its admission twin already used, and the one that actually matters. The transaction still opens with a guarded `UPDATE`. **CI runs MariaDB 10.11 and is blind to `ER_CHECKREAD`**, so this was verified by reading rather than by a test failing.

Three proto changes, all additive reply fields, all documented with an "empty means not reported" convention so an older `sf-database`'s proto3 defaults read as absent rather than as a value. No schema migration.

## Notes

The plan contained three factual errors that the implementation caught, all corrected at the criterion or in the close-out: a done-criterion that assumed every dimension is charged, one asking for a subset property that was unsatisfiable until `instance placed` gained `disk_gb`, and a wrong line number for the preflight redirect (`node_inst_netdesc_op.py:159` passes exactly one node; the "every node except this one" list is built at `:200`).

All twelve done-criteria are ticked and verified. `pre-commit run --all-files` passes.


sfconductor:

The conductor will merge this at `b189689ab9aa` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `b189689ab9aa`, bypassing the merge queue.

PR #4053, opened 2026-09-03
Agent operation deadlines phase 7: docs and CI coverage

### LINKED ISSUE #3770: Guests suite stalls silently for 39 minutes: _await_agent_state deadline is renewed by any instance event, not by progress
## Symptom

The Guests job made no progress for 39 minutes and was killed by the action's
60 minute timeout. Two tests never completed, and because the step was killed
rather than failing, the job reported no failing test name at all.

- Run: https://github.com/shakenfist/shakenfist/actions/runs/31856630647
  (merge group for PR #3764, `gh-readonly-queue/develop/pr-3764-63449cad...`)
- Job: [Guests (collection) / Smoke tests (collection)](https://github.com/shakenfist/shakenfist/actions/runs/31856630647/job/94942757562),
  step "Run functional tests"

```
02:37:26  {3} ...test_networking.TestNetworking.test_provided_extra_dns [178.612570s] ... ok
          (38m 44s of complete silence across all five workers)
03:16:10  ##[error]The action 'Run functional tests' has timed out after 60 minutes.
03:16:11  'No tests in repository'      <- stestr failing had no results to report
```

## Which tests hung

Diffing the completed-test lines in the log against the test methods in
`shakenfist/deploy/shakenfist_ci/guest_ci_tests/`, 27 of 29 tests reported a
result. The two that never did are:

- `shakenfist_ci.guest_ci_tests.test_state_changes.TestStateChanges.test_lifecycle_power_cycle`
- `shakenfist_ci.guest_ci_tests.test_ubuntu.TestUbuntu.test_ubuntu_pings`

Both block on a guest agent reaching a state.

## The awaits are not bounded the way they look

`_await_agent_state` in `shakenfist/deploy/shakenfist_ci/base.py:436` reads
like it has a 500 second deadline. It does not, because the deadline is renewed
by *any* instance event rather than by progress toward the state being waited
for:

```python
time_since_last_progress = time.time()
while time.time() - time_since_last_progress < 500:
    i = self.system_client.get_instance(instance_uuid)
    ...
    if i['agent_state'] and i['agent_state'].startswith(desired):
        return

    events = self.system_client.get_instance_events(instance_uuid, limit=1)
    if events:
        last_event = events[0]
        time_since_last_progress = last_event['timestamp']   # <-- any event

    time.sleep(5)
```

`get_instance_events(..., limit=1)` returns the most recent event of any type.
So an instance that is going nowhere but still emitting events — retries,
status, agent chatter, anything the cluster does to it on a cadence shorter
than 500 seconds — holds the loop open indefinitely. The variable is named
`time_since_last_progress`, but nothing in the query restricts it to progress.

`_await_instance_create` at line 471 has the identical construction with a 180
second window, so the same applies there.

This is the only construct in the guest CI base class that can produce an
unbounded wait; `await_agent_command` in the client is capped at 120s and
`_await_instance_deleted` at 300s of wall clock.

Which specific event was doing the renewing cannot be recovered from this run —
the job was killed before the suite wrote results, so there is no per-instance
event log in the artefacts. That is itself part of the problem.

## What the post-mortem says

Nothing. Every clingwrap log check passed: no ERROR-level messages, no
`Traceback`, no `ConnectionFailedError`, no `Waiting to acquire lock`, no
`StatusCode.UNAVAILABLE`, no `WORKER TIMEOUT`. `This CI run sent 0 SIGTERM
signals while shutting down.` Whatever wedged did so quietly, which is what
distinguishes this from the existing infrastructure issues.

## Why this is systemic

PR #3764 wraps credential-carrying fields in `SecretStr`. It changes no guest
agent, power management, network or scheduler code, and both stalled tests
exercise paths the diff does not touch.

The sibling Debian 12 tier job in the same run failed separately with the
familiar 507 `sufficient_idle_cpu` signature (#3498), so the under-cloud was
under load at the time.

## Relationship to existing issues

- #3696 is the closest neighbour but is a *different* symptom: there the runner
  lost communication entirely and uploaded no logs. Here the runner was healthy
  throughout, uploaded full logs, and ran the whole post-mortem playbook
  afterwards. It was the test suite that stopped, not the runner.
- #3516 (sf-sidechannel orphans agent operations in 'executing') is a plausible
  mechanism given both stalled tests await an agent, but nothing in this run
  confirms an orphaned operation.

## Suggested fix

Separate the two clocks in `_await_agent_state` and `_await_instance_create`:
keep the event-renewed progress window, but add an absolute ceiling on the
whole await so a quietly wedged instance fails its own test with a captured
console dump instead of consuming the job's entire 60 minute budget and
destroying the evidence. Restricting the renewal to events that actually
represent progress would be the more precise fix, but the absolute ceiling is
what turns future occurrences into diagnosable failures.

Found during merge CI triage.

*(Filed with Claude Code assistance)*

mikalstill:
Phase 7 of `docs/plans/PLAN-agent-operation-deadlines.md`: documentation
and CI coverage, and the last implementation phase before the push audit
(phase 8).

The phase plan is
`docs/plans/PLAN-agent-operation-deadlines-phase-07-docs-and-ci.md`,
committed as the first commit here.

## What the survey found

The master plan's row for phase 7 was written before phases 3 to 6
executed and was wrong in four places. All four are corrected at source
in the planning commit.

1. **The documentation was not deferred from phase 3.** Phase 3's
   `cf094551b` wrote both the API reference and the release-note section
   on the two parameters. What was actually missing: phase 5's retry and
   reaper (published in exactly one place, `state_machine.md`), phase 6's
   client changes, and an operator-guide page for agent operations, of
   which there was none.
2. **`AGENT_OPERATION_FAILURES` must not narrow to a plain reference.**
   The row said to. Following it literally is a regression -- see below.
3. **There is a third event-renewed await.** `_await_objects_ready`
   backs the network, artifact and blob readiness helpers, so #3770 is
   not fixed without it. The row's two addresses had also drifted ~86
   lines.
4. **Commit `a0cc243ad` half-landed.** It fixed the shared-clock flake
   (~9% of merge-queue runs at the time) in the guest copy of
   `test_agentops.py` and never in the near-identical smoke copy -- the
   suite that runs on every pull request. That is fixed here.

## The commits

| Step | Commit | What |
|---|---|---|
| plan | `69ecfe8fe` | The phase plan, plus the four corrections above |
| 7a | `ebbf5c0da` | One clock per agent operation, and fail fast on a terminal state |
| 7b | `e11869867` | The exception tuple the client can actually raise |
| 7c | `b4bc82551` | Absolute ceilings on the three event-renewed awaits (`Fixes #3770`) |
| 7d | `42d733a3a` | Four functional tests for the timing budgets |
| 7e | `55fee2ef6` | Operator guide page, and the release note extended |
| 7f | `ea7ffe4b2` | Close-out |

## Three things a reviewer should look at

**A latent regression from phase 6, repaired by 7b.** Phase 6 changed
`await_agent_command()`'s timeout path in client-python from raising
`AgentCommandError` to raising the new `AgentAwaitTimeout`.
`base.AGENT_OPERATION_FAILURES` did not name the new exception, so since
client-python#380 merged, `_await_instance_ready`'s cloud-init retry loop
has silently stopped retrying a timed-out health check -- it retried them
before. Adding `AgentAwaitTimeout` restores that coverage.

This is why 7b **declines the master plan's instruction** to narrow the
tuple to a plain `AgentOperationFailed` reference. The instruction
conflates a catch site, which wants breadth, with an assert site, which
wants precision: only the two `test_get_missing_file` assertions narrow.
The operator confirmed that decision during planning.

**Decision 3 was corrected during execution.** The plan said to filter
events server-side by `event_type`. It does not: every client event
getter funnels into `_get_events()`, which takes one type and not a set,
and the server is `Optional[str]` all the way down to
`mariadb.get_object_events`, so three types would be three HTTP requests
per poll per object -- against a suite whose events traffic
`load_budget.py` already excludes from the database load budget by hand
and explains at length. It reads ten events per poll and filters
client-side instead, at the same request count.

The progress-event set was verified rather than assumed, and gained
`audit`: between a network reaching `initial` and reaching `created`, the
only events on the network object are `audit`, so `status` and `mutate`
alone would have covered a network create by its two bracketing state
changes and nothing in between.

**The new functional tests are unverified.** They need a real cluster,
so this PR's own CI run is the first time they execute. If the smoke
suite goes red on `smoke_ci_tests/test_agentop_deadlines.py`, that is a
phase-7 defect rather than a flake to re-queue.

## Status is deliberately still `In progress`

Fourteen of the phase plan's fifteen Definition-of-done items tick. The
fifteenth is "CI has run green on the branch", which this PR is what
answers. The plan's own risk section says the phase is not marked
complete on unit tests alone, so the flip to `Complete` -- and the index
arithmetic to `8 of 9` -- is a follow-up commit once CI is green.

*(Pull request prepared with Claude Code)*

### DIFF
--- a/docs/developer_guide/api_reference/instances.md
+++ b/docs/developer_guide/api_reference/instances.md
@@ -694,11 +694,17 @@ is meaningful.
 
 !!! note "What happens when a budget runs out"
 
-    An operation which exhausts either budget moves to `expired`, a
-    terminal state distinct from `error`. `error` means the operation
-    itself failed; `expired` means a budget you set ran out, and the
-    state's message says which. Nothing is retried and no further
-    command in the operation runs.
+    An operation which exhausts either budget, and for which no retry is
+    possible (see below), moves to `expired`, a terminal state distinct
+    from `error`. `error` means the operation itself failed; `expired`
+    means a budget you set ran out, and the state's message says which.
+    A wall-clock deadline running out is always final -- there is no
+    time left for a further attempt to deliver anything in -- but a
+    progress-timeout stall may be retried a bounded number of times
+    before landing in `expired`; see
+    [Agent Operations](../../operator_guide/agent_operations.md#retry-and-the-executor-reaper)
+    in the operator guide for the retry rules and the node-local reaper
+    that backs them up.
 
     Three places enforce this, so an abandoned operation is retired
     wherever it happens to be sitting: when it reaches the head of the
--- a/docs/operator_guide/agent_operations.md
+++ b/docs/operator_guide/agent_operations.md
@@ -1,0 +1,203 @@
+# Agent Operations
+
+An agent operation is a request handled by the in-guest `sf-agent2`
+process rather than by the hypervisor directly: executing a command,
+copying a blob into the instance, or fetching a file out of it. The
+API server creates an `AgentOperation` object for each request, which
+is enqueued against the target instance and dispatched by the
+sidechannel daemon once the instance's agent channel is ready. For the
+request parameters themselves -- `deadline_seconds`,
+`progress_timeout_seconds`, and what each of the three creating calls
+accepts -- see
+[the API reference](../developer_guide/api_reference/instances.md#bounding-how-long-an-agent-operation-may-take).
+This page is about what happens operationally once a request is
+enqueued.
+
+## One executor per instance
+
+The sidechannel daemon runs at most one executor thread per instance
+at a time: its dispatch loop skips any instance that already has a
+live executor, and `Instance.agent_operation_next()` leaves an
+executing operation at the head of the queue rather than handing out a
+second one. This is deliberate -- it is what makes dispatch crash
+safe, since the queue entry survives a daemon restart -- but it has an
+operational consequence worth knowing: an operation that wedges takes
+out that instance's *entire* agent surface until something ends it.
+Every other queued operation against the same instance, however
+unrelated, waits behind it. Nothing about this is per-command; it is
+per-instance.
+
+This is why the timing budgets below exist, and why the reaper further
+down exists too: a wedged operation with no budget and no reaper would
+block an instance's agent operations indefinitely.
+
+## The two budgets
+
+Every agent operation carries two independent timing budgets, either
+of which the caller may set explicitly per request. If they do not,
+the server applies a default.
+
+- **`AGENT_OPERATION_DEFAULT_DEADLINE`** (default 600 seconds,
+  `shakenfist/config.py:240`) is a wall-clock budget. It is counted
+  from the moment the API server received the request, not from when
+  the agent picked the work up -- so time spent queued behind another
+  operation on the same instance, and any preflight work such as
+  fetching a blob onto the hypervisor, both count against it before
+  the operation ever executes.
+- **`AGENT_OPERATION_DEFAULT_PROGRESS_TIMEOUT`** (default 30 seconds,
+  `shakenfist/config.py:259`) is how long a transfer may go without
+  making forward progress. It applies only to the commands that can
+  report progress at all -- the transfers behind `agent/get` and
+  `agent/put` -- and never to `agent/execute`: nothing an executed
+  command does is observable as progress, so the API refuses a
+  `progress_timeout_seconds` on that endpoint rather than accept one
+  that could never fire.
+
+A caller may send `0` for either parameter to disable it entirely;
+omitting the parameter takes the default instead. See the API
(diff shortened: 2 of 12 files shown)


sfconductor:

The conductor will merge this at `ea7ffe4b22b5` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `ea7ffe4b22b5`, bypassing the merge queue.

Issue #98, opened 2020-06-28
sf-client should gracefully handle common exceptions


mandoonandy:
sf-client should not have unhandled exceptions, especially resource not found.

```root@andy-200624-esom2noe-sf-1:~# sf-client network show 021adbe5-af42-4223-b6f9-6054f8c63649
Traceback (most recent call last):
  File "/usr/local/bin/sf-client", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/client/main.py", line 230, in network_show
    _show_network(ctx, CLIENT.get_network(network_uuid))
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/client/apiclient.py", line 206, in get_network
    '/networks/' + network_uuid)
  File "/usr/local/lib/python3.6/dist-packages/shakenfist/client/apiclient.py", line 105, in _request_url
    'API request failed', method, url, r.status_code, r.text)
shakenfist.client.apiclient.ResourceNotFoundException: ('API request failed', 'GET', 'http://localhost:13000/networks/021adbe5-af42-4223-b6f9-6054f8c63649', 404, '{"error": "network not found", "status": 404}')
```

PR #4070, opened 2026-09-05
Fix release asset upload in github-release job

mikalstill:
The `github-release` job downloaded artifacts with a bare
`actions/download-artifact@v8`. A bare download picks its own destination,
which is not the `dist/` directory that the following
`softprops/action-gh-release@v3` step globs with `files: dist/*`. Because
`fail_on_unmatched_files` defaults to false, a glob that matches nothing is
only a warning -- so the job reported success having attached no assets.

This looked intermittent rather than simply broken because these jobs share
a persistent `[self-hosted, static]` runner pool. When the release job
happened to land on the same runner as the build, the glob matched the build
job's leftover `dist/` in the reused workspace. That is to say even the
apparent successes were publishing unverified leftovers rather than the
digest-checked artifact download.

The fix names the artifact and its destination path -- exactly what the
`publish-pypi` job in the same workflow already does -- and sets
`fail_on_unmatched_files: true` so an empty release fails loudly.

This propagates the reviewed fix from shakenfist/development PR #98.

**Observed impact in this repo:** latent. No release has yet been published
through this path, so the defect has never had the opportunity to produce an
empty release here. Fixing it now means the first release does not have to
discover it.

Verified with the consistency audit's `release-process` criterion, which
reports `fail` on `develop` and `pass` with this change applied.

https://claude.ai/code/session_01GPedY5oTJXa7Mo343F8pSD

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,29 +138,42 @@ jobs:
       attestations: write
 
     steps:
-      # This job does not check out, so the workspace it starts in is
-      # whatever the last job on this runner left behind -- the static
-      # pool is persistent, and download-artifact adds to a directory
-      # rather than replacing it. Downloading into the per-job
-      # runner.temp is what makes the set of files here exactly the
-      # set that was built: nothing else can have put a file there.
-      # Everything downstream has to be pointed at the same place,
-      # including the publish step, whose own default is ./dist.
+      # This job works in the checked-out workspace, while
+      # publish-collection and github-release below download into
+      # runner.temp. The asymmetry is not an oversight:
+      # gh-action-pypi-publish is a composite action that delegates to
+      # a Docker container action, and a container sees the workspace
+      # mounted at /github/workspace and nothing else of the runner's
+      # filesystem. runner.temp is not mounted at its host path -- on
+      # a self-hosted runner it is a sibling of the workspace, not a
+      # child -- so packages-dir has to be relative to the workspace,
+      # and an absolute path fails there however it is spelled. The
+      # action needs a real workspace regardless, since it writes its
+      # container trampoline into .github/.tmp/ before running it.
+      #
+      # Checking out is what makes that workspace safe to use. This
+      # pool is persistent and a job which skips checkout inherits
+      # whatever the last job left behind, but checkout cleans with
+      # "git clean -ffdx", which removes a stale gitignored dist/ --
+      # the same guarantee runner.temp buys the other jobs.
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
       - name: Download distribution artifacts
         uses: actions/download-artifact@v8
         with:
           name: dist
-          path: ${{ runner.temp }}/dist/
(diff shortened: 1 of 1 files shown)


mikalstill:
Pushed a follow-up commit in response to the review.

The download in every publishing job that does not run a checkout now
targets `${{ runner.temp }}/dist/` rather than the shared workspace.
Those jobs start in whatever the previous job left on the persistent
static pool, and `download-artifact` extracts *into* a directory
rather than replacing it, so a stale file could still join the set
being published. `runner.temp` is per job, so nothing else can have
written there.

Everything that reads the distribution moved with it: `subject-path`
on the attestation, `files` on the release, and `packages-dir` on the
PyPI upload -- that last one matters because the action defaults it to
`./dist` and does not fail when the directory is wrong, so leaving it
out would have kept the upload reading the shared workspace.
`fail_on_unmatched_files: true` stays.

This covers `publish-collection` as well. It downloads into
`${{ runner.temp }}/dist-collection/` now, and the `ansible-galaxy
collection publish` command that globs that directory moved with it.

The `build` and `sign-tag` jobs are deliberately untouched: they check
out, and checkout cleans with `git clean -ffdx`. The explanatory
comments were reworked to state the invariant rather than recount the
incident. This follows the reviewed template in shakenfist/development
pull request 100.

mikalstill:
Round three.

Round two sent every publishing job's download to `${{ runner.temp }}`. That is right for `publish-collection` and `github-release`, and wrong for `publish-pypi`: `pypa/gh-action-pypi-publish` delegates to a Docker container action, and the container gets exactly two bind mounts. From the real `docker run` of a library-utilities v0.8.8 release:

```
--workdir /github/workspace
-v ".../_work/_temp":"/github/runner_temp"
-v ".../_work/<repo>/<repo>":"/github/workspace"
```

`RUNNER_TEMP` is mounted, but at `/github/runner_temp` -- not at its host path, which is what `${{ runner.temp }}` expands to. So `packages-dir` never resolved inside the container. It has to be relative to the workspace.

So `publish-pypi` goes back to `dist/` in the workspace and gains an `actions/checkout`. Checkout is what makes that safe: `git clean -ffdx` removes a stale gitignored `dist/` from an earlier job on the persistent static pool -- the same guarantee `runner.temp` was bought for, obtained the only way the container can use.

The distinction is not "publishing job" but *who reads the download*, so the other two are deliberately unchanged. `publish-collection` hands its tarball to an `ansible-galaxy collection publish` **`run:` step on the host**, not a container action, so it keeps `${{ runner.temp }}/dist-collection/` and gains no checkout. `github-release` hands its files to `action-gh-release`, a JavaScript action, likewise on the host.

Validated rather than reasoned about: library-utilities v0.8.8 released with this shape, published both files to PyPI with provenance, and attached both assets to the GitHub release.

Audit: `release-process` still reports only the guard complaint on this branch alone, and passes when merged with `release-tag-guards` (merge tested locally, then aborted). `pre-commit run --all-files` passes.

PR #4071, opened 2026-09-05
Clean up the declared dependency set

mikalstill:
Fixes the two dependency consistency audits, one commit each.

### Declare the dependencies we import directly (#4044)

Four packages were imported from a name that only appeared in the
generated indirect dependency block. That block is regenerated from
whatever the direct dependencies happen to resolve to, so an import of
one of its names works only until the intermediate library drops the
requirement.

Three are now declared above the `START_OF_INDIRECT_DEPS` marker at the
versions they already resolved to, and `tools/pin-indirect-dependencies.sh`
dropped the generated copies:

| Package | Imported by | Was arriving via |
| --- | --- | --- |
| `pylogrus` | `logship.py` (`JsonFormatter`) | `shakenfist-utilities` |
| `jsonschema` | `tools/render-review.py` | `flasgger` |
| `packaging` | `tools/build-collection.py` | the oslo stack |

The fourth needed no declaration, because the import should not have
been there: `test_dnsmasq.py` reached for `six.moves.builtins` to patch
`open()`, which on a project requiring Python 3.11 or later is spelled
`builtins`. `six` stays in the generated block, since other packages
still require it.

### Stop declaring dependencies we do not use (#4043)

Eleven dependencies were declared and imported nowhere, all of them
dating from the conversion of `requirements.txt` to `pyproject.toml`, so
none arrived with a reason recorded. Each was traced to how the package
is actually used before choosing between removing it and recording why
it is installed.

Three are genuinely used without being imported, and now carry the
`# not-imported:` annotation the audit asks for:

* `uv`, which the node bootstrap runs as `uv pip install`.
* `clingwrap`, a CLI installed into the node virtualenv, whose built-in
  targets include one for Shaken Fist.
* `grpcio-tools`, which `protos/_make_stubs.sh` runs as
  `python3 -m grpc_tools.protoc` from this same environment so that
  generated stubs match the runtime `protobuf` and `grpcio` pins.

Four are still required transitively, so dropping the direct declaration
moves them into the generated block rather than out of the install:
`gevent` and `greenlet` arrive with the `gunicorn[gevent]` extra, `pbr`
with the oslo stack, and `urllib3` with `requests`.

The remaining three leave the dependency closure entirely, taking it
from 87 packages to 84: `chardet` (`requests` resolves text encodings
through `charset-normalizer`), `requests-toolbelt` and `grpcio-status`.
Nothing in the tree imports `grpc_status`, `google.rpc` or
`status_pb2`.

### The one finding not fixed here

`flask-request-id-middleware` is left exactly as it is. It **is**
imported, by `external_api/app.py`, but the audit derives candidate
module names from the distribution name and cannot get from
`flask-request-id-middleware` to the `flask_request_id` the wheel
actually installs. Annotating it here would assert something untrue, so
the fix is an `IMPORT_NAME_ALIASES` entry in the development
repository, which is what `unused-declared-dependency.md` asks for:
shakenfist/development#99.

### Verification

* `pre-commit run --all-files` passes.
* `tools/pin-indirect-dependencies.sh` resolves and installs the full
  closure cleanly, and reports the block up to date afterwards.
* `scripts/audit-check.py` reports `unused-declared-dependency` and
  `undeclared-direct-dependency` as pass (the former with the
  development-repo alias applied).

https://claude.ai/code/session_011cF33EswqGwZBR8QTRssfQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,17 @@ dependencies = [
     "semver==3.0.4",                      # bsd
     "SQLAlchemy==2.0.52",                 # mit
 
+    # Imported directly, rather than reached through whichever package
+    # happens to require them today, so what keeps them installed is a
+    # declaration and not a transitive pin the reconciler could drop.
+    # logship.py imports pylogrus (which arrives via
+    # shakenfist-utilities), tools/render-review.py imports jsonschema
+    # (via flasgger) and tools/build-collection.py imports packaging
+    # (via the oslo stack).
+    "jsonschema==4.26.0",                 # mit
+    "packaging==26.3",                    # apache2
+    "pylogrus==0.4.0",                    # mit
+
     "greenlet==3.5.5",                    # mit
     "gevent==26.8.0",                     # mit
     "gunicorn[gevent]==26.2.0",           # mit
@@ -131,7 +142,6 @@ dependencies = [
     "iso8601==2.1.0",
     "itsdangerous==2.2.0",
     "jsonschema-specifications==2025.9.1",
-    "jsonschema==4.26.0",
     "MarkupSafe==3.0.3",
     "mistune==3.3.4",
     "netaddr==1.3.0",
@@ -139,9 +149,7 @@ dependencies = [
     "oslo.config==10.7.0",
     "oslo.i18n==6.9.0",
     "oslo.utils==10.2.0",
-    "packaging==26.3",
     "pycparser==3.0",
-    "pylogrus==0.4.0",
     "pyparsing==3.3.2",
     "python-dotenv==1.2.3",
     "pytz==2026.3.post1",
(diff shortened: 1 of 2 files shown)


sfconductor:

The conductor will merge this at `c814cd7d4378` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

The conductor will merge this at `c814cd7d4378` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `c814cd7d4378`, bypassing the merge queue.

Issue #4077, opened 2026-09-05
Functional coverage gaps in agent operation deadlines: retry, attempt bound, and the no-budget combination


mikalstill:
Found by the phase 8 push audit of `docs/plans/PLAN-agent-operation-deadlines.md` (wave 2b, test review).

The agent operation deadlines plan landed good unit coverage — six new test
modules and about 1,800 added lines in
`shakenfist/tests/test_daemon_sidechannel_executor.py`. Its functional coverage
is `shakenfist/deploy/shakenfist_ci/smoke_ci_tests/test_agentop_deadlines.py`,
which has four tests:

* `test_default_deadline_is_published`
* `test_expiry_frees_the_executor_slot`
* `test_silent_execute_survives_the_progress_timeout`
* `test_queued_operation_expires_on_its_deadline`

CLAUDE.md prefers functional coverage to unit coverage where a project can only
have one. These are the behaviours the four do not reach.

## Worth adding

**Retry to completion.** Nothing functionally exercises phase 5's
`EXECUTING -> QUEUED` edge — that a stalled operation is re-dispatched and then
succeeds. This is achievable with a controlled stalling guest command, and it is
the plan's central behavioural claim.

**Attempt-bound exhaustion.** Nothing drives an operation to
`AGENT_OPERATION_MAX_ATTEMPTS` and asserts it then reaches a terminal state
rather than looping. Same mechanism as above, with a command that stalls every
time.

**The no-deadline, no-progress-timeout combination.** No test, unit or
functional, drives `expire_if_out_of_budget()` with both the `deadline_seconds=0`
sentinel and a disabled progress timeout at once. The audit's security review
found that this combination is exactly what parks an executor slot indefinitely
(#4074), and it went unnoticed partly because nothing tests it.

## Deliberately not proposed

**The reaper's daemon-restart path.** Recovering an operation left executing by
a dead `sf-sidechannel` needs a daemon restart mid-suite. The project's own
`test_health.py` declines to restart daemons in shared CI for exactly this
reason, and the path is already well covered at the unit level. Recorded here so
the omission is a decision rather than a gap nobody noticed.

Related: #4074, #3995, #3516.

*(Audit assisted by Claude Code)*

Issue #4076, opened 2026-09-05
Client capability cache can be stale across a rolling sf-api upgrade, turning a gated parameter into a 500


mikalstill:
Found by the phase 8 push audit of `docs/plans/PLAN-agent-operation-deadlines.md` (cross-repository contract step). Filed so this is a decision rather than an oversight; it may well be accepted as-is.

## Symptom

`shakenfist_client`'s `_collect_capabilities()` (`client-python/shakenfist_client/apiclient.py:305-308`)
fetches the API root page **once**, in the `Client` constructor, and
`check_capability()` substring-matches that cached text for the life of the object.

Behind a load balancer during a rolling `sf-api` upgrade, a client can read the
root page from an already-upgraded worker, see `agentoperation-deadlines`
advertised, and then POST `deadline_seconds` to a worker which has not
restarted yet.

That is worse than an ignored parameter. `log_request()`'s `kwargs.update(j)`
(`shakenfist/external_api/base.py:1222-1231`) merges the JSON body into the
handler's kwargs unconditionally, so an undeclared `deadline_seconds` arrives
as an unexpected keyword argument to a handler that has no such parameter.

## Scope

This is a property of the capability mechanism generally rather than of the
agent operation deadlines work, and the exposure window is a single deploy.
What raises it above noise is that the deadline parameters are the first
feature to depend on the capability token for request *validity* rather than
for graceful degradation — every earlier use of `check_capability()` guards a
feature the client can simply decline to use.

## Options

1. Accept it, and record the reasoning: rolling upgrades are short, the failure
   is a clear 500 rather than silent corruption, and a retry after the deploy
   succeeds.
2. Re-read capabilities on a 500 that looks like an unexpected-kwarg failure,
   and retry once.
3. Make `log_request()` drop body keys the handler does not declare, rather than
   merging them blind. This is the broadest fix and would help every future
   capability-gated parameter, but it changes a shared code path.

Related: the phase 8 push audit plan in `docs/plans/`.

*(Audit assisted by Claude Code)*

Issue #4075, opened 2026-09-05
Which budget expired an agent operation is not reachable through the API


mikalstill:
Found by the phase 8 push audit of `docs/plans/PLAN-agent-operation-deadlines.md` (cross-repository contract step).

## Symptom

An agent operation can now reach `expired` for two quite different reasons — its
wall-clock deadline ran out, or it made no forward progress inside its progress
timeout. Which one happened is not reachable through the API.

`expire()` (`shakenfist/operations/agentoperation.py:320-349`) records the reason
as the state row's message. But:

* `AgentOperation.external_view()` (`:187-214`) returns
  `error_message = self.error`.
* `baseobject.error` (`shakenfist/baseobject.py:629-638`) returns `None` unless
  the state value ends in `error`. `expired` does not.
* `_external_view()` (`baseobject.py:690-707`) renders the `State` to its value
  string, dropping the message.

So an expired operation comes back over the wire as `state='expired'`,
`error_message=None`, and the reason exists only in the instance's event log.

## Why it matters

The two reasons want different responses. A wall-clock expiry on a big transfer
means "ask for a longer deadline". A progress-timeout expiry means "the guest
agent stalled", which is shakenfist#3995's situation and is not fixed by a
bigger deadline. A client that cannot tell them apart cannot advise its user,
and cannot make a sensible retry decision either.

`docs/operator_guide/agent_operations.md:92-99` states the limitation honestly.
But `docs/release_notes/v07-v08.md:753` tells a reader that `expired` means a
budget ran out "and the state's message says which", without saying that the
message is not on the wire — which is exactly the sentence an integrator would
read before writing code that cannot work.

## Suggested fix

Either surface the state message on the external view for non-error terminal
states, or add an explicit discriminator field (something like
`expiry_reason: 'deadline' | 'progress'`). The second is probably better: it is
a small enumerated fact rather than free prose, and a client can branch on it
without parsing English.

Whichever is chosen, `docs/release_notes/v07-v08.md:753` needs to stop implying
the message is already available.

Related: #3995, #3516.

*(Audit assisted by Claude Code)*

Issue #4074, opened 2026-09-05
agent/execute with deadline_seconds=0 parks an instance's only executor slot indefinitely, with no operator ceiling


mikalstill:
Found by the phase 8 push audit of `docs/plans/PLAN-agent-operation-deadlines.md` (wave 2d, security review).

## Symptom

A single request parks an instance's only agent executor slot for as long as the
guest command runs, with no server-side bound and no operator knob:

```
POST /instances/<uuid>/agent/execute
{"command_line": "sleep 31536000", "deadline_seconds": 0}
```

The sidechannel dispatch loop runs at most one executor per instance
(`_dispatch_loop` skips any instance with a live executor), so while that
operation runs, **no other agent operation against that instance can be
dispatched at all**. That is the same failure mode as #3516, which the agent
operation deadlines plan was written to fix — reintroduced as a
client-requestable condition.

## Mechanism

Both timing budgets are disabled at once, and it takes only one parameter to do it:

1. `deadline_seconds: 0` is the documented "no wall-clock deadline" sentinel.
   `AgentOperation.effective_deadline()` returns `None` for it
   (`shakenfist/operations/agentoperation.py:288`).
2. The `agent/execute` endpoint **always** stores `progress_timeout = 0.0`,
   because no command it builds can report progress
   (`shakenfist/external_api/base.py:274-279`). So
   `effective_progress_timeout()` returns `None` too — the caller does not
   have to ask for this half, and cannot decline it.

`expire_if_out_of_budget()` therefore returns `False` on every pass, and the
socket loop's only remaining exit is the abort file
(`shakenfist/daemons/sidechannel/main.py:1293`). The guest's own ping replies
keep the socket alive.

The node-local reaper cannot recover it either:

* Case one needs `not executor` (`main.py:1693`) — there is a live executor.
* Cases two and three both hinge on `deadline_passed()` (`main.py:1722`),
  which is `False` by construction.

The reaper's own docstring (`main.py:1636-1645`) already concedes a narrower
version of this for an executor wedged *before it connects*. The audit's
finding is that the same hole is reachable by a healthy, connected executor
running a long command.

## Why this is worse than a known limitation

`291054e98` (phase 4) deleted the unconditional
`AGENT_OPERATION_EXECUTION_TIMEOUT = 900` backstop, whose comment named exactly
this #3516 scenario. There is no `AGENT_OPERATION_MAX_DEADLINE`, and
[...]

PR #4078, opened 2026-09-05
Agent operation deadlines phase 8: push audit

mikalstill:
Phase 8 of [PLAN-agent-operation-deadlines.md](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-agent-operation-deadlines.md): the push audit, plus the phase 7 closeout that was never run.

`PUSH-AUDIT.md` normally runs against `develop...HEAD` before a push. Every phase of this plan is already merged, so that diff is empty and every command in the template would have reported success against nothing. The baseline here is instead an explicit list of twelve merges — the nine phase merges, plus the three defect fixes that landed against this plan's code *outside* its phase branches (#3933, #3970, #4025). Correcting that description in the master plan was itself one of the survey's findings.

## What the audit found

**Wave 1 clean.** pre-commit, `tox`, proto freshness, and the four style greps run against each of the twelve merges separately. Both blocking template rules — SQL pushdown and cached FK lists — had zero candidates anywhere in the footprint, confirmed by running the checks rather than assuming.

**Three blocking findings.** Two are fixed here; one is filed.

### The one that matters: #4074

`agent/execute` with `deadline_seconds: 0` disables **both** timing budgets at once. The zero sentinel makes `effective_deadline()` return `None`, and the execute endpoint always stores `progress_timeout = 0.0` because no command it builds can report progress — so the caller does not ask for that half and cannot decline it. `expire_if_out_of_budget()` then returns `False` on every pass, and all three reaper cases are unreachable: one needs the executor to be absent, the other two hinge on `deadline_passed()`. Phase 4 deleted the unconditional 900-second backstop whose comment named this exact #3516 scenario, and there is no `AGENT_OPERATION_MAX_DEADLINE` for an operator to bound it with.

This plan existed to stop a stuck operation monopolising an instance's only executor slot. It succeeded for the case it set out to fix, and then handed callers a parameter combination that reinstates the same monopoly on request. No phase could have seen it: phase 3 added the parameter, phase 4 removed the backstop, phase 5 wrote the reaper that cannot recover it, and each was correct on its own terms.

Filed rather than fixed here, because the fix is an operator-visible ceiling plus a restored backstop — a behaviour change larger than an audit phase should carry. The plan's risk section anticipated this outcome and said to file at high priority rather than quietly downgrade it. Scoped honestly: `requires_instance_ownership` confines the starvation to the caller's own namespace, and daemon shutdown is unaffected.

### The two fixed here

Both were true when written and became false when a later phase landed — the cost of writing tests and documentation against a plan still in motion, not coding errors.

- **A regression guard that swallowed its own regression.** `RequestAllThreadsExitTestCase` wrapped `_request_all_threads_exit()` in `try: ... except KeyError: pass`, commented "#3931 … fixed separately". `b915cb018` fixed it on 2026-08-27; those tests were written on 2026-08-29, and no `KeyError` path remains.
- **A release note telling operators there is nothing to look for.** `docs/release_notes/v07-v08.md` claimed both client changes were capability-gated so behaviour was unchanged across an upgrade. Neither half is true: the same document says in bold seventy lines earlier that the default got tighter (600 s + 30 s replacing a 900 s backstop — #3995's mechanism), and `_await_agentop` raises `AgentOperationFailed` with no capability check at all.

## The only production code change

`_request_thread_exit()` joined a thread that may not have started. `start_instance_executor()` deliberately registers an executor *before* starting its thread, and `reap_instance_executors()` guards that exact window with `if t['thread'].ident is None: continue` and a comment explaining why. The shutdown path — generalised by `2341ae0c4` to serve monitors and executors alike — did not carry the guard over, so a SIGTERM in that window raises `RuntimeError` out of graceful shutdown and leaves the remaining threads unsignalled. Four lines plus a comment, mirroring the guard eight lines away, with a test that was mutation-tested: removing the guard fails it.

This is the one change here that wants a human read before merge.

## Disposition

3 blocking, 16 advisory. Five fixed here, seven filed, four declined with stated reasons, one absorbed into #4074. Full table in the phase plan.

Filed: #4074, #4075, #4076, #4077, and client-python #388, #389, #390.

## Closeout

Phase 7's work merged as `4a122bcd3` and its gating CI run passed with all four new deadline tests reported `ok` rather than skipped — but step 7f never ran, so the tables still said In Progress. That is corrected here.

The master plan gains a **Known defects** section naming #4074, #3995 and #4039, and a table of the three repairs that landed outside its phase branches, so a reader arriving at a completed plan does not have to search the tracker to find either. The plan now reads `9 of 9`, Complete.

## Verification

- `pre-commit run --all-files` passes (ten hooks, including unit tests and mypy).
- `tox` passes: `py3`, `flake8`, `cover`.
- `python3 tools/check-plan-status.py` passes.
- The new guard test was mutation-tested against its own fix.

https://claude.ai/code/session_0152ZmvXNTwMGPMsmek95bru


sfconductor:

The conductor will merge this at `b98297dd64f4` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `b98297dd64f4`, bypassing the merge queue.

Issue #4021, opened 2026-09-01
test_no_unbudgeted_fixed_rate_database_polling: eight workload pairs (api and queues locks, state and attribute writes) certified as polls on a high-spread tier run


mikalstill:
Found during merge CI triage of PR #4016. That PR changes the `.vv` generator
in `InstanceVDIConsoleHelperEndpoint`, `Node.spice_server_cert_subject`, and
two tests. Its own new functional test ran and passed in 20.5s in this very
job, and none of the eight pairs reported below touches the VDI console path,
so it did not cause this.

## Symptom

`cluster_ci_tests.test_database_tier.TestDatabaseTier.test_no_unbudgeted_fixed_rate_database_polling`
failed in the **Debian 12 tier (collection)** job of PR #4016's merge group.
It was the only failing test in the run.

- Run: https://github.com/shakenfist/shakenfist/actions/runs/33557865633
- Job: [`Debian 12 tier (collection) / Smoke tests (collection)`](https://github.com/shakenfist/shakenfist/actions/runs/33557865633/job/100023059909) (100023059909)
- Base: `0bb391635` (develop, after #4002)

```
testtools.matchers._impl.MismatchError: !=:
reference = []
actual    = [(('AcquireLock', 'api'), 0.31640630200629366),
 (('AcquireLock', 'queues'), 0.4496300081142068),
 (('FindNetworkInterfaces', 'queues'), 0.2831156289587039),
 (('FindNetworks', 'api'), 0.4496300081142068),
 (('ReleaseLock', 'api'), 0.31640630200629366),
 (('ReleaseLock', 'queues'), 0.4496300081142068),
 (('SetObjectState', 'api'), 1.2156663182347072),
 (('UpdateInstanceAttributes', 'queues'), 0.43310944781695704)]
```

```json
"activity_qps_per_window": [244.732, 157.968, 166.027, 120.191],
"activity_spread": 2.036, "activity_spread_required": 1.8,
"nodes": 3, "standing_instances": 4, "unbudgeted_ceiling_qps": 0.25,
"pairs_seen": 373, "pairs_fixed_rate": 120, "pairs_metronomic": 86,
"known_polls_recognised": [
  "GetNodeDaemonState/cleaner", "GetNodeDaemonState/cluster",
  "GetNodeDaemonState/net", "GetNodeDaemonState/queues",
  "GetNodeDaemonState/resources", "GetNodeDaemonState/sidechannel",
[...]

PR #4081, opened 2026-09-05
Triage failed merge queue runs automatically

mikalstill:
When `Functional tests` fails on a `merge_group` event, GitHub ejects the pull
request from the merge queue and somebody has to work out whether the pull
request broke something or whether the test cloud fell over again. This
automates that first pass, following the same steps the `merge-ci-triage` skill
describes: find the first failing step, classify the failure as PR-caused or
systemic, record a systemic failure against its tracking issue, and recommend
re-queueing or fixing first. The verdict is posted as a comment on the ejected
pull request.

## Why `workflow_run` and not a job in `functional-tests.yml`

The obvious implementation is another job gated on `can_merge` having failed.
It races the queue: once `Can merge` reports its failure the queue ejects the
pull request and tears down the `gh-readonly-queue` ref the run is on, and a
triage which takes minutes sits in exactly that window. Moving the triage into
an earlier *step* of `can_merge` avoids the race but holds the merge queue open
while a model reads logs. A job reachable on `merge_group` would also fall
under the `merge-group-cancellation` audit, whose correct behaviour -- be
cancelled when a newer merge group supersedes you -- is the opposite of what a
triage job wants.

`workflow_run` runs afterwards, on the default branch, in the base repository
context, and none of that can cancel it. The `branches: ['gh-readonly-queue/**']`
filter keeps the workflow from being invoked for every pull request run of the
suite as well; without it each of those leaves a skipped run in the Actions
list.

The workflow never checks out or runs the code it is triaging. It reads the
failed run through `gh` and checks out the default branch, which is what makes
a writable token safe in a workflow that reads a pull request diff.

## The verdict document

Each triage produces one JSON document matching `tools/merge-triage-schema.json`,
published both as the `merge-triage-<run id>` build artifact and embedded in a
collapsed `<details>` section of the pull request comment, the way the automated
reviewer embeds its review. The private-ci conductor will consume these to track
which merge failures have been triaged and which of them blamed the pull
request, so three properties exist for that consumer:

- **A document is always written.** A model which answers in prose, or produces
  nothing, yields `verdict: unknown` with an `error` rather than a missing file.
  A missing document is indistinguishable from a triage that never ran.
- **The envelope is not model output.** Repository, run id and pull request
  number are written from what GitHub said and overwrite whatever the model put
  in those fields, so a verdict cannot be filed against the wrong failure. Only
  the fields in `MODEL_FIELDS` are taken from the response at all.
- **A cited tracking issue has been checked.** It has to exist and to carry a
  reference to the failed run, or it is dropped with the reason appended to
  `evidence`.

## Verification

Dry-run against a real failure -- run 33952027147, the merge group for #4080 --
classified it systemic against #4021 (the `(FindNetworks, api)` unbudgeted-poll
signature) and #3772 (the `507 sufficient_idle_cpu` family), with a re-queue
recommendation. That is the verdict manual triage reached, and it took about
two minutes.

`tools/merge-triage.py` has 12 unit tests covering the extraction and rendering
edge cases; both mechanisms that keep model output out of the envelope were
mutation tested. `pre-commit run --all-files` passes.

Note that `workflow_dispatch` only appears once this is on `develop`, so the
first test of the trigger itself happens after it lands.

https://claude.ai/code/session_01G2iPhBZTSqq2fKt7ovFipv

### DIFF
--- a/.github/workflows/issue-fix.yml
+++ b/.github/workflows/issue-fix.yml
@@ -645,9 +645,15 @@ jobs:
           # subscription credit is exhausted (HTTP 429). Every model being
           # out of credit exits 1 and produces no output, which the assess
           # step below sees as "no changes".
+          #
+          # The prompt goes in by file, not as an argument: Linux caps a single
+          # argv element at 128KiB and an issue with a long body and a long
+          # comment thread can reach that, where the exec fails outright rather
+          # than truncating and no model runs at all.
           ${{ runner.temp }}/claude-model-fallback.sh \
             --models "${MODELS}" \
-            -- "$(cat ${{ runner.temp }}/claude-prompt.txt)" \
+            --prompt-file ${{ runner.temp }}/claude-prompt.txt \
+            -- \
             --dangerously-skip-permissions \
             --max-turns "${MAX_TURNS}" \
             --output-format text \
--- a/.github/workflows/merge-failure-triage.yml
+++ b/.github/workflows/merge-failure-triage.yml
@@ -1,0 +1,137 @@
+# Triage a failed merge queue CI run with Claude Code
+#
+# When "Functional tests" fails on a merge_group event, GitHub ejects the pull
+# request from the merge queue and somebody has to work out whether the pull
+# request was at fault or whether the cluster fell over again. This automates
+# that first pass: it follows the merge-ci-triage skill, posts a verdict on the
+# pull request, and records a systemic failure against its tracking issue.
+#
+# WHY workflow_run AND NOT A JOB IN functional-tests.yml:
+# The obvious implementation is a job in the merge group run itself, gated on
+# "can_merge" having failed. It races the queue. Once can_merge reports its
+# failure the queue ejects the pull request and tears down the
+# gh-readonly-queue ref the run is on, and a triage job which takes minutes is
+# in exactly that window. Putting the triage in an earlier step of can_merge
+# avoids the race but holds the whole merge queue open while a model reads
+# logs. A job reachable on merge_group would also fall under the
+# merge-group-cancellation audit, whose correct behaviour -- be cancelled when
+# a newer merge group supersedes you -- is the opposite of what a triage job
+# wants. workflow_run runs afterwards, on the default branch, and cannot be
+# cancelled by any of that.
+#
+# The branches filter is doing real work: without it this workflow would be
+# invoked for every pull request run of the test suite as well, and every one
+# of those would show up in the Actions list as a skipped run. Merge group runs
+# are the only ones on a gh-readonly-queue ref.
+#
+# SECURITY CONSIDERATIONS:
+# - workflow_run runs in the base repository context with a writable token, and
+#   is deliberately never given the merge group's code. The checkout is the
+#   default branch; the failed run is read through gh, never executed.
+# - The model does read logs and a pull request diff, which are untrusted
+#   input. Its GitHub reach is bounded by the token permissions below: it can
+#   comment on issues and pull requests and nothing else. It cannot push, and
+#   the workflow does not check out or run any of the code it is triaging.
+# - The comment body is model output and goes through
+#   tools/neutralise-pr-body.sh before posting, so an @mention or an
+#   issue-closing keyword cannot fire on publication.
+#
+# NOTES:
+# - The verdict is uploaded as a build artifact and embedded in the pull
+#   request comment, both as tools/merge-triage-schema.json documents. The
+#   private-ci conductor consumes these to track which merge failures have been
+#   triaged and which of them blamed the pull request, so a document is written
+#   even when the model produces nothing usable -- see the header comment on
+#   tools/merge-ci-triage.sh.
+# - Scratch files live in runner.temp, never the workspace.
+
+name: Merge failure triage
+
+on:
+  workflow_run:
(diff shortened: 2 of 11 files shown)


mikalstill:
Addressed the review: all four fix items, both document items, and six of the seven suggestions.

- **1** — `actions: read` added. The block named scopes without it, which sets every unnamed scope to `none`, so every evidence read would have 403'd. Good catch; the dry run that validated the design ran as me, which is exactly where a permissions bug hides.
- **2** — `summary` is nullable in the schema now, matching its sibling prose fields.
- **3** — gather failures are recorded rather than swallowed: each appends a note that travels in `evidence`, the prompt is told a listed gap is missing rather than uninteresting, and with neither the failed job list nor the logs readable no model runs at all.
- **4** — an unstated `tracking_issue_action` defaults to `none` rather than being guessed as `commented`, and a dry run forces `none` unconditionally. Verification no longer gates on the action, so a dry run exercises that path too and records what it found without dropping the citation.
- **5** — the security note now says what the token actually allows, and that the prompt's "do not close anything" is a prompt-level constraint rather than a token-level one.
- **6** — the skill references are gone; it is a personal skill, not something in this repository, so the procedure is described directly.
- **7, 8, 10, 11, 12, 13** — taken: fences and `</details>` are stripped from model prose before it reaches the comment; an EXIT trap over an early envelope keeps the always-a-document promise for the paths that fail before a model runs; the tracking issue check pages every comment and matches the run URL rather than the bare id; envelope construction moved from jq into `merge-triage.py`, which makes the run id numeric by construction and the queue ref parse testable; the response scan prefers a fenced block and bounds the fallback; the audit link matches the github.com form used elsewhere in the file.
- **9** — partly. The ref parsing you asked to see pinned moved into Python and has a table of shapes including a base branch with a slash in it. A harness driving the shell with a stub `gh` is not in this round.

Ten new tests; the four new guards were mutation tested. Verified end to end with a dry run against run 33944911413, plus the unreadable-run, wrong-event and non-numeric-run-id paths.

mikalstill:
Addressed the second review: all five fix items, the document item, and six of the seven suggestions.

- **1** — nothing this job executes is left in the workspace. The driver, `merge-triage.py` and its schema, the neutraliser and the model wrapper are staged into `runner.temp` before the model starts, the way `issue-fix.yml` does it. Two hazards, one fix: bash reads the running driver lazily, and the extractor and neutraliser run *after* the model exits. With the staging in place a workspace write does nothing — the checkout is discarded with the runner and never pushed. I did not also restrict `--allowedTools`; the model legitimately needs `Bash` for arbitrary `gh` sub-commands, and once the executed set is out of reach the residual risk is the token, which a tool list does not bound.
- **2** — you were right about which end matters. Both are kept now, the tail with the larger share, elision marked inline with the byte count. Verified against run 33968991434, whose `--log-failed` is 1.25MB.
- **3** — `~~~` and `<details open>` both stripped; the test counts opens as well as closes.
- **4** — the gather-notes append moved out of the model branch. Confirmed against a blinded `gh`: all four notes now travel on the path where nothing could be read.
- **5** — verification is scoped to the claim it verifies. A claimed write that does not check out loses the claim and keeps the number ("referenced only" is now reachable — the dry run below produced one); an issue that cannot be read loses both. The claimed action is captured before the dry run overwrites it, so that path is still exercised. This also turned up a related bug: `gh api` prints a 404's error body on *stdout*, so an unreadable issue yielded plenty of text and took the wrong branch. Readability is the exit status now.
- **7, 8, 10, 13, 14** — taken. Table cells are flattened for newlines, pipes and backticks (the embedded JSON is untouched); the response scan prefers an object carrying a `verdict` key so a trailing appendix cannot degrade a good verdict to `unknown`; `REPO` is validated beside the run id; the header says "printed at startup"; the evidence comprehension runs the regex once.
- **9** — four.
- **6** — half. There is a new `test_merge_triage_workflow_seams.py` pinning the staging, the step order, `OUTPUT_DIR` against the artifact paths and the permissions block. A shell harness with a stub `gh` is still not in the tree; the branches it would cover were exercised by hand instead, against a stubbed model and a read-only `gh` shim, for all six tracking-issue outcomes and the no-evidence path.
- **11** — kept, and now said out loud in `ci.md` and in the script. Silence on an ejected pull request reads as "triage never ran", which is the ambiguity the always-write-a-document rule exists to remove; moving it from the artifact to the thread is not a fix.
- **12** — noted; the first `workflow_dispatch` with `dry_run: true` after this lands is the plan.

Four new tests, and every new guard was mutation tested. `pre-commit run --all-files` passes.

mikalstill:
Third round addressed in f16bd626: both fix items (the neutralisation was fail-open, and the tracking issue check could pass vacuously against an empty run URL) plus nine of the eleven suggestions. The two items the review marked as needing no action are left as they are. The biggest change is a new `shakenfist/tests/test_merge_ci_triage_shell.py`, which runs the driver for real with `gh` stubbed on PATH and the model wrapper emitting a canned response -- fifteen cases over the citation downgrades, the dry run forcing, the log budget and the dedup marker. Every guard added this round was mutation tested: thirteen deliberate breakages, each caught by the test that claims to hold it.

sfconductor:
Merged at `d06f51c2ac84`, bypassing the merge queue.

PR #4085, opened 2026-09-05
Let explicit address requests take over halos.

### LINKED ISSUE #4059: sf_instance cannot replace an instance with a static address: the deletion halo always 409s the recreate
### Summary

`sf_instance` cannot replace an instance whose networkspec pins a static
address. The replacement path deletes the old instance and immediately
recreates it, but deleting an instance puts its address into a
`deletion-halo` reservation for `IP_DELETION_HALO_DURATION` (default 300s).
The recreate therefore always fails:

```
('API request failed', 'POST', 'https://.../api//instances', 409,
 '{"error": "address 10.0.7.10 in use", "status": 409}')
```

The instance is already gone by then, so the failure is destructive: a
playbook re-run against an existing deployment destroys a working node and
cannot bring it back for five minutes.

### This is already known, and worked around in the test rather than fixed

`shakenfist/deploy/ansible_module_ci/004.yml` is the only scenario that
exercises replacement, and it deliberately steps around this case:

```yaml
    # NOTE(mikal): we need to change the IP address here as well, otherwise we
    # bump into the deletion halo around the previous IP address.
    - name: Change the instance and check for replacement
      ...
          - "network_uuid={{ ci_network_b['meta']['uuid'] }},address=10.1.0.6,float=True"
```

Because the test changes the address on the way through, replacement at a
fixed address has no coverage at all.

### Reproduction

Two runs of the same task, the second with `ram` bumped from 1024 to 2048 to
make the spec legitimately dirty:

```yaml
- name: Create the instance
  sf_instance:
    name: sfrepro-vm
    cpu: 1
    ram: 2048          # was 1024 on the first run
    disks:
      - "20@sf://label/system/centos:7"
    networkspecs:
      - "network_uuid={{ net_a.meta.uuid }},address=10.0.7.10,float=true"
      - "network_uuid={{ net_b.meta.uuid }}"
    state: present
    await: true
```

Run 1 creates it. Run 2 fails in 8 seconds with the 409 above. Immediately
afterwards the address is visibly haloed:

```json
{
  "address": "10.0.7.10",
  "reservation_type": "deletion-halo",
  "user_type": "unknown",
  "user_uuid": null,
  "reserved_at": 1788499555.7071402
}
```

A leftover instance in state `error` is also left behind, because the API
creates the instance object before reserving the address and calls
`enqueue_delete_due_error()` on failure.

### Suggested fix

The module knows it is replacing rather than creating, so it can wait the
halo out. Retrying the create on a 409 for up to `IP_DELETION_HALO_DURATION`
would make replacement work without any server change, and the existing
`await_timeout` budget is the natural place to spend that time.

Alternatively the server could let a create explicitly requesting a haloed
address take it -- the halo protects against surprising *random*
reallocation, which is not what an explicit `address=` is asking for.

### Versions

- Cluster: 0.8.0rc5.dev1189+g9483ab90c.d20260903
- Module: `shakenfist/deploy/collection/plugins/modules/sf_instance.py` at 9483ab90c

---
*(Investigated with Claude Code)*

mikalstill:
Deleting an instance places each of its addresses into a deletion-halo reservation for IP_DELETION_HALO_DURATION (300 seconds by default). An instance create explicitly requesting that address was refused with a 409 for the whole period, which made delete-and-recreate at a static address impossible: the sf_instance replacement path destroyed a working instance and could not bring it back for five minutes.

The halo exists to stop a recently released address being surprisingly handed out by *random* allocation, which is not a risk an explicit address request is exposed to. Add an evict_halo flag to the ReserveAddress RPC, threaded through mariadb.reserve_address() and IPAM.reserve(), and set it only on the instance-create path that reserves a caller-specified address. The takeover is a guarded UPDATE whose WHERE clause requires the existing row still be a deletion-halo, so a concurrent real reservation can never be overwritten; random allocation is unchanged and still respects the halo.

The ansible module CI scenario 004 previously changed the IP address on replacement specifically to step around this defect; it now keeps the same address, restoring coverage of replacement at a fixed address. A new functional test also covers immediate recreate-at-the-same-address via the REST API directly.

Plan alignment: docs/plans/PLAN-generic-allocator.md -- that plan deliberately leaves IPAM's allocation semantics unchanged, so this fix is inside IPAM's own scope and does not cut across it.

Fixes #4059

Triggered-By: issue-fix workflow, https://github.com/shakenfist/shakenfist/actions/runs/33922410353

Assisted-By: Claude Code


sfconductor:

The conductor will merge this at `3ddb7729aa69` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:

sfconductor:

Issue #4087, opened 2026-09-05
Every instance placement in a cluster's first minutes is unguarded, and the capacity reconciler then records a node at twice its own limit


mikalstill:
`admit_instance_placement()`'s guarded `UPDATE` is supposed to make
`used_cpus + requested <= limit_cpus` an invariant of
`scheduler_node_capacity`. The CI headroom probe recorded a row at
`used_cpus: 7` against `limit_cpus: 3`, with `cpu_measured: 0` and
`cpu_committed_row_present: true`.

The guard was not breached. It was never evaluated: the placements that
produced that number were admitted before the row existed, and the
reconciler's first pass then computed the counter from ground truth and
faithfully wrote the over-limit figure.

## What was measured

From the phase 2 baseline harvest (66 `merge_group` runs of
`functional-tests.yml`, 2026-08-30 to 2026-09-05, 204 job-runs with a usable
series; dataset in `docs/plans/data/ci-cloud-sizing-baseline/`):

* A capacity row above its own `limit_cpus` appears in **18 of 85,563**
  ledgered node-samples (0.021%), inside **2 of 204** job-runs (1.0%), both
  `Debian 12 tier` on the `slim-tier` topology.
* **All 18 of those node-samples are the first samples after the capacity
  rows first appear.** Both occurrences begin at sample index 12 of the job's
  series, which is the first sample taken after the reconciler's first pass.
* The node's `limit_cpus` did not move in either run (3.0 throughout, and the
  harvest records the ledger as the range it moved over).

| Run | Job | Node | Committed | Limit | Measured | Duration | Job outcome |
| --- | --- | --- | --- | --- | --- | --- | --- |
| [33752413862](https://github.com/shakenfist/shakenfist/actions/runs/33752413862) | Debian 12 tier | `750651c4` | 7, then 6 | 3 | 0, then 6 | 17 samples (~4 min) | **passed** |
| [33948911843](https://github.com/shakenfist/shakenfist/actions/runs/33948911843) | Debian 12 tier | `2060c55b` | 6 | 3 | 0 | 1 sample (15 s) | failed |

In the first of those the *measurement* also reached 6 on a node whose ledger
is 3, so this is not a phantom counter: 6 vCPU of instances really were
running on a node the scheduler would admit 3 onto.

## The mechanism

Five steps, all of them in the tree today.

**1. There are no capacity rows until the reconciler's first pass, and that
pass is not prompt.** `reconcile_scheduler_capacity` is registered as
[...]

PR #4089, opened 2026-09-05
Phase 2 of right-sizing the CI test clouds: the measured baseline

mikalstill:
This plan was written from three hand-collected data points and said so.
Phase 1 built an instrument; phase 2 harvests the 66 merge runs of its
output that were banked inside the artifact retention window, and replaces
every guessed figure in the master plan's *Situation* section with a
measured one.

**217 records over 55 merge runs, 204 with a usable committed-CPU series,
2026-08-30 to 2026-09-05.** The dataset is committed under
`docs/plans/data/ci-cloud-sizing-baseline/` (1004K) because the bundles it
came from expire ninety days after their merge run.

### What the measurement changed

**Utilisation does not predict failure.** Pooled, fuller job-runs fail more
— but that comparison is confounded (three of the four instrumented jobs
share a topology and the fourth is both the fullest and the worst), and it
is marked not-to-be-quoted rather than banked. Within each job, holding
topology and suite fixed, the cluster-wide committed fraction separates
passing from failing runs *not at all*: for `Debian 12 tier` the median p90
is 0.750 either way. What separates them, in every job, is whether the
scheduler ran out of candidates — **90%** of job-runs with a
`sufficient_idle_cpu` abort failed, against **15%** without. The plan's
central claim survives as "the cloud running out fails the job", not as
"how full it gets predicts failure".

**The cluster-wide figure hides the node that refuses.** The per-node
maximum exceeds it by a median of 2.25×; a node sits at its ledger in 65%
of `slim-primary` job-runs and 100% of tier ones. D21 is defended, with the
correction that the statistic *saturates* and so must be read as a goal
rather than an alarm.

**Only CPU ever refuses.** Memory, disk space and disk bandwidth recorded
zero drops in over 23,000 evaluations each, so phase 0's D5 narrows to its
operative clause — though the memory ledger is 3× physical and the tier's
p90 committed vRAM reaches 101% of it, which is not spare capacity.

**D7 closes at zero fallbacks** across 85,563 ledgered node-samples.

### What it falsified

- The master plan claimed `slim-primary` failures "contain no such refusal
  at all". It drops candidates in 110 of 154 job-runs and aborts in 14; it
  simply has other hypervisors to fall through to. Corrected in place.
- D17 originally named six instrumented jobs. Only four carry the probe —
  `Ansible modules` is gated out by `test_kind`, and `Node lifecycle` calls
  the composite action directly and never reaches the workflow the probe
  lives in. Checked against real artifacts, not reasoned about. `Node
  lifecycle` is the best performer in the failure table, so no
  utilisation/pass-rate claim is made across that table.
- D22's dataset size estimate was an order of magnitude low, twice.

### The defect this surfaced

`scheduler_node_capacity` has no rows until the capacity reconciler's first
pass, which `schedule.every(5).minutes` puts five minutes after the cluster
daemon starts — there is no pass at daemon start or at election. For a
measured **135 to 210 seconds of every one of the 204 job-runs**, every
admission takes P7's fail-open branch, and the CPU pre-filter sees only a
once-a-minute count of *running* domains. In **176 of 204 job-runs
instances were already running before any capacity row existed.** The first
pass then writes a node above its own `limit_cpus`, after which it refuses
every create while measuring idle — the #3772 signature without exhaustion,
and a defect a bigger cloud would mask rather than fix.

Filed as #4087. Not fixed here: this phase files, and phase 4 does not
depend on it being fixed.

### Also in this PR

- `/admin/resources` publishes `capacity_degraded` (D19), so a sampler can
  tell a failed capacity read from an unpopulated table.
- `ci_headroom_report.py` gains `summary_record()` and `--json` (D18), with
  the prose rendered from the record so the two cannot disagree. Verified
  byte-identical on stdout against a real bundle.
- `ci_headroom_harvest.py` (D16/D17), which fails loudly on a bundle name
  it does not recognise rather than skipping silently.

### Testing

176 unit tests pass; `pre-commit run --all-files` is green;
`check-plan-status.py` agrees. Every headline number above was reproduced
independently from the committed dataset before being written down, and the
contiguous-prefix and pre-row-running findings were re-derived over a random
sample of cached bundles.

No topology is proposed anywhere — phase 4 decides that and phase 3 gates
it. Steps 2g and 2h remain open by design: they need
shakenfist/actions#52 merged and real merge runs behind it.

https://claude.ai/code/session_014r9MQ3V3nqfpEx2H8g2nfy


sfconductor:

The conductor will merge this at `da1d4564c4d1` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `da1d4564c4d1`, bypassing the merge queue.

Issue #4097, opened 2026-09-05
VDI console tokens: close the uncovered cross-repository seams


mikalstill:
The phase 10 push audit of PLAN-kerbside-vdi-tokens ran a lens with no `PUSH-AUDIT.md` heading: it enumerated every value that crosses a repository boundary in this feature and asked, for each, *which test in which repository would fail if the writing side changed its shape*.

The motivation is that all three defects this plan shipped were correct on both sides of a seam and wrong across it. Each repository's tests mock its peer, and a mock is written from the same understanding that wrote the code, so it agrees with the code rather than with the peer. kerbside#201's own commit message says it: the bug "stayed latent because the unit-test fixture set both the instance's node and the node's name to 'n1', making them indistinguishable". shakenfist#4009 is the same shape — no test in any repository parsed a `.vv`.

The audit graded each seam **covered** (naming the test), **fixture-only**, or **uncovered**. Most of the JWT and scrape seams came back genuinely covered by kerbside's `sf-e2e-functional.yml` lane, which is the only mechanism in the four repositories that reads a real message from a peer. What follows is the remainder, with the cheapest fix for each. They are independent and can land separately.

### 1. `host_subject` fails open, and the e2e lane cannot see it — 4 lines (kerbside)

`sources/shakenfist.py` does `node.get('spice_server_cert_subject') or None`. If Shaken Fist renames or drops that field, the value becomes `None`, kerbside-proxy maps an empty subject to "no pinning", and **the lane still passes** — a live proxied session opens with backend TLS identity unpinned, which is the pre-fix state restored silently.

Add to `drive-happy-path.py`, beside the existing audit-row check:

```python
c = db.get_console(source, instance_uuid)
if not c.get('host_subject'):
    raise SystemExit('the scraped console carries no host_subject; '
                     'backend TLS pinning is inert')
```

Highest value-per-line in the audit. (The Shaken Fist half — a warning when the certificate subject cannot be rendered — landed with the phase 10 PR.)

### 2. Nothing checks that what Shaken Fist emits is what ryll accepts — ~25 lines (kerbside)

[...]

Issue #4096, opened 2026-09-05
/admin/vditokenpubkey reads the entire cluster_config table on every call


mikalstill:
`AdminVDITokenPublicKeyEndpoint.get()` → `vdi_tokens.get_signing_material()` → `mariadb.get_cluster_config()`, which is an uncached read of **every** `cluster_config` row.

To render four public fields (`kid`, `alg`, `public_pem`, `created`), the API worker materialises every cluster secret — `AUTH_SECRET_SEED`, the MariaDB password, the signing key's private PEM — in memory and across the gRPC channel. The endpoint is callable by any authenticated namespace (it mirrors `/admin/cacert`, deliberately, since the material it serves is public), so any holder of an `admin.read`-scoped or wildcard token can loop on it.

Nothing is disclosed — the response is built field by field from a fresh dict — so this is not a leak. It is an unnecessary per-request cost on the database tier and unnecessary transit of secrets that the caller has no business causing.

Fix: add a single-key `get_cluster_config_value(key_name)` to the three-layer client and use it here, so the endpoint reads one indexed row.

Found by the phase 10 cross-repository push audit of PLAN-kerbside-vdi-tokens (`docs/plans/PLAN-kerbside-vdi-tokens-phase-10-push-audit.md`). Graded advisory; recorded there with its disposition.

mikalstill:
Adding the security rationale, which the original filing omitted — it argued only the per-call cost.

The phase 10 audit's F1 finding stopped `KERBSIDE_JWT_SIGNING_KEY` being exported into every daemon's `os.environ` (and thus into every privexec child and every `/proc/<pid>/environ`). That fix is real but partial, and the reason is this issue: there is no keyed read. `mariadb.get_cluster_config()` and the `GetClusterConfig` RPC both return the entire table, so `load_cluster_config()` filters only *after* receipt — the unencrypted Ed25519 private PEM still crosses the wire to every daemon on every node at startup, and `daemons/cluster/main.py:660` pulls the whole table again each maintenance pass to read scheduled-task stamps.

So the residual exposure after F1 is: the key is in the address space of daemons that have no use for it, and on the wire to all of them. An attacker who can read a daemon's memory is not shut out; one who can read its environment now is.

[...]

Issue #4095, opened 2026-09-05
vdi_tokens: false race comment, and two paths that bypass SigningKeyError


mikalstill:
Three small defects in `shakenfist/util/vdi_tokens.py`.

**1. The race comment describes a guarantee the write does not provide.** `ensure_signing_key()`'s docstring says the row is re-read "so that a concurrent writer who won the race is the one whose material we return". `_direct_set_cluster_config` is an unconditional `INSERT ... ON DUPLICATE KEY UPDATE` upsert, so there is no winner — the **last** writer overwrites the first, and `ensure_signing_key` writes `{'active_kid': ..., 'keys': [new]}` rather than merging, so a concurrent create discards the loser's key entirely. The phase 1 plan specified an atomic create-if-absent; that was not implemented.

Reachability is narrow: `ensure_signing_key()` has exactly one caller in the tree, `sf-ctl`. Nothing on the API path calls it (`mint_console_token` raises `SigningKeyError` when material is absent, and `/admin/vditokenpubkey` returns 404), so this needs two operators running `sf-ctl` at once — which the operator guide already warns against. Consequence if it happens: a token minted with the losing key can never be verified, and the console open fails with `401 unknown signing key`. Availability only.

Fix: either correct the docstring to describe the upsert honestly, or add `INSERT ... ON DUPLICATE KEY UPDATE key_name = key_name`, whose rowcount 0 on collision gives the create-if-absent primitive the plan specified.

**2. `KeyError` escapes where `SigningKeyError` exists.** The module defines `SigningKeyError` for material that is "missing or internally inconsistent", and `active_signing_key` uses it for one inconsistency — but `active_signing_key` and `public_view` then index `material['active_kid']` and `material['keys']` directly. A stored row that is valid JSON but missing either key raises `KeyError` out of `AdminVDITokenPublicKeyEndpoint.get()` as a 500 with no useful message.

**3. A late import without a circular-import reason.** `import jwt` is deferred with the stated reason "to keep PyJWT off the module import path for callers that only touch the key helpers". That is a lazy-loading optimisation, not a cycle, and `CLAUDE.md` reserves late imports for cycles. PyJWT is already a hard dependency pulled in by flask-jwt-extended in every process that imports this module. Move it to the top. (The two `from shakenfist import mariadb` late imports are correct and carry proper reasons.)

[...]

Issue #4094, opened 2026-09-05
Direct .vv endpoint lacks the instance-state gate its proxy sibling has


mikalstill:
`InstanceVDIProxyConsoleHelperEndpoint.get()` gates on three things: `KERBSIDE_URL` configured (404), instance in `created` (406), and a SPICE `video['vdi']` (409). Its sibling `InstanceVDIConsoleHelperEndpoint.get()` — the direct `.vv` — gates on none of them, and `redirect_instance_request` returns early rather than erroring for an unplaced instance, so nothing upstream supplies the check.

`instance_from_db.ports` resolves through `_db_get_attribute`, which returns `{}` when unset. So for an instance in `error`, `deleted`, or any pre-`created` state that has a placement, the endpoint returns **HTTP 200 with a `.vv` containing `port=None`**. The user sees an opaque virt-viewer parse or connect failure where the proxy path would have said the instance is not ready.

Only the state gate is a genuine asymmetry — the VDI-type gate is correctly absent, since the direct path legitimately supports `type=vnc`. A small shared precondition helper would stop the two drifting again.

Second, smaller item in the same pair of endpoints: `InstanceVDIConsoleHelperEndpoint` declares `requires_admin=True` in its `swagger_helper`. That flag is documentation-only — it appends "Requires authentication as a member of the system namespace" to the published description — but the endpoint is ownership-gated, not admin-gated, and the proxy endpoint immediately below correctly omits it. The published OpenAPI therefore tells a namespace user their own console endpoint is admin-only.

Found by the phase 10 cross-repository push audit of PLAN-kerbside-vdi-tokens (`docs/plans/PLAN-kerbside-vdi-tokens-phase-10-push-audit.md`). Graded advisory; recorded there with its disposition.

PR #4099, opened 2026-09-06
Kerbside VDI tokens phase 10: push audit

mikalstill:
The final phase of [PLAN-kerbside-vdi-tokens](docs/plans/PLAN-kerbside-vdi-tokens.md): a retrospective `PUSH-AUDIT.md` run over the whole plan as one body of work, across **four repositories** and eleven merge ranges. Every pull request in that plan was reviewed inside its own repository; nobody had looked at the four sides as one system.

Eight agents ran: wave 1 in both halves, the four `PUSH-AUDIT.md` judgment headings, and a fifth lens this plan added because none of the template's headings covers a value crossing a repository boundary.

## It did not find nothing

**One blocking defect, in kerbside** (fixed in shakenfist/kerbside#412). A failed VDI signing-key fetch errored the Shaken Fist console source; an errored source is skipped *before* the scrape loop that would have kept its consoles alive; the cleanup afterwards runs unconditionally. Result: the source's entire console inventory deleted, breaking the direct and proxy console routes that have nothing to do with tokens. It triggers against any cluster that has not run `sf-ctl ensure-kerbside-signing-key`.

That makes it the **third** defect from the one configuration no phase ever tested — Kerbside integration not configured — after #4003 and #4009, and the first that destroys data. The plan's *Post-completion defects* section called that gap out. It was not two bugs in one repository; it was a class, and it spans repositories.

**One high-severity finding, in ryll** (fixed in shakenfist/ryll#358). Its SPICE TLS verifier trusted the full public WebPKI root set even when a `.vv` supplied a private cluster CA, and discarded hostname checking — so for a backend with no pinned subject, any publicly issued certificate completed the handshake. That code predates this plan and sits outside every audited range; it surfaced only because verifying "is `host_subject` pinning actually enforced?" means reading the code that enforces it.

## What this PR changes

- **The signing key is no longer exported into every daemon's environment.** `load_cluster_config()` copied *every* `cluster_config` row into `os.environ` in every process on every node, and privexec passes `env=None` so children inherit it. The row holds the unencrypted Ed25519 private PEM, so code execution as any daemon user on any node yielded the ability to forge a console token for any instance — undetectably, since kerbside verifies offline by design. A row now reaches the environment only if it is a declared `SFConfig` field, or does not match `SECRET_CONFIG_KEY_RE`. That ordering is load-bearing and commented: `AUTH_SECRET_SEED` matches the secret pattern via `_SEED$` but *is* declared and must keep being exported, so a naive "don't export secrets" filter would break authentication cluster-wide.
- **The `host_subject` fail-open is now discoverable.** `_spice_host_subject_from_cert` returned `None` — silently, with no log at all — if the certificate subject held any attribute outside its eight-name map, which is plausible cluster-wide on an operator's own PKI. It now warns, naming the unrenderable OID, throttled per reason per process. The fail-open itself is unchanged; it is documented and deliberate. Also folded in: `read_spice_server_cert_subject` warned on every failed read, and `observe_this_node` runs every 15 seconds from both sentinels, so one damaged certificate meant eight warnings a minute forever.
- **The `.vv` type-collapse regression test could not fail.** `test_vdi_console_file.py` was added to hold #4009's fix in place. Its `host` assertion is genuine — it validates against a live `get_nodes()` response, not a fixture. Its `type` assertion was hollow: the test leaves `video` unset, the server defaults `vdi` to `'spice'`, and the `startswith('spice')` collapse is a no-op for it. **Deleting that line left the test passing.** Added a test that creates `vdi='spiceconcurrent'` and asserts the internal enum never reaches the viewer, mutation-verified: with the collapse removed the new assertion fails `'spice' != 'spiceconcurrent'` while the old one still passes.
- **Two false claims in this plan, corrected at source.** I wrote that the plan changed no database schema (false: `node_attributes` v4 `ALTER`, plus kerbside's alembic revision) and that R1 added new `mariadb.py` functions and a counter (false: it extended the existing trio with one field). Both were checkable assertions made without checking; the agents falsified them because their briefs said to verify rather than assume.
- **Two documentation dedupes** that the consistency commit `f9117ca3d` missed.

## What was verified rather than believed

The management session spot-checked six findings against the tree and mutation-tested two fixes. Reinstating the kerbside bug fails exactly six tests and nothing else. Removing the `.vv` type collapse fails the new assertion and nothing else.

The audit also confirmed real strengths, which the Findings section records with the same weight: the cross-page documentation consistency check came back genuinely clean across all four repositories; the mint gate survived six attempted cross-namespace routes; kerbside reads no backend address from any JWT claim; the replay cache is bounded and reaped; ryll publishes via OIDC Trusted Publishers with attested provenance; and the phase-9 e2e lane is the strongest thing this plan built — the only mechanism in the four repositories that reads a real message from a peer.

## Follow-ups filed

#4093, #4094, #4095, #4096, #4097 here; shakenfist/kerbside#409–#411; shakenfist/client-python#392–#393; shakenfist/ryll#357.

#4097 is the seam package: for every value crossing a repository boundary, which test would fail if the writing side changed it. Several answers are "none", including the `.vv` `type=` key that #4009 got wrong. Four changes totalling under a hundred lines, all inside CI machinery that already exists, close the majority.

The recommendation the audit leaves behind is that the seam lens becomes a standing `PUSH-AUDIT.md` heading for cross-repository plans rather than a per-plan invention.

This completes the plan: 12 of 12.

https://claude.ai/code/session_01RbYd92imhYufuWps6ZfEyF


mikalstill:
All five items addressed in 0843b85dc. Both factual claims were checked against the tree before acting, and both held.

**1 (docs state the export rule incorrectly)** — correct, and a fair catch given this PR's own 2c heading reports the cross-page consistency check as clean. `_exportable_cluster_config_key()` exports a row if it is declared **or** does not match the secret pattern. Both pages now state it that way round, with the `AUTH_SECRET_SEED` ordering rationale kept in the developer doc and out of the operator guide, as suggested.

**2 (the mitigation is overstated)** — correct. There is no keyed read; `get_cluster_config()` and the `GetClusterConfig` RPC both return the whole table, so the private PEM still crosses the wire to every daemon at startup and to the cluster daemon each maintenance pass. Both pages now claim only what is true — the key no longer persists in any process's environment and is not inherited by privexec children — and name the residue explicitly. #4096 has been updated to carry the security rationale alongside the per-call cost one.

**3 (status ahead of the evidence)** — correct, and it caught this plan violating its own decision 7. Phase 10 is back to In progress and the index to 11 of 12. The Disposition table now names shakenfist/kerbside#412 and shakenfist/ryll#358 rather than branch names, and records that neither a branch nor an open PR satisfies the definition of done. Status flips to Complete when both merge.

**4 (the type-collapse claim is wrong)** — correct, and this is the one I am most glad you caught, because it is an overstatement in an audit whose whole purpose is recording what is falsifiable. `test_instance_vdiconsolehelper.py` was added by `d36298a43`, the #4009 fix itself, so the unit coverage pre-existed and does fail under mutation. Only the functional assertion was hollow. Reworded in both plans to say the collapse had unit coverage but no functional coverage, and that the mutation result is what pins the scope of the finding.

**5 (warn-once never cleared)** — taken. `_SPICE_SUBJECT_WARNED` is cleared whenever a read succeeds, so the throttle covers one episode rather than the life of the process. Your scenario is the one that matters: a reissued certificate that later regresses to the same cause would otherwise be silent until a restart, swallowing exactly the "pinning has just been disabled for this node" event the warning exists to report. Added `test_a_repaired_then_broken_certificate_warns_again`, mutation-tested — with the clear removed it fails `2 != 1` and nothing else does.

`pre-commit run --all-files` and `tox` green.

Two of the four fix items were errors in the audit write-up rather than in the code, which seems like the right result for a phase about claims that were never checked.

mikalstill:
Round two addressed in 0f5d7afb2. **1 fix, 1 document, 4 consider, 2 info** — down from 4 fix. One item I have pushed back on with evidence.

**1 (empty subject renders as `''`)** — real, and the sharpest kind of finding: it is incomplete generalisation from *last round's* fix. `','.join([])` is `''`, which is not `None`, so `read_spice_server_cert_subject()` counted it as success — and the clear-on-success I added last round then wiped the warn-once state, actively masking a genuine warning reported moments earlier. Empty subjects now warn (`empty-subject`) and return `None` like the other two paths. Two tests added, including one that pins the masking behaviour specifically; mutation-tested, both fail without the guard and nothing else does.

**3 (release notes)** — added. Agreed this is cluster-wide operator-visible behaviour, not just a Kerbside detail; the bullet spells out the pattern, says declared options and non-secret-named rows are unaffected, and links the operator guide.

**5 (docs understate when the warning repeats)** — taken, and it is the same class as round one's item 1: prose this PR added promising more than the code does. Now "once per episode", with the repaired-then-regressed case named. Also documented the new empty-subject case while there.

**6 (index cell carries volatile issue numbers)** — taken. You are right that `check-plan-status.py` validates the term and the arithmetic but not the prose, so those numbers would have gone stale with nothing to notice. Trimmed to description plus phase link; the Disposition table already holds them.

**2 (withheld rows dropped with no log line)** — taken as a comment rather than a log. `load_cluster_config()` runs at import time, before logging is configured, which is exactly why there is no `LOG` in scope — so a debug line would need a bare `logging.getLogger` call at import time in every process. Both branches now say why they are silent and name `sf-ctl show-config` as the diagnosis path.

**4 (private `ObjectIdentifier._name`)** — taking your stated alternative rather than the primary suggestion, because I checked and the public API is worse here. On cryptography 50.0.1, `NameAttribute.rfc4514_attribute_name` returns `2.5.4.5` for `serialNumber` — the dotted string, which is already logged as the `oid` field. Switching would lose the human name the warning exists to give an operator and duplicate a value we already have. Recorded as a deliberate private-API choice, with that evidence, noting the getattr guard and the test that pins it.

**9 (no end-to-end assertion the key is absent from a live environ)** — agreed, and it is a nice turn: the seam lens applied to the fix the seam lens produced. The unit tests assert against a mocked `os.environ` populated by the function under test, which is the fixture-only shape this PR's audit names as the root cause of the plan's shipped defects. It needs to read `/proc` on a cluster node, so it belongs with the node-exec machinery rather than in this PR — added to #4097 with the proposed check.

**7, 8** — informational, no action. Agreed on both; the two callers are separate single-threaded daemon processes, and the second minimal instance buys an assertion that actually exercises the collapse.

`pre-commit run --all-files` green.

**Convergence:** fix items 4 → 1, and the single fix was in machinery added last round rather than in the original change. That is the signal to stop after this one: if round three returns only `consider` items I intend to land and let the auto-filed issues carry them.

mikalstill:
Round three addressed in `7bc2c44c1`. Triage was **1 fix, 1 document, 9 consider, 2 info**.

## Fixed

**1 — the docstring naming "an organisation".** Confirmed and corrected. `NameOID.ORGANIZATION_NAME` is in `_SPICE_SUBJECT_SHORT_NAMES` mapped to `O`, so an organisation-qualified subject renders fine and never reaches the fail-open. The reviewer is right that this contradicted the operator guide added on the same branch. Now reads "a serialNumber, a dnQualifier or a custom OID".

**5 — the `'Unknown OID'` placeholder.** Verified against the installed cryptography 50.0.1: `ObjectIdentifier('1.3.6.1.4.1.99999.1')._name` returns the literal string `'Unknown OID'`, which is truthy, so `or 'unknown'` never fired. Taken rather than deferred, because the structured `attribute` field is the one a Loki query groups on and a constant string that looks like a name is worse than a null. A named accessor maps it back to `None`; the message falls back to the dotted string. The comment claiming the read "degrades to unknown" was wrong for the same reason and is corrected.

This one came with a test, and the test was mutation-verified: removing the mapping produces exactly one failure.

**3 — the kerbside defect missing from our own operator docs.** Taken, and the situation moved while the review was being written. kerbside#412 **merged as `e2a493ea6` at 05:29Z**, fifteen minutes before this review posted, carrying both F-B1 and the F5 `audit_events` write. It is *not* in a tagged release — v0.5.0 and earlier are still affected — so the guidance is version-accurate rather than pointing at an issue tracker: provision the signing key before rolling daemons on a cluster Kerbside already scrapes. Added to `docs/operator_guide/vdi_console_tokens.md` as a warning admonition and to `docs/release_notes/v07-v08.md`. The audit's disposition table now records the merge reference.

**6 — the concurrent path asserting only `type`.** Taken. The reviewer's point that this "repeats the narrowness the seam lens found" is the correct reading of it. The tls-port/host-subject assertion is factored into `_assert_tls_port_is_pinned()` and called from both tests.

**10, 11** — `dict[str, Optional[str]]`, and the blank line that was silently reflowing a published release-notes list from tight to loose. Both taken.

## Verified, no change needed

**4 — first caller to pass `video=`.** Checked rather than left to the merge queue: `video=None` is a keyword on `create_instance()` in the released **v0.8.3** (`apiclient.py:555`) and is placed into the JSON body verbatim at line 567. It has been there since the package was first named. No signature or serialisation surprise waiting in the merge queue.

## Declined

**2, 7, 8** — operator-visible signalling for withheld rows, case normalisation of config names, and a `verify-config` listing of what would be withheld. These are all real, and they are all a feature in `sf-ctl` rather than part of an audit of an already-merged plan. They ride on their auto-filed issues. Item 7 the review itself notes is pre-existing behaviour of the redaction mechanism, not something this change regresses.

**9, 12** — informational. On 12: the blocking finding is no longer open, per the merge above. ryll#358 remains open, which is why this plan stays at *In progress* / 11 of 12 rather than Complete.

## Convergence — and why I am not asking for round four

Fix items went **4 → 1 → 1**. But the more useful number is where they came from: every fix and document item this round was about prose or machinery *this pull request added in an earlier round* — the docstring came from round two, the `'Unknown OID'` fallback came from round two, and item 6 is about the test round one asked for. None of round three touched the change the pull request actually exists to make.

That is the generator loop the skill describes: the reviewer is reviewing its own previous instructions, and each round of fixes supplies the next round's material. Two of round two's items had the same shape, so this is a trend rather than one round's noise.

So I am stopping here rather than requesting another pass. The remaining nine `consider` items are auto-filed and can be judged on their own merits, outside a loop that is now largely self-feeding. My recommendation is to land this once CI is green.

Issue #4100, opened 2026-09-06
An explicit 'thin: false' on snapshot is indistinguishable from omitting it


mikalstill:
`InstanceSnapshotEndpoint.post` (`shakenfist/external_api/snapshot.py:60`)
tests `thin` for falsiness rather than for `None`:

```python
if not thin:
    thin = config.SNAPSHOTS_DEFAULT_TO_THIN
```

so an explicit `thin: false` is treated identically to omitting the key, and
`SNAPSHOTS_DEFAULT_TO_THIN` wins either way. A caller cannot ask for a
non-thin snapshot on a cluster configured to default to thin.

The reason is real and is not the handler's fault: `shakenfist_client`
transmits the key unconditionally (`apiclient.py:706`,
`data={'all': all, 'device': device, 'thin': thin}`, with the CLI flag
defaulting to `False`), so honouring an explicit `false` today would make
`SNAPSHOTS_DEFAULT_TO_THIN` inert for every shipped client.

## The recorded blocker is wrong

Two comments say this is waiting on phase 4 of
[PLAN-api-input-validation](docs/plans/PLAN-api-input-validation.md) — the
handler comment at `snapshot.py:66` and the note above `UNDECLARED_BY_DESIGN`
in `shakenfist/tests/external_api/test_parameter_declarations.py:43`.

Phase 4 does not unblock it. The request validation layer is *check-only*:
`validation.check()` returns findings and `validate_request` calls through,
never coercing or injecting, so the handler still receives `thin=False` from
`log_request`'s body merge in every mode. Nor is a schema layer needed to draw
the distinction at all — `'thin' in flask.request.json` draws it today.

The only real blocker is the client. Resolving this needs, in order:

1. A `shakenfist_client` release which omits `thin` from the request body when
   the caller did not ask for it.
2. An API capability token (`API_CAPABILITIES` in
   `shakenfist/external_api/app.py:309`) so the server can tell a client that
   omits the key from one that does not — otherwise honouring `false` still
   breaks every older client.
3. Then the handler can test `is None`, and the declaration's description can
   stop saying "False is currently treated the same as omitting the
   parameter".

Found while implementing phase 4 of PLAN-api-input-validation, which corrects
both comments to point here.

Issue #4098, opened 2026-09-06
Generate the four hand-authored use_kwargs schemas from their declarations


mikalstill:
Four endpoints hand-author a `get_args` marshmallow schema and bind it with
`@use_kwargs`, alongside the `swagger_helper()` parameter declaration that
describes the same parameters:

* `shakenfist/external_api/blob.py:178` — `offset`, `limit` (location `query`)
* `shakenfist/external_api/artifact.py:1015` — location `json_or_query`
* `shakenfist/external_api/instance.py:2080` — `all` (location `json_or_query`)
* `shakenfist/external_api/network.py:784` — location `json_or_query`

That is two sources of truth for the same parameters, and decision D2 of
[PLAN-api-input-validation](docs/plans/PLAN-api-input-validation.md) noted at
the time that "the four endpoints that hand-author schemas today are exactly
the ones whose documentation has drifted".

**The obvious fix is wrong.** The master plan's phase 4 line originally read
"fold the hand-authored `get_args` schemas into the compiled path", which
reads as "delete the four `@use_kwargs` decorators and let request validation
take over". It cannot: `validation.check()` returns findings and
`validate_request` calls through, so the compiled path never coerces or
injects anything, and `log_request` merges only the JSON body — nothing merges
`flask.request.args`. `@use_kwargs` is therefore the sole mechanism by which a
query-string parameter reaches a handler at all. Deleting it from `blob.py`
would silently revert `offset` and `limit` to their signature defaults,
reading a whole blob where the caller asked for a byte range.

**The defensible fix** is to keep `@use_kwargs` and generate its schema from
the same declaration list `swagger_helper()` receives: hoist the parameter
list to a module-level name, pass it to both, and build the marshmallow schema
from the shared token mapping in `shakenfist/external_api/validation.py`.

Two things make that more than a refactor, which is why it was deferred out of
phase 4 rather than done alongside enforcement:

* The declaration vocabulary has no element for a default, so the generated
  schema has to take `load_default` from the handler's signature at runtime.
* `blob.py`'s hand-rolled negative-offset check would move into the field,
  changing the error message callers see.

Deferred from phase 4 of PLAN-api-input-validation (decision D19), which
judged that two request-visible changes in the phase that flips enforcement on
was the shape that gave phase 3 four review rounds. Phase 6 is a plausible
home, since it already touches how values reach handlers.

Enforcement does not depend on this: the compiled check runs first and answers
[...]

PR #4101, opened 2026-09-06
API input validation phase 4: declaration fixes ahead of enforcement

### LINKED ISSUE #3739: Ref-resolving decorators consume an undeclared 'namespace' body parameter
Found by the phase 3 warn-only measurement of PLAN-api-input-validation
(docs/plans/PLAN-api-input-validation-phase-03-compile-and-warn.md), in
the functional CI run https://github.com/shakenfist/shakenfist/actions/runs/31691743944.

Nine findings of this signature, from three cluster suites:

    validation-reason=unknown-parameter validation-parameter=namespace
    method=GET route=/artifacts/<artifact_ref>
    validation-response-status=200 (and 404)

An undeclared body key normally reaches the handler and raises
TypeError (a 400 carrying interpreter text). These requests *succeeded*,
which is how the measurement noticed that three decorator families pop
`namespace` out of kwargs before the handler ever sees it:

- `_resolve_artifact_ref` — shakenfist/external_api/artifact.py:57
- `arg_is_instance_ref` — shakenfist/external_api/base.py:770
- `arg_is_network_ref` — shakenfist/external_api/base.py:896

So every route carrying any of these decorators accepts a functional
`namespace` body parameter (cross-namespace name disambiguation via
`resolve_lookup_namespace`, used by the shared-artifact and trust
machinery) which appears nowhere in the published OpenAPI and is
unknown to the compiled validation schemas. The CI caller that
triggered the findings is legitimate use:
`shakenfist_ci/cluster_ci_tests/test_artifacts.py:678` sends
`data={'namespace': ...}` on `GET /artifacts/shared-name`.

Phase 1's declaration audit missed it because the audit derives
parameters from handler signatures and route tables; a kwarg consumed
by a decorator is invisible to both. This is the same blind-spot class
as the `derived_location()` defect recorded in the phase 3 outcome.

Why it matters: phase 4 turns unknown-parameter findings into
rejections. Without declaring `namespace` on these routes first,
enforcement breaks the shared-artifact lookup path — the exact
working-caller breakage the warn window exists to prevent.

The fix is to declare `namespace` (body, `namespacename`, optional) on
every route carrying one of the three decorators, which also publishes
the cross-namespace lookup feature in the OpenAPI for the first time.
Note the coupling: `test_openapi_spec.py`'s STRUCTURED_PARAMETERS
completeness check and the phase 3 derivation generator both react to
declaration changes, and the declaration must describe what
`resolve_lookup_namespace` actually accepts per object type.

Labelled `automated-fix-attempted` because this is owned by
PLAN-api-input-validation (it gates phase 4) and needs the audit/spec
coupling handled with it, not a same-day patch.

mikalstill:
Phase 4 of [PLAN-api-input-validation](docs/plans/PLAN-api-input-validation.md),
up to its measurement gate. This lands the declaration fixes that
enforcement depends on; **it does not turn enforcement on** —
`API_VALIDATION_MODE` still defaults to `warn` and no response changes.

Phase 3 ran warn-only validation on sfcbr for eight days and one full
functional CI run. Every finding was explained, and two of the five
signatures were declaration bugs rather than intended rejections. Both
would have become a 400 for a working caller the moment enforcement is
on, so both are fixed here, and the phase then pauses to watch them go
quiet before the switch is thrown.

## The undeclared `namespace` (fixes #3739)

Four decorators — `arg_is_instance_ref`, `arg_is_network_ref`,
`arg_is_artifact_ref` and `arg_is_visible_artifact_ref` — pop a
`namespace` key out of kwargs before the handler runs and resolve the
object reference with it. The warn window saw nine findings against one
route family. The real exposure is **55 handler methods**, of which 51
declared nothing at all, and the shipped client sends the key from
`get_instance`, `delete_instance`, `get_artifact`, `get_network` and
`delete_network`. Enforcing without this would have refused the official
client's own cross-namespace lookups.

The declarations say what the parameter does on each route family rather
than repeating one string 55 times, because the semantics genuinely
differ: the artifact read routes widen an unqualified name across shared
and trusted artifacts where the write routes deliberately do not, and the
floating network belongs to no namespace and so is reachable only when
the parameter is omitted.

**Declaring them is not enough on its own, because nothing would keep it
true.** The audit derives a handler's parameters from its signature, its
`use_kwargs` schema and its `flask.request.args` reads, and a kwarg a
decorator removes appears in none of the three — the defect was invisible
by construction rather than by oversight. So the derivation gains a fifth
source: `decorator_kwargs()` resolves each decorator to its definition in
the package and collects the literal keys it removes from kwargs. A
consumed parameter nothing declares is now an audit failure.

That check was proven to fire rather than assumed to. With the derivation
in place and the declarations reverted it reports 51 problems and nothing
else; with them applied it reports none.

## Metadata values are not strings

On 2026-08-17 a tenant stored structured state under four namespace
metadata keys: 21 findings carrying a dict and one a list, against a
declaration of `string`. Every request answered 200, because the handler
passes the value to `add_metadata_key()`, which serialises whatever it is
given. There are **fourteen** such declarations and only one was
exercised by traffic during the window.

Neither `dict` nor an `arrayof*` token is wide enough, since callers
store both shapes under the same parameter, so the vocabulary gains an
`any` token: no `type` key, a `format` annotation, and therefore a
published schema which constrains nothing while still saying what the
parameter is.

This is the one token that deliberately validates nothing, so two things
hold it in place. A test pins the exact set of fourteen declarations
using it. And the specification test's completeness derivation — which
asked whether a schema was an object or an array — now also fires on a
schema with no `type`, because otherwise the widest token available would
have been the one type change that table could never catch.

## Deliberately not here

* **Enforcement.** Decision D22 gates it behind a confirmatory reading:
  one functional CI run with zero `namespace` findings, plus 72 hours in
  which the traffic that produced the metadata signature has actually run
  and did not produce it. An absent signature whose generator did not run
  is not evidence — phase 3's log records three separate occasions where
  a quiet channel was mistaken for a quiet system.
* **The `get_args` fold** (#4098). The master plan asked to "fold the
  hand-authored schemas into the compiled path". Read as "delete the four
  `@use_kwargs` decorators" that is a defect: the compiled path is
  check-only and nothing merges `flask.request.args`, so `@use_kwargs` is
  the sole mechanism by which a query parameter reaches a handler.
  Removing it from `blob.py` would read a whole blob where the caller
  asked for a byte range. The master plan is corrected at source.
* **`thin: false` on snapshots** (#4100). Two comments in the tree said
  this waits on phase 4. It does not — validation never injects, so the
  handler sees `thin=False` whatever the mode. The blocker is a client
  which sends the key unconditionally. Both comments now say so.

## Also found

* A mutation in `tools/check-api-declaration-guards.sh` was silently
  defanged by this change — it deleted a `swag_from` by pasting the
  decorator's text, which stopped matching — so it became a no-op while
  still reporting success. Rewritten to bound the deletion by line.
* `requires_namespace_exist_if_specified` is dead on two routes, sitting
  after the ref decorator that already popped what it reads.
  Pre-existing, recorded, untouched.

## Verification

`pre-commit run --all-files` clean at every commit. The guard mutation
harness catches all 34 mutations including three added here. The
published specification still validates with zero errors.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -362,8 +362,10 @@ does not start**. The rules, in full in
 3. A `path` parameter must be `required=True`.
 4. A raw request body is declared as `api_base.RAW_BODY_PARAMETER`, and
    cannot be combined with named body parameters.
-5. Every kwarg the handler accepts is declared; decorator-injected
-   `*_from_db` objects are not parameters.
+5. Every kwarg the handler accepts is declared, including one a
+   decorator pops out of `kwargs` before the handler runs (the ref
+   decorators' `namespace`); decorator-injected `*_from_db` objects
+   are not parameters.
 6. The type is a token from `api_base.ARGTYPES`. Objects and arrays of
    objects can only be declared in the body, since outside one there is
    no schema object to nest a structure in.
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -313,7 +313,7 @@ strings, because semantic validation of them is not built yet. Only
 ## What is not checked yet
 
 Enforcement is off, so a correct declaration still does not stop a
-caller sending something else. Three known gaps in the derivation
+caller sending something else. Five known gaps in the derivation
 itself:
 
 (The published specification itself *is* checked:
@@ -347,3 +347,28 @@ the type vocabulary landed.)
   every unfound name would report the whole API and mean nothing.
   Within a decorator this *can* resolve, an unreadable pop key or an
   unfollowable delegation is reported rather than skipped.
+* A decorator that *reads* a parameter without removing it stays
+  invisible. The derivation recognises `kwargs.pop` and
+  `del kwargs[...]` because those are what make a kwarg the handler
+  cannot receive; a decorator that acts on `kwargs.get('x')`,
+  `'x' in kwargs` or `kwargs['x']` and leaves the key in place is not
+  seen, so nothing requires `x` to be declared. There is no example in
+  the tree: `_resolve_artifact_ref` had the only one, an
+  `if 'artifact_uuid' in kwargs:` branch which resolved an artifact
+  from a body key that no handler accepts — so the call after it
+  raised `TypeError` and answered 400 with interpreter text. It was
+  deleted rather than declared, since declaring it would have
+  published a parameter that never worked, and
+  `test_artifact_uuid_is_not_a_parameter` pins the property that made
+  it dead. A new one would have to be added deliberately.
+* A handler cannot both take a raw request body and sit behind a ref
+  resolving decorator. `swagger_helper()` refuses
+  `RAW_BODY_PARAMETER` combined with any named body parameter, while
+  the audit requires a decorator-consumed kwarg such as `namespace` to
+  be declared in the body — so such a handler could satisfy neither
+  rule: declaring the parameter aborts sf-api at import, omitting it
+  fails CI. No handler in the tree has both, and nothing detects the
+  combination; it is recorded here so the first author to hit the
+  contradiction does not have to derive it. The fix, if it is ever
+  needed, is a rendering for "a raw body plus these named
+  parameters", not an exemption from either rule.
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -168,11 +168,20 @@ are listed in `UNDOCUMENTED_BY_DESIGN` in
 | a segment of the mounted route in `app.py` | `path` |
 | in a schema bound with `@use_kwargs(..., location='query')` | `query` |
 | read from `flask.request.args` | `query` |
+| popped out of kwargs by one of the handler's decorators | `body` |
 | anything else | `body` |
 
-That last row is the default because `log_request` merges the JSON
-request body into the handler's kwargs, and the official client sends
-a JSON body for every method including `GET`.
+The last two rows are the same default, because `log_request` merges
+the JSON request body into the handler's kwargs and the official client
+sends a JSON body for every method including `GET`. The decorator row
+is called out because it is the one source that reaches neither the
+signature nor any of the rows above: `arg_is_instance_ref`,
+`arg_is_network_ref` and the two artifact ref decorators each do
+`kwargs.pop('namespace', None)` and resolve the lookup with it, so
+`namespace` is a functional parameter of all 55 routes behind them
+while being invisible to a derivation that reads only signatures. It
+went undeclared for years on that account
+([#3739](https://github.com/shakenfist/shakenfist/issues/3739)).
 
 **3. A `path` parameter must be `required=True`.** The route cannot
 match without it, and an optional path parameter makes the published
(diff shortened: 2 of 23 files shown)


mikalstill:
All seven actionable items from the first round are addressed in
7991656f5. Both `fix` items turned out to be the same fault: a check
that reported success without meaning it.

- **Mutation 32** matched the namespace tuple by its nine-space
  indentation, which is a substring of the twelve-space declaration in
  the handler above, so the splice removed that handler's remaining
  parameters, responses, five decorators and body. It still parsed, so
  the harness reported a catch of "a handler vanished". Re-anchored,
  with assertions that refuse the write rather than corrupt the tree.
- **`any` compiled through the unrecognised-type fallback**, whose only
  output is a warning that the specification contains a type the
  compiler does not know. sf-api logged fourteen at every start. The
  rendering now lives in `validation.ANY_VALUE_FORMAT` and `ARGTYPES`
  builds from it; measured after, zero warnings across 139 schemas.

On the `artifact_uuid` suggestion: it is deleted rather than declared.
No handler in `external_api` accepts that kwarg and no route mounts
one, so the branch resolved an artifact and then raised `TypeError`,
answering 400 with interpreter text. Declaring it would have published
a parameter that never worked.

The four namespace descriptions are hoisted into `api_base` with a test
asserting which constant pairs with which decorator, which took ~145
lines out of the endpoint modules.

mikalstill:
All ten items from the second round are addressed in 5168935b8.

- **Item 1 (fix)** was round one's own shadow. Deleting
  `_resolve_artifact_ref`'s dead `artifact_uuid` branch left
  `test_artifact_uuid_branch_tenant_foreign_namespace_rejected` passing on
  a 404 raised by `resolve_lookup_namespace()` before any lookup, with a
  vacuous `from_db.assert_not_called()`. Replaced with three tests over
  `arg_is_visible_artifact_ref` -- the half of that function the
  parametrised cases cannot reach. Both lookups are patched in each, so
  the assertion is which one ran; mutating `if widen and not
  body_namespace:` to `if widen:` fails exactly the second and to `if
  False:` exactly the first. That also takes item 10's unit-test option,
  so the artifact path is no longer the least-covered part of the branch.
- **Items 5 and 7 (doc)** landed here rather than at step 5: the `any`
  token joins the type-vocabulary paragraph and the body-only sentence,
  and the release note says that fifty-five GET/DELETE routes begin
  publishing a request body and the metadata `value` stops being
  published as a string. Neither is a server behaviour change, but both
  are visible to anyone regenerating a client, and shipping the diff a
  release ahead of the note explaining it had nothing behind it.
- **Item 4** was a real inconsistency, so it was taken as machinery:
  `UNDECLARED_BY_DESIGN` no longer exempts a decorator-consumed kwarg.
  `test_accepted_parameters_are_declared` honoured it while `audit()` --
  behind both the derivability test and the pre-commit fixer -- has no
  exemption path, so an entry would have satisfied one guard and failed
  the other with neither message explaining why. The exemption is for a
  kwarg the signature names and the handler ignores; one a decorator pops
  is one a caller can send and act on.
- **Item 6** is pinned by two cases: a `minimum` and a `pattern` on the
  typeless token are each refused at import time, and relaxing either
  check in `_validated_constraints` fails exactly one of them.
- **Items 2 and 3** were taken as message and documentation edits rather
  than machinery. Both shapes are fail-closed today and neither exists in
  the tree; an unfollowable delegation now names its remedy instead of
  only its symptom, and the bare-name blind spot joins the gap list.

Verification on the pushed head: full unit suite 4229 passed / 0 failed,
`pre-commit run --all-files` clean, and the guard harness catches all 34
mutations.

Issue #3056, opened 2026-03-08
Consistency: GitHub security settings


mikalstill:
This project is not yet compliant with the **GitHub security settings** consistency audit.

**Audit spec:** [development/audits/github-security.md](https://github.com/shakenfist/development/blob/main/audits/github-security.md)

### What needs to be done

- Enable secret scanning and push protection

mikalstill:
Enabled on the repository:

```
$ gh api -X PATCH repos/shakenfist/shakenfist --input - --jq '.security_and_analysis'
{"dependabot_security_updates":{"status":"enabled"},
 "secret_scanning":{"status":"enabled"},
 "secret_scanning_non_provider_patterns":{"status":"disabled"},
 "secret_scanning_push_protection":{"status":"enabled"},
 "secret_scanning_validity_checks":{"status":"disabled"}}
```

`.github/workflows/codeql-analysis.yml` was already present with the job-level permissions the audit asks for, and `allow_auto_merge` and `delete_branch_on_merge` were already true, so secret scanning and its push protection were the whole of the gap.

Re-running `development/scripts/audit-check.py` against this repository now reports `[pass] github-security: Security settings and CodeQL are compliant`.

Closing by hand rather than from a pull request: `security_and_analysis` is not part of `.github/exported-config/`, so there is no in-repo change to carry a `Fixes #3056`.

*(Assisted by Claude Code)*

PR #4107, opened 2026-09-06
Clear four consistency audit findings.

### LINKED ISSUE #3057: Consistency: Workflow standards (flake8wrap)
This project is not yet compliant with the **Workflow standards (flake8wrap)** consistency audit.

**Audit spec:** [development/audits/workflow-standards.md](https://github.com/shakenfist/development/blob/main/audits/workflow-standards.md)

### What needs to be done

- Add `shellcheck disable=SC2086` directive to `tools/flake8wrap.sh` with explanatory comment

### LINKED ISSUE #3314: Consistency: CI review automation
This project is not yet compliant with the **CI review automation** consistency audit.

**Audit spec:** [development/audits/ci-review-automation.md](https://github.com/shakenfist/development/blob/main/audits/ci-review-automation.md)
**Template:** [development/templates/ci-review-automation/README.md](https://github.com/shakenfist/development/blob/main/templates/ci-review-automation/README.md)

### Automated check details

Missing pr-retest.yml

---
*This issue was created automatically by the consistency audit workflow.*

### LINKED ISSUE #3418: Consistency: Workflow standards (devpi cache fallback)
This project is not yet compliant with the **Workflow standards (devpi cache fallback)** consistency audit.

**Audit spec:** [development/audits/workflow-standards.md](https://github.com/shakenfist/development/blob/main/audits/workflow-standards.md)

### Automated check details

9 devpi-backed env block(s) missing a PIP_EXTRA_INDEX_URL pypi fallback: code-formatting.yml:27, codeql-analysis.yml:20, docs-tests.yml:19, functional-tests.yml:26, pin-indirect-dependencies.yml:29, publish-website.yml:17, release.yml:26, scheduled-tests.yml:24, test-drift-fix.yml:69. Add "PIP_EXTRA_INDEX_URL: https://pypi.org/simple/" alongside PIP_INDEX_URL so a devpi cold-cache miss (empty index for a first-touch package) falls back to pypi instead of failing with "from versions: none"

---
*This issue was created automatically by the consistency audit workflow.*

mikalstill:
Clears four open consistency audit issues, one commit each.

| Commit | Issue | Change |
|---|---|---|
| `69760968b` | #3057 | `tools/flake8wrap.sh` gains the `# shellcheck disable=SC2086` directive and a comment saying the word splitting is deliberate |
| `96be372ca` | #3418 | `PIP_EXTRA_INDEX_URL: https://pypi.org/simple/` added to ten devpi-backed `env` blocks |
| `109968895` | #3314 | Adds `pr-retest.yml`, replaces the hand-rolled `pr-re-review.yml` with the current template, and deletes the retired comment addresser |

#3056 is not represented here. The `github-security` criterion reads the
live GitHub API and `security_and_analysis` is not part of
`.github/exported-config/`, so there was no in-repo change to carry a
`Fixes` line. Secret scanning and its push protection have been enabled
on the repository directly and the issue is closed, with the evidence in
[a comment on it](https://github.com/shakenfist/shakenfist/issues/3056).

## Two findings were wider than their issue text

**#3314** was filed as "Missing pr-retest.yml". Running
`development/scripts/audit-check.py` against the tree shows the
criterion also fails on two more counts, so all three are fixed here --
the issue is "not compliant with this audit", not "missing one file".

The one that matters is `pr-re-review.yml`. It open-coded the trigger
handling in about thirty lines of inline shell rather than calling
`shakenfist/actions/pr-bot-trigger@main`, and the copy had drifted from
what it was copied from: `+1` rather than `rocket`, a differently worded
refusal, and no trigger-phrase check of its own, so it could not tell
"phrase not matched" from "not authorized". More to the point, once
`pr-bot-trigger` grew a fork pull request guard the inline copy did not
inherit it. `pr-ref` is `.head.ref` -- a branch name carrying nothing to
say which repository it belongs to -- and fork pull requests are commonly
opened from the fork's default branch, so it is literally `main`. It also
held a scarce `claude-code` runner for every comment containing the
phrase including unauthorised ones, checked out the default branch rather
than the pull request's merge ref (so a re-review judged a change against
the conventions as they were *before* it), had no concurrency group, and
left a write-scoped token in the `.git/config` of the tree Claude Code is
then pointed at with `--dangerously-skip-permissions`. The template fixes
all of that and is copied in verbatim.

The retired comment addresser goes too -- `pr-address-comments.yml`,
`address-comments-with-claude.sh`, `render-review.py` and
`review-schema.json`, all four in one commit, because deleting the
workflow and keeping the scripts leaves exactly the copy that propagates
to the next project. It went unused, and it is not inert: it triggers on
`issue_comment`, so it holds `contents: write` on the pull request branch
for a feature nobody wants.

**#3418** named nine `env` blocks. There are ten now:
`merge-failure-triage.yml` and `issue-fix.yml` have been added since the
issue was filed, and `pin-indirect-dependencies.yml` was fixed
separately. That file is the pattern the other ten now match.

## Fallout repaired

`docs/developer_guide/ci.md` loses the addresser's workflow row and bot
command and gains `pr-retest.yml`'s. `pyproject.toml`'s `jsonschema`
declaration was justified by `render-review.py` importing it directly;
that role passes to `tools/merge-triage.py`, whose own comment claiming
`jsonschema` is not a runtime dependency is corrected to say what it
actually means -- that shakenfist may not be installed where it runs.
Every remaining mention in the tree is under `docs/components/`, which
`sync-external-docs.yml` imports from the sibling repositories.

## Verification

`development/scripts/audit-check.py` against this branch:

```
[pass] flake8wrap: flake8wrap.sh has correct SC2086 handling
[pass] ci-review-automation: Automated review and developer automation workflows are compliant
[pass] devpi-fallback: All devpi-backed jobs set a PIP_EXTRA_INDEX_URL fallback
[pass] github-security: Security settings and CodeQL are compliant
```

`pre-commit run --all-files` passes clean on the final tree, including
actionlint over both new workflow files.

Fixes #3057
Fixes #3314
Fixes #3418

https://claude.ai/code/session_01Bsq4TYFMGUFUo6AVABTPAk


mikalstill:
Round 1 addressed in 475840635 -- both `fix` items, both `document` items that can be fixed in this repository, and two of four `consider` items.

- **1 (retest fires the merge lane)** -- verified, and documented rather than narrowed. The four jobs you named are gated `merge_group || workflow_dispatch` and `check_paths` skips its filter on dispatch, so a retest is the merge suite even on a docs-only branch. I took your second option because the failure worth re-trying is normally a merge queue one, so running that lane is usually what is wanted -- the description was what was wrong. `ci.md` now names the four jobs, says the docs-only skip does not apply, and points at the Actions tab for a single job. `pr-retest.yml` itself is untouched: see the note on 3 and 9 below.
- **2 (pr-fix-tests.yml guard)** -- fixed. Confirmed `pr-bot-trigger` does not do this itself, so the exclusion only ever exists in the caller.
- **6 (ci.md "for automation")** -- reworded to say what the JSON is for and that nothing reads it now.
- **5 (vendored docs)** -- cannot be fixed here. Confirmed at the source: `shakenfist/development`'s `docs/automated-pr-review.md` still carries the tool table rows, the `--validate` example and the bot command. A branch is being prepared there.
- **7 (jsonschema)** -- comment extended, and the investigation made the case *weaker* than the finding assumed rather than stronger: `merge-ci-triage.sh` runs the script with the runner's bare `python3` and never installs shakenfist, so CI takes `_validate_basic()` whatever pyproject declares. The comment now says that, and says the pin is for the full check wherever shakenfist *is* installed.
- **4 (seam test)** -- added `shakenfist/tests/test_bot_command_workflow_seams.py`, which also closes the second gap in your test-coverage section. The workflow set is derived from the tree (anything triggered by `issue_comment`) rather than listed, with `KNOWN_BOT_COMMAND_WORKFLOWS` as a floor so an empty derivation fails instead of passing vacuously. Seven properties, each mutation tested: dropping the bot guard, a dispatch target that does not exist, `functional-tests.yml` losing `workflow_dispatch`, the guard phrase drifting from the action phrase, `pr-re-review.yml` hand-rolling the trigger again, firing on plain issue comments, and the derivation missing a known workflow. Every one is caught by the test named for it.

**3 and 9 are deliberately not addressed here.** Both are real, and both are edits to files this PR copies in *verbatim* from `shakenfist/development/templates/ci-review-automation/`. Fixing them on the commit that adopts the template forks the copy immediately; they belong upstream, where every deployment inherits them. They are going into the same `development` branch as 5.

8 and 10 are `none` and taken as recorded. The CodeQL alert on `pr-re-review.yml:130` is the known one -- the template README documents it as triaged and accepted, because the fork guard lives in a composite action the scanner cannot see, and both mitigations it produced (`persist-credentials: false`, the explicit `same_repo` check) are already in this file.

`pre-commit run --all-files` is clean on the final tree.

mikalstill:
Round 2 raised no `fix` items. Three of its seven `consider` items are taken in 65971fa47; the branch is also rebased onto the current develop. Landing after this.

**Taken**

- **3 (`_triggers()` drops list-form `on:`)** -- correct, and the most useful item in either round, because it was a defect in the derivation the whole file rests on. Fixed for all three shapes (`on: push`, `on: [a, b]`, and the block mapping). Demonstrated rather than argued: with a list-form bot command carrying no bot guard dropped into the tree, the old derivation passes 7 of 7 and the new one fails with the guard's own message.
- **4 (`same_repo` on pr-fix-tests.yml)** -- taken, and the reasoning is right that it applies more strongly here than where it was already stated. The previous commit called this the lane worth guarding hardest and then guarded it least.
- **6 (ci.md / workflow phrase sync)** -- taken, both directions. The stale entry this branch removed is exactly what it would have caught.

**Declined**

- **2 (merge ref)** -- already fixed, in `shakenfist/development`'s `templates/ci-review-automation/`, which is where this file is copied from verbatim. Worth recording what measuring it found, because the premise here is mostly wrong and what is underneath is worse. `refs/pull/N/merge` does **not** reliably disappear on a conflict: every open pull request in this repository has one, including #4088, which GitHub reports as `CONFLICTING`. What it is not is reliably *current*. GitHub recomputes the merge commit asynchronously after a push, and this pull request was serving a merge ref one push behind its head -- so a re-review of it would have produced a real, careful review of superseded code. That is the same class of bug as reviewing the default branch, which is what the merge-ref checkout was introduced to fix, arriving again through a different door. The upstream fix confirms the merge commit's second parent is the head, retries briefly for a recomputation, and only then falls back to the head ref, saying on the pull request which one it used.
- **1 (PIP_EXTRA_INDEX_URL)** -- the pinned index is devpi's `root/pypi`, a mirror of the same upstream, so resolution does not currently change; scoping it per job trades a real cold-cache failure for a posture that only starts to matter if a non-mirror index is added. Round 1 recorded that constraint as item 8 and it stands.
- **5 (persist-credentials / hooksPath seam tests)** -- worth having, but this branch is clearing audit findings, and the `same_repo` half of it now has a guard to assert. Reasonable follow-up.
- **7 (link the dispatched run)** -- `gh run list` immediately after `gh workflow run` races the dispatch; the returned run may be someone else's or absent.

**A correction to my previous comment.** I said round 1's item 5 needed a fix in `shakenfist/development` and that a branch was being prepared for it. That was wrong, and I repeated the finding without checking it. `render-review.py` and `review-schema.json` still exist, in `shakenfist/actions/review-pr-with-claude/`; the vendored document describes *those* and says in as many words that "a project carrying its own copy is a leftover of the retired comment addresser, and the consistency audit reports it" -- it is the document that justifies this pull request's deletion, not stale text. The cited line 176 sits inside a section headed "The comment addresser, retired" and instructs nobody to use the phrase. The vendored copy is identical to its source apart from link rewriting. No upstream documentation change is needed.

`pre-commit run --all-files` clean. The rebase resolved one conflict by hand: develop moved `pylogrus` out of the direct-dependency block in 05d3b2aad, so neither side could be taken wholesale.

sfconductor:
Merged at `65971fa4776b`, bypassing the merge queue.

Issue #4115, opened 2026-09-06
Inner veth ends stay at MTU 1500, so the 7950 advertised to guests only holds inside the virtual network


mikalstill:
## Summary

`ip link add <name> mtu <N> type veth peer name <peer>` sets the MTU on the
primary device only; the peer keeps the kernel default of 1500. Every veth pair
Shaken Fist creates is built that way, and only the *outer* end is ever fixed up
afterwards. So both ends of the network namespace's datapath sit at 1500 while
the bridges either side of them, the instance taps, and the MTU advertised to
guests by DHCP are all 7950.

## Evidence

The kernel behaviour, on this tree's own command:

```console
$ unshare -r -n ip link add veth-test-o mtu 7950 type veth peer name veth-test-i
$ ip -o link show
2: veth-test-i@veth-test-o: <BROADCAST,MULTICAST,M-DOWN> mtu 1500 ...
3: veth-test-o@veth-test-i: <BROADCAST,MULTICAST,M-DOWN> mtu 7950 ...
```

The command is built in two places, both with the MTU before `type` and so both
affecting the primary only:

- `shakenfist/util/network.py:265-276` (`_create_interface_inner`)
- `shakenfist/daemons/privexec/util.py:140-145` (`create_interface`)

Both default `mtu` to `config.MAX_HYPERVISOR_MTU - 50`, which is 7950 with the
shipped `MAX_HYPERVISOR_MTU` of 8000 (`shakenfist/config.py:652-655`).

`BridgedVXLanNetwork._apply_create_on_network_node()` then explicitly re-sets the
MTU of each *outer* end as it enslaves it to a bridge, with a comment pointing at
bug 952 (adding an interface to a bridge overwrites the bridge's MTU):

- `veth-<vxid>-o` to the vx bridge's MTU (`bridged_vxlan_network.py:426-430`)
- `egr-<vxid>-o` to the egress bridge's MTU (`bridged_vxlan_network.py:450-455`)

Neither inner end is ever touched. The committed operator guide shows the result
from a real deployment (`docs/operator_guide/networking/overview.md:120-140`):

```
12: veth-e2300f-i@if13: ... mtu 1500 ...
14: egr-e2300f-i@if15: ... mtu 1500 ...
```

against `vnet1` at 8900 in the same document, and dnsmasq advertising
`dhcp-option=26,7950` to every instance
(`shakenfist/managed_executables/dnsmasq.py:197`,
[...]

Issue #4114, opened 2026-09-06
Unrouting an address never releases its IPAM reservation, so the floating pool only recovers on network delete


mikalstill:
## Summary

`DELETE /networks/<ref>/route/<address>` tears down the datapath but never
releases the IPAM reservation. The address stays out of the floating pool until
the owning network is deleted, so a workload which routes and unroutes addresses
against a long lived network consumes the pool permanently.

## Mechanism

`NetworkUnrouteAddressEndpoint.delete()`
(`shakenfist/external_api/network.py:683-700`) looks the reservation up,
validates that this network owns it, emits an audit event, enqueues the
`unroute_address` operation — and stops. There is no `fn.ipam.release(address)`
anywhere on the path.

Compare the defloat path, which calls `api_util.release_floating_ip(ni)`
(`shakenfist/external_api/interface.py:131`). That helper exists precisely to do
the two things the routed path skips:

```python
def release_floating_ip(ni):
    # Inverse of assign_floating_ip. The host side teardown (veth and DNAT) is
    # handled asynchronously by the interface_defloat job; here we release the
    # IPAM reservation and clear the interface record so the address returns to
    # the pool. Mirrors the same two steps in NetworkInterface.delete().
```

`assign_routed_ip()` (`shakenfist/external_api/util.py:33-38`) has no inverse at
all — nothing in the tree releases a `ReservationType.ROUTED` address on unroute.

Neither reclamation path picks it up either:

- The floating IP reaper builds its "legitimately in use" list for routed
  addresses out of the reservations themselves
  (`shakenfist/daemons/network/floating_ip_reaper.py:187-213`): an address counts
  as in use if its `reservation_type` is `ROUTED` and the network named by
  `user_uuid` still exists. Whether it is *actually routed* is never checked, so
  the test is circular and an unrouted reservation vouches for itself.
- `sf-cluster` releases a `ROUTED` reservation only when the owning object is
  missing, or has been in state `deleted` for more than 300 seconds
  (`shakenfist/daemons/cluster/main.py:176-199`). A live network never qualifies.

So the address comes back on network delete, and not before.

## Why it matters

The k3s orchestration allocates routed addresses for metallb service pools,
which is exactly the shape that allocates and releases them repeatedly against a
network that outlives them.

The floating pool is a fixed CIDR block, and both allocation paths in
`reserve_random_free_address()` raise `CongestedNetwork` when it is full
[...]

PR #4118, opened 2026-09-07
Make routed and floating addresses answer from inside their own network

### LINKED ISSUE #3662: Instances cannot reach their own network's routed addresses (no u-turn routing)
## Summary

Instances on a virtual network cannot reach that network's own routed addresses: traffic from an in-network client to a routed address is refused rather than being u-turned back into the network by the network node. Floating addresses have the equivalent problem at the NAT layer (no hairpin). The result is that no Shaken Fist address type is currently both stable across instance replacement and reachable from inside the network.

## Reproduction

Measured today on a live cluster (namespace `k3s-test-006`, node network `ef6ed529-de87-4e56-8b8c-c58cf9eb59e9`, three `debian:12` instances built by the k3s orchestration tooling):

1. Allocated a routed address with `route_network_address()`: got `192.168.10.37`.
2. Claimed it on instance `k3s-k3s-node-001` with `ip addr add 192.168.10.37/32 dev eth0` (this is exactly the mechanism metallb speakers use for their pool addresses, which Shaken Fist's anti-spoofing evidently permits).
3. Results, all against `https://192.168.10.37:6443/ping` (a k3s supervisor which binds all addresses):

| Client | Result |
|--------|--------|
| The holding instance itself | HTTP 200 |
| External host on the physical network | HTTP 200 in 7ms |
| Another instance on the same virtual network | connection refused after ~3s |

The refusal (rather than a timeout) suggests the SYN never reaches the holder and the network node actively rejects or fails to route it: in-network clients send the off-subnet destination to their default gateway (the network node), which routes external ingress for this address into the network but apparently not traffic arriving from the network side.

For completeness, floating addresses were tested the same way: an in-network client curling another instance's floating address times out entirely (NAT hairpin is not implemented).

## Why this matters

- **MetalLB / LoadBalancer services are invisible inside their own network.** Any non-cluster instance sharing a network with a k3s cluster built by shakenfist/client-python-k3s cannot reach the cluster's LoadBalancer service addresses. (Cluster members can, but only because kube-proxy intercepts the LB address in iptables before routing happens — which is how this stayed hidden.)
- **It blocks a stable API join address for control plane replacement.** The k3s tooling wants replace-instead-of-upgrade node lifecycle for control plane nodes, which needs a join address that survives replacement and is reachable by joining nodes *inside* the network. A routed address claimed by the current control plane node would be exactly that — external kubeconfig access and in-network joins through one stable address — but the in-network leg is refused. The client tooling is working around this by tracking a mutable `join_address` in cluster metadata instead (k3s agents only need the registration address at join time), but that is a workaround, not an equivalent: it leaves kubeconfigs pinned to whichever address they were generated against.

## Suggested fix

Have the network node route (or NAT-reflect, for the floating case) traffic that arrives *from* the virtual network and is destined to that network's own routed addresses back into the network, the same way external ingress for those addresses is handled. The routed-address case seems the more valuable of the two: it makes routed addresses location-transparent, so anything metallb-like (or a hand-claimed service address) behaves identically for all clients.

Happy to help with the fix; the reproduction above is easily scripted.

*(Issue drafted with Claude Code assistance.)*

mikalstill:
Fixes #3662.

Both of Shaken Fist's externally reachable address types answer from
everywhere except the network they belong to. This makes them answer from
inside it too, so a routed or floating address behaves the same whether the
client is an instance on the network, elsewhere in the cluster, or outside
it entirely.

## The routed address

A routed address is exactly one route, in the network node's **root**
namespace, pointing the `/32` at the network's bridge. An instance's packet
never reaches that table. It goes to its default gateway, which is the
network's own namespace, and the namespace decides for itself.

Routed addresses come from the floating pool, and the namespace has that
whole block on link on its egress veth — that is where the floating gateway
address lives. So the namespace treats the routed address as a neighbour on
the egress bridge and ARPs for it there. Nothing answers. Unlike a floating
address, which is anchored as a `/32` on a veth *inside* the namespace and so
has something to reply for it, a routed address is configured on no interface
anywhere; it exists only as a route. The packet dies at the instance's own
gateway while the same address answers happily from outside the cluster.

The fix is the matching `/32` inside the namespace, pointing back out the veth
into the network — the same u-turn the root namespace already performs for
everybody else.

## The floating address

The namespace's DNAT rule carries no incoming interface match, so it fires for
in-network traffic exactly as it would for traffic from outside, and sends the
packet straight back out the veth it arrived on. The holder then replies
directly over the virtual network's layer 2, because the caller is its
neighbour, and that reply never passes through the namespace again. Nothing
un-DNATs it, so the caller receives a packet from an address it never spoke to
and waits until it gives up.

The fix is one POSTROUTING masquerade scoped by `-s <netblock>` and
`-m conntrack --ctstate DNAT`, which puts the namespace back on the return
path for the u-turn and leaves everything else alone. Both matches are load
bearing, and they exclude different things: `-s` is what preserves the
caller's real address for clients outside the cluster, whose traffic is
DNAT'ed here too and would otherwise be masqueraded as well; `--ctstate DNAT`
is what excludes a routed address turning around in the same namespace, and
anything else forwarded back into the network unrewritten, since a routed
address is never DNAT'ed and its whole point is that it arrives unrewritten.

The cost is stated rather than hidden: the holder sees an in-network caller as
the network's gateway. That is inherent to hairpin NAT, and a small extension
of what a floating address already is — an address its holder cannot see.
Something that needs to know who is calling should be reached on a routed
address.

## Three commits

1. **Install a network's NAT rules idempotently.** Prerequisite, and two
   defects in its own right. `_enable_nat`'s "do we have NAT already?" probe
   listed POSTROUTING in the *root* namespace while every rule it guards is in
   the network's namespace, so it could not see the rules it was checking for.
   Usually it matched nothing and a rebuilt network re-appended all three
   rules; and because the comparison was a substring search over the whole
   listing, an unrelated `172.16.0.0/16` rule matches a `172.16.0.0/24`
   network — on which path it returned `RULES_ALREADY_PRESENT`, which
   `util_concurrency.enable_nat` turns into `EnableNATFailed`, so the network
   failed to come up. Each rule now carries its own `-C`. Turning the rules
   into data also showed the return-direction FORWARD rule has been matching
   the literal string `'vx_veth_inner'` rather than the interface; harmless
   only because the namespace's FORWARD policy is ACCEPT. The rules move to
   `shakenfist/util/iptables.py`, which imports nothing and so can be read by
   something other than the root daemon -- which is what lets commit 3 both
   add a rule and audit for it without writing it out twice.
2. **Route a routed address inside its own network.** The `/32` above, plus
   `ip route replace` on both routes (they are re-applied, and `add` errors on
   an existing route — which would also abandon the second route because the
   first raised), plus tolerance of an absent route on both deletes.
3. **Hairpin a floating address turned around inside.** The masquerade above,
   plus the audit which gets it onto a cluster that already exists --
   `Network.is_okay()` now asks the namespace for the hairpin rule, one
   `iptables -C` per NAT-providing network per maintain pass, because nothing
   has ever re-run a network's namespace iptables setup while the network was
   otherwise working.

## Reaching existing clusters

This applies to both halves, and the iptables half was missed in the first
round of this PR -- see the review response below.

Nothing re-applies either route while a network is healthy — the
reconciliation that does fires only for a network which has drifted far enough
to need rebuilding. A cluster upgrading while all its networks were well would
never acquire the namespace route for the addresses it already had. So the
maintain pass now reads the namespace routing table for the networks which
actually have routed addresses (a handful even on a large cluster, one exec
each) and re-routes only the missing ones, without dragging the network
through a rebuild. A namespace which cannot be read at all is left to the
rebuild path, which owns that drift.

The same reasoning is why the unroute deletes tolerate a missing route: an
address routed before this branch has only one of the two, and failing the
unroute against the absent one would leave it reserved forever.

## Verification

- Full unit suite green; all ten pre-commit hooks pass, including mypy.
- New functional test, `cluster_ci_tests/test_in_network_reachability.py`: the
  black box version of the bug report. Two instances on a network, one claims
  a routed address the way metallb would, the other pings it; then the same
  for a floating address. It derives the guest's interface name from the MAC
  the API assigned rather than assuming `eth0`.
- Ten mutations, each failing the assertions which claim to prove it, against
  a green baseline: dropping the namespace route, going back to `ip route
  add`, dropping the delete's tolerance, restoring the `'vx_veth_inner'`
  literal, dropping the `-C` check, dropping the hairpin rule, dropping its
  conntrack match, making the reconciliation never find anything, letting it
  run on a hypervisor, and letting an unreadable namespace report every
  address as missing.
- The `iptables` and `ip link`/`ip route` behaviours the change depends on
  (`-m conntrack --ctstate DNAT`, `-C` idempotency, `ip route replace`, exit
  code 2 for an absent route) were reproduced locally in a user namespace
  rather than assumed.

## Review round 1

All twelve items triaged: 2 `fix`, 2 `document`, 8 `consider`. Eleven
addressed, amended into the three commits above rather than added as a
fourth.

The one that mattered was item 1, and it was right: `_enable_nat` is only
reached from `_apply_create_on_network_node`, and neither `is_created()` nor
the dnsmasq check has ever looked at iptables, so a cluster upgrading while
its networks were healthy would never acquire the hairpin rule. That is the
same gap this PR had already closed for routed addresses, not carried over.
Rather than write the rule out a second time in the audit, the rule table
moved to `shakenfist/util/iptables.py` and both the installer and the audit
derive from it; a unit test asserts the audited rule is one the install
actually writes. The audit uses `iptables -C` rather than parsing `-S`,
because iptables normalises what it prints and the dotted-quad netmask the
network object holds lists back as a prefix length -- verified in a user
namespace, not assumed.

Item 5 was also right and is corrected here, in the rule's comment, the test
docstring and the docs: `-s` excludes the external caller, `--ctstate DNAT`
excludes the routed-address u-turn. The rules themselves were always correct.

Also addressed: the route-repair guards now read `net_ip_op` history rather
than `net_op`, so a repair which can never succeed is no longer re-enqueued
every pass forever (2); the namespace probe moved below the pending-operation
gate (10); `get_host_routes` validates its tokens (4) and gained the parsing
test its sibling already had (3); `_apply_unroute_address` declines a deleted
network the way its sibling does (6); the cross-network routed-address case
is documented as unsupported, with the missing blank line fixed (7);
`_add_floating_ip` now uses the shared idempotency helper, which grew an
error-text return so the distinct message was not lost (9); copyright header
(11); the redundant idempotency test dropped with its rationale folded into
the stronger test above it (12).

Not addressed: item 8, an end-to-end assertion that an externally originated
connection still reaches the holder with the caller's real address. It needs
a listener in the guest or conntrack inspection on the network node, and it
guards against a future edit to the rule rather than against a code change;
the `-s` match is pinned by unit test instead. Happy to add it if you disagree.

Nineteen mutations now, each failing the assertions which claim to prove it
against a green baseline -- the original ten plus nine for this round:
dropping the NAT audit from `is_okay`, making the rule check always answer
present, letting the install's rule drift from the audited one, dropping the
address validation, pointing the repair guards back at `net_op`, moving the
probe back above the pending gate, removing the unroute deleted-network
guard, dropping the append failure text, and dropping the conntrack match.

## Review round 2

Eleven items: 2 `fix`, 1 `document`, 7 `consider`, 1 `none`. Ten addressed,
amended into the same three commits.

Both `fix` items were right and are small. `enable_nat`'s client raised a bare
`EnableNATFailed()`, so the `error_text` the previous round added to the reply
never reached the `ErrorReport` -- an operator saw `network.nat.enable_failed`
and nothing else. The exception now carries the reply's error name, text and
network uuid, the way `HashFailed` does, and a test asserts all three reach
`str(exc)`. `shakenfist/util/iptables.py` is now in the tox mypy rollout at
`--strict`, which it passes.

The `consider` item worth the most was 3, and it is the mirror of the bug this
PR is about: the maintain pass audited the in-namespace route but not the
root-namespace one, which can be lost while the bridge survives -- leaving an
address unreachable from *outside* the cluster with nothing looking for it, in
a network `is_created()` still calls healthy. Both tables are read now, and an
address missing from either is re-routed; re-routing installs both, so the
repair path is unchanged.

Also taken: the stale `-o vx_veth_inner` FORWARD rule an older Shaken Fist
wrote is now deleted before the install, tolerating absence, so an upgraded
namespace converges on exactly what `util_iptables` describes rather than
carrying the broken rule beside the fixed one forever (4); `mesh_okay` folds
into `routed_here` so a network heading for mesh repair does not pay for a
probe whose answer that branch discards (6); `is_nat_okay()` gained its own
node-role guard, so calling it from a hypervisor reports "nothing here to
audit" rather than emitting a status event saying the rules have drifted (8);
the `net_ip_op` history's two properties -- it counts a user's errored unroute
against the automatic route repair, and only a manual route or unroute clears
a quiesced one -- are documented on the guard (9); the docs name the subject
of "to make that work" now that four paragraphs separate it from the DNAT
explanation (10); and the new `xt_conntrack` requirement is called out as a
hard one, since a node without it fails network creation rather than degrading
(7).

Not addressed: item 2, the succeed-but-not-converge repair loop. The reviewer
is right that the shape exists, but it is not new here and not specific to
these probes -- a dnsmasq which starts cleanly and still fails
`is_dnsmasq_running()` drives the identical full-recreate-every-pass loop
today. Both new probes are pinned to their installers rather than
independently reimplemented: `-C` takes the same argument list `-A` wrote,
from one module, with a test asserting the audited rule is one the install
writes; and a `/32` lists without its prefix, which is what `get_host_routes`
matches on. Both were checked against real `iptables` and `ip` in a user
namespace rather than assumed. A second circuit breaker keyed on a different
signal is more machinery to grow an untested edge, so the existing ERROR
breaker stays the guard. Happy to add it if you disagree.

Seven more mutations, each failing the assertions which claim to prove it
against a green baseline: raising `EnableNATFailed` without the reply detail,
dropping the node-role guard, dropping the floating-network guard, reading
only the namespace routing table, probing routes for a drifted mesh, leaving
the stale FORWARD rule in place, and treating an absent stale rule as a
failure. Twenty-six across the three rounds.

## Review round 3

Twelve items: 2 `fix`, 1 `document`, 7 `consider`, 2 `none`. Eight
addressed, amended into the same three commits.

Both `fix` items were right. Item 1 is the one that mattered, and it is a
defect in the cleanup the round before added: `remove_netns_iptables_rule`
issued a single `-D`, and `-D` removes one matching rule. The `_enable_nat`
this PR replaced appended unconditionally on every create and every
maintain-driven recreate, so the namespaces the cleanup exists for can hold
several copies of the broken `-o vx_veth_inner` rule -- and removing one of
them left the rest, with nothing ever looking again, because the `-C` for the
corrected rule cannot match a broken one. The delete now loops until iptables
says there is nothing left, bounded at 32 with a warning if it ever reaches
the bound.

Duplicates of the *current* rules are deliberately left where they are, which
is now stated at the helper rather than implied. Extra copies of a rule which
is byte for byte the one we want are inert -- the first match wins and they
all say the same thing -- while converging on one copy means deleting the
rule and appending it again, which opens a window with no rule at all. That
window is not acceptable for `_add_floating_ip`, which installs a live
floating address's DNAT through the same helper. The comment on the install
no longer claims more than that: the namespace ends up holding no rule which
is not in `util_iptables`, and may hold more than one copy of a rule which
is.

Item 2 was right on the substance and wrong on the exit code, which is worth
recording because the docstring said so: `ip netns exec` against a missing
namespace exits **255**, reproduced against iproute2 rather than assumed --
not the 1 the review claims. The gap it names is real, though. The
deleted-network guard does not cover every namespace which is not there: a
network in `delete_wait`, or one on a network node which rebooted before the
namespace was rebuilt, is not `deleted` and still has nothing to exec into,
so the unroute errored and the address stayed reserved forever -- exactly the
outcome the exit-code-2 tolerance beside it was added to prevent. 255 is now
tolerated separately with an audit event, and every other exit code still
raises. The code itself is named once, in `util_network`, with the
distinction from `ip`'s own exit codes written down.

Also taken: the `xt_conntrack` requirement is now in the installation
prerequisites and the upgrade notes rather than only in a subsection of the
networking overview, with the one-line check an operator can run before
rolling the daemons (7); `check_for_iptables_rule` says in its docstring that
its argument is interpolated into a command sf-privexec runs as root through
a shell and must never carry user-supplied data (6); the retired
`RULES_ALREADY_PRESENT` enum member carries a comment recording that
returning it is the bug this PR fixed, and that its number stays claimed (8);
the route audit's `except ProcessExecutionError: return []` logs the
exit code and stderr before returning, so a persistent exec-layer fault no
longer looks exactly like a healthy network (9); and the two rule helpers
have direct unit tests for their own contracts rather than only being walked
over the rule table by `_enable_nat`, including the append-failure text, the
tolerance of an absent rule, the drain, and the bound (10).

Not addressed, all `consider`:

- **3**, a targeted NAT repair instead of a full recreate. The routed-address
  branch exists because a NetIPOp task for it already did; a NAT-only repair
  needs a new NetOp task type, a dispatcher entry and a coalescing decision.
  That is more surface than the one-off upgrade wave saves: the recreate is
  idempotent, runs at background priority, and is gated by both the
  pending-operation check and the circuit breaker, so it is one recreate per
  NAT network, once.
- **4**, the audit's cadence. One `iptables -C` per NAT-providing network per
  pass, which is the same order as the `check_for_interface` probe
  `is_created()` already runs for every network on every pass. Worth
  measuring if the audit ever grows past one rule; not worth a skip-N-passes
  mechanism now.
- **5**, narrowing the repair's guard history to `route_address`. The
  property is real and was documented on the guard last round. Item 2's fix
  removes the most likely source of the errored unroutes that would trip it,
  and a per-task-type history is new query surface for a case which already
  wants operator attention.

Five more mutations, each failing the assertions which claim to prove it
against a green baseline: a single `-D` rather than draining every copy, a
changed duplicate bound, dropping the 255 tolerance on unroute, swallowing
every unroute failure rather than only the missing namespace, and never
removing the stale rule. The first attempt at the bound assertion read the
constant, so it moved with the mutation and could not fail; it asserts the
literal now. Thirty-one across the four rounds.

Each of the three commits passes the unit suite on its own, all ten
pre-commit hooks pass on the final tree, and the two documentation links
resolve.

## Merge CI round 1

The first merge group ejected the PR, and the cause was this PR's own new
test rather than anything it changes. 110 of the 111 cluster tests passed on
every topology and the one failure was
`test_routed_and_floating_addresses_answer_from_inside` -- in cleanup, after
both of its assertions had already passed. Both pings succeeded with 0%
packet loss on Debian 12 tier, Debian 12 cluster and Ubuntu 24.04 cluster,
and the routed one even shows the ICMP redirect the new namespace route
produces, so the run is evidence for the change:

```
routed:   3 packets transmitted, 3 received, 0% packet loss   (ttl=64)
          From 192.168.242.1: icmp_seq=2 Redirect Host (New nexthop: ...)
floating: 3 packets transmitted, 3 received, 0% packet loss   (ttl=63)
```

What failed was the `addCleanup(unroute_network_address, ...)` the routed
commit registered. testtools runs cleanups *after* `tearDown()`, and
`BaseNamespacedTestCase.tearDown()` ends by deleting the namespace, so by the
time the cleanup ran the namespaced client could not authenticate: 401 on the
route DELETE, then 404 `namespace not found` from `/auth`. Deterministic, and
identical on all three topologies. The `pull_request` gate could not have
caught it -- `cluster_ci_tests` do not run there, so the merge group was this
test's first execution.

The unroute now happens at the end of the test body instead, with a comment
recording why, the way `smoke_ci_tests/test_loki.py` and
`cluster_ci_tests/test_namespace_claims.py` already do for the same trap.
Nothing leaks by moving it: `tearDown()` deletes the network, and that is
what releases the address. Moving it also turns it into coverage rather than
cleanup, which is the better reason it belongs in the body -- the endpoint
answers 403 or 404 unless the address really is routed by this network, so
the call returning is an assertion about the half of the routed-address
lifecycle the ping cannot see, and the unroute path is one this PR changes.

Amended into commit 2 rather than added as a fourth commit, and the whole
tree passes all ten pre-commit hooks.

## Not in this PR

Two adjacent defects found while tracing, filed rather than fixed here:

- #4114 — unroute never releases the IPAM reservation, so the floating pool
  only recovers on network delete.
- #4115 — inner veth ends stay at MTU 1500, so the 7950 advertised to guests
  only holds inside the virtual network.

https://claude.ai/code/session_01UrU2FCqx1mUnQCiACcSqcy

### DIFF
--- a/docs/operator_guide/installation.md
+++ b/docs/operator_guide/installation.md
@@ -52,6 +52,13 @@ Each machine in the cluster should match this description:
   standard MTU of 1,500 bytes would result in fragmentation. I generally
   select 9,000 bytes.
 * Has at least 1 gigabit connectivity on the "mesh interface".
+* Has the iptables connection tracking match (`xt_conntrack`, what
+  `iptables -m conntrack` loads) available if it will be a network node. It
+  is present in any stock Ubuntu or Debian kernel, so this only bites a
+  custom or heavily stripped kernel. A network node without it cannot bring
+  up a NAT providing virtual network at all, rather than degrading -- see
+  the floating IP discussion in [the networking overview](networking/overview.md)
+  for why the rule needs it.
 
 Your ansible control node needs `ansible-core >= 2.15`.
 
--- a/docs/operator_guide/networking/overview.md
+++ b/docs/operator_guide/networking/overview.md
@@ -372,13 +372,56 @@ work equally well for other traffic.
     intended for a specific floating IP, but in return must have been configured
     to use that floating IP.
 
-The implementation of routed IPs is relatively trivial. For each routed IP, a
-route on the network node into the relevant virtual network bridge is created.
-Such a route might look like this:
+The implementation of routed IPs is relatively trivial. For each routed IP,
+two routes are created on the network node, one in each of the two routing
+tables which decide where the address's traffic goes.
+
+The first is in the network node's own routing table, and is how traffic
+from outside the cluster arrives. It puts the address onto the virtual
+network's bridge, where whichever interface was configured to answer ARP for
+it picks it up:
 
 ```
-ip route add 192.168.15.29/32 dev br-vxlan-e2300f
+ip route replace 192.168.15.29/32 dev br-vxlan-e2300f
 ```
+
+The second is inside the virtual network's own network namespace, and is how
+an instance *on* that network reaches the address:
+
+```
+ip netns exec 17be6538-8f96-4ccb-b71e-a7e3022fead3 \
+    ip route replace 192.168.15.29/32 dev veth-e2300f-i
+```
+
+That second route is needed because a routed IP comes from the floating pool,
+and as the routing table above shows, the namespace has the whole floating
+block on link on its egress veth. Without a more specific route an instance's
+packet reaches its default gateway and the namespace then ARPs for the address
+out on the egress bridge, where nothing answers -- unlike a floating IP, a
+routed IP is not configured on an interface anywhere on the network node, it
+exists only as a route. The `/32` turns that into a u-turn back into the
+network, which is the same delivery the network node performs for everybody
+else, and it means a routed address behaves identically whether the client is
+on the virtual network the address is routed to, or outside the cluster
+entirely (github issue #3662).
+
+A client on a *different* Shaken Fist virtual network is not covered by
+either route, and remains unsupported. That namespace also has the whole
+floating block on link on its egress veth, so it ARPs for the routed address
+on the egress bridge -- and unlike a floating IP, which the namespace owning
+it answers for from its egress veth, a routed IP is configured on no
+interface there and nothing replies. Reach a routed address from another
+virtual network the same way anything outside the cluster does, by way of the
+network node.
+
+Both routes use `ip route replace` rather than `ip route add` because they are
+re-applied: the network daemon's maintenance pass restores a network's routed
+addresses whenever it rebuilds the network on the network node, and also
+checks each maintenance pass that both routes are still there -- the one in
+the root namespace and the one inside the network's namespace. An address
+missing either is re-routed, which installs both, so a network which is
+otherwise healthy is not dragged through a rebuild to recover one route.
+Removing a routed IP deletes both routes, and tolerates either being absent.
 
 ## Interface naming conventions
 
--- a/docs/operator_guide/networking/overview.md
+++ b/docs/operator_guide/networking/overview.md
@@ -166,6 +166,7 @@ target     prot opt source               destination
 Chain POSTROUTING (policy ACCEPT)
 target     prot opt source               destination
 MASQUERADE  all  --  172.16.0.0/24        anywhere
+MASQUERADE  all  --  172.16.0.0/24        anywhere             ctstate DNAT
 ```
 
 So, the default route is via the egress veth at `egr-e2300f-i` (inside the
@@ -177,9 +178,11 @@ which is wired to the linux kernel VXLAN interface at `vxlan-e2300f`. Finally,
 traffic to other floating IPs on `192.168.15.0/24` is routed to the egress
 bridge as well.
 
-The iptables `MASQUERADE` entry is there to convert internal addresses to the
-external `192.168.15.194` "floating gateway" address so instances can talk outside
-their virtual network.
+The first iptables `MASQUERADE` entry is there to convert internal addresses to
+the external `192.168.15.194` "floating gateway" address so instances can talk
+outside their virtual network. The second one, which only matches connections
+whose destination has been rewritten, is for floating IP traffic which turns
+around here instead of leaving -- see the floating IP section below.
 
 Perhaps a diagram would help!
 
@@ -311,11 +314,53 @@ target     prot opt source               destination
 Chain POSTROUTING (policy ACCEPT)
 target     prot opt source               destination
 MASQUERADE  all  --  172.16.0.0/24        anywhere
+MASQUERADE  all  --  172.16.0.0/24        anywhere             ctstate DNAT
 ```
 
 So our floating IP of `192.168.15.29` is DNAT'ed to the instance's IP of
 `172.16.0.37`. This means floating IP traffic "bounces" off the network namespace.
-To make that work, the inside of the veth is configured with the floating IP:
+
+The second `MASQUERADE` rule above handles the case where the client is
+itself an instance on the same virtual network. That traffic is DNAT'ed by
+the same rule and then sent straight back out the veth it arrived on, so the
+reply would travel directly over the virtual network's own layer 2 and never
+return through the namespace -- which means it would never have its source
+rewritten back to the floating address, and the client would discard it.
+Masquerading those u-turned connections to the network's gateway address puts
+the namespace back on the return path, where conntrack can undo the
+destination rewrite.
+
+The two matches on that rule each exclude something different, and both are
+needed. The `-s` match confines the rule to connections which started inside
+the virtual network, which is what preserves the real source address for
+clients outside the cluster -- their traffic is DNAT'ed here too, so without
+`-s` it would be masqueraded as well. The `--ctstate DNAT` match confines the
+rule to connections whose destination was rewritten, which is what excludes a
+routed IP turning around in the same namespace, since a routed IP is never
+DNAT'ed and its whole point is that it arrives unrewritten.
+
+The cost is that the instance holding the floating IP sees such a connection
+as coming from the network's gateway rather than from the calling instance;
+if the destination needs to know who is calling, use a routed IP, which is
+not rewritten at all.
+
+The network daemon's maintenance pass checks each cycle that this rule is
+present in the namespace, and rebuilds the network on the network node if it
+is not. That check is how a cluster which was upgraded while all of its
+networks were healthy acquires the rule at all -- nothing else re-runs the
+namespace's iptables setup for a network which is otherwise working.
+
+`--ctstate DNAT` means a network node now needs the iptables conntrack match
+(`xt_conntrack`, which is present in any stock Ubuntu or Debian kernel and is
+what `iptables -m conntrack` loads). This is a hard requirement rather than a
+degradation: a node which cannot install the rule fails the whole
+`enable_nat` step, so a NAT-providing network will not come up there at all
+rather than coming up without in-network floating IPs. The error reported
+against the failed operation carries iptables' own complaint about the rule
+it refused.
+
+To make the DNAT bounce work, the inside of the veth is configured with
+the floating IP:
 
 ```bash
 debian@test:~$ sudo ip netns exec 17be6538-8f96-4ccb-b71e-a7e3022fead3 ip a
--- a/docs/operator_guide/upgrades.md
+++ b/docs/operator_guide/upgrades.md
@@ -86,6 +86,24 @@ the existing definition. Recreating the instance is the supported
 route; this is a workaround, and it is on you to confirm the instance
 comes back.
 
+### Network nodes need the iptables conntrack match
+
+Since the release which made floating addresses answer from inside their own
+network, a NAT providing network's namespace carries a hairpin masquerade
+rule matched with `-m conntrack --ctstate DNAT`. A network node whose kernel
+lacks `xt_conntrack` will fail to install that rule, and the network will not
+come up there at all.
+
+Every stock Ubuntu and Debian kernel has it, so this is only a concern if you
+run a custom kernel. Check before you roll the daemons:
+
+```bash
+iptables -m conntrack --help > /dev/null && echo ok
+```
+
+See the floating IP discussion in [the networking overview](networking/overview.md)
+for what the rule does and why both of its matches are needed.
+
 ## MariaDB schema migrations
 
 Starting with v0.8, Shaken Fist uses MariaDB to store object state data. The
--- a/protos/privexec.proto
+++ b/protos/privexec.proto
@@ -44,6 +44,15 @@ message EnableNATReply {
     enum Errors {
         OK = 0;
         IPTABLES_FAILED = 1;
+
+        // Retired, and a trap: do not return this. The client
+        // (util.concurrency.enable_nat) raises EnableNATFailed for any
+        // reply which is not OK, so "the rules were already there" --
+        // the normal outcome every time the maintain loop recreates a
+        // healthy network -- became a hard failure and the network
+        // never came up (issue 3662). Each rule now carries its own
+        // iptables -C instead, and success is success either way. The
+        // number stays claimed so no later member reuses it.
         RULES_ALREADY_PRESENT = 2;
     }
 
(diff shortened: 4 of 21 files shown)


mikalstill:

Eleven of the twelve items are addressed, amended into the three existing commits. The upgrade-path gap (item 1) is closed by moving the rule table into a shared `shakenfist/util/iptables.py` that both sf-privexec and the new `Network.is_nat_okay()` audit derive from, so the install and the audit cannot drift. Item 5 was correct: the `-s` and `--ctstate` justifications were attached to the wrong matches, and are corrected in the comment, the test docstring, the docs and the PR body.

Item 8 is declined with reasons in the PR body. Full details of the round are in the "Review round 1" section there.

mikalstill:
Round two addressed: both `fix` items, the `document` item, and six of the seven `consider` items, amended into the same three commits (see the "Review round 2" section of the description).

The one declined is item 2, the succeed-but-not-converge repair loop -- the shape is real but not new here or specific to these probes, and both probes are pinned to their installers rather than independently reimplemented. Reasoning is in the description; happy to add the counter if you disagree.

Item 3 turned out to be the most valuable of the round and is fixed: the root-namespace route is now audited alongside the in-namespace one.

Full unit suite green, all ten pre-commit hooks pass including mypy (which now covers `shakenfist/util/iptables.py`), each of the three commits is green on its own, and seven new mutations each fail the assertions claiming to prove them.

sfconductor:

The conductor will merge this at `3bf40522216d` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `3bf40522216d`, bypassing the merge queue.

PR #4123, opened 2026-09-07
Gate enqueue on the issue link check

mikalstill:
A pull request here can merge with the issue it repairs left open, and
nothing says so.

## Why this repository in particular

GitHub acts on an issue-closing stanza only when it appears in a pull
request's **description**. A stanza wrapped in backticks is a code span and
is never parsed. A stanza in a **commit message** is parsed only for a commit
pushed to the default branch in the ordinary way — and the merge queue
advances develop from its own staging ref, so a commit stanza on this
repository does nothing whatsoever. The convention of putting the stanza in
the commit message is buying traceability here, not closure.

An audit on 2026-09-07 found five issues sitting open behind merged work:
4087 behind 4106, 3999 behind 4108, and 3370, 3371 and 3373 behind 3374,
whose description was empty. 4087 needed both failures at once, which is why
this has looked rare while being one mistake away from happening every time.
The first four have since been dealt with by hand; 3373 turned out not to be
resolved at all, and stays open.

## What this adds

An `issue_link_check` job which runs on pull requests and joins the enqueue
gate's `needs`, so a description which will not do what it says fails while
it can still be edited. The implementation is the reusable workflow in
shakenfist/actions.

It also prints what the pull request will shut, which is the other direction
of the same defect — three issues were shut by accident in the same hundred
days by a keyword sitting beside a reference in ordinary prose. That
direction is reported and never enforced, because descriptions here
legitimately carry headings and table cells of the same shape and a check
which cried wolf on those would be ignored inside a week.

`X-No-Autoclose: #NNNN` in a description opts out, for a pull request which
is deliberately only part of the work.

## Verification

Ten pre-commit hooks pass, including actionlint. The checker itself carries
21 unit tests in shakenfist/actions, and was run against this repository's
real pull requests: 4106 fails naming 4087, 3374 fails naming 3373, 4118 and
4108 pass, and 3816 passes while reporting the issue it shut by accident.

It was also run against the four most recent bot-authored fix pull requests
— 4121, 4122, 4124 and 4130 — which all pass. `issue-fix.yml` appends its
own `Fixes #NNNN` as the first line of the description after
`neutralise-pr-body.sh` runs, so the automated fixer and this gate agree
rather than fighting.

The gate is documented in `docs/developer_guide/ci.md`, under the merge
queue pattern.

https://claude.ai/code/session_01UrU2FCqx1mUnQCiACcSqcy

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -762,6 +762,27 @@ jobs:
       - name: "Immediate success"
         run: true
 
+  # GitHub only acts on an issue-closing stanza in a pull request's
+  # description. A stanza wrapped in backticks is a code span and is
+  # ignored, and a commit message stanza is not parsed at all on the merge
+  # queue path -- so a fix can merge with the issue left open and nobody
+  # any the wiser. This fails while the description can still be edited.
+  # It also prints what the pull request will close, which is how a closing
+  # keyword in ordinary prose shows itself before it shuts a live bug.
+  issue_link_check:
+    name: "Issue links"
+    if: github.event_name == 'pull_request'
+    uses: shakenfist/actions/.github/workflows/issue-link-check.yml@main
+    # This file grants only "contents: read" at the top, and a called
+    # workflow cannot hold more than its caller was given, so the reads it
+    # needs are granted here rather than widening the whole file.
+    permissions:
+      contents: read
+      pull-requests: read
+      issues: read
+    with:
+      runs_on: '["self-hosted","static"]'
+
   can_enqueue:
     name: "Can enqueue"
     needs:
@@ -769,6 +790,7 @@ jobs:
       - sanity_checks
       - check_paths
       - credential_scan
+      - issue_link_check
     if: always() && github.event_name != 'merge_group'
     permissions:
       actions: read
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -769,6 +769,15 @@ jobs:
   # any the wiser. This fails while the description can still be edited.
   # It also prints what the pull request will close, which is how a closing
   # keyword in ordinary prose shows itself before it shuts a live bug.
+  #
+  # This job deliberately carries no same-repo guard, unlike
+  # smoke_collection above. The called workflow checks out only
+  # shakenfist/actions and reads the pull request entirely through the
+  # API, so no fork-controlled code executes on the runner -- and
+  # skipping fork pull requests would silently exempt them from the
+  # gate, since can_enqueue treats a skipped dependency as success.
+  # Those are the terms on which the exposure is accepted; it is not an
+  # omission.
   issue_link_check:
     name: "Issue links"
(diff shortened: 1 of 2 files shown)


mikalstill:
Addressed in 3d5b432. Notes on each item:

**1 (bot issue-fix PRs fail by construction) — respectfully, this does not
happen, and no code changed.** The reading stops at
`neutralise-pr-body.sh` and misses what runs after it.
`issue-fix.yml:880` appends its own `Fixes #${ISSUE_NUMBER}` as the
*first* line of the body, after the neutralisation pass and above the
model prose — with a comment there explaining that the ordering is
deliberate, so an unbalanced fence in model output cannot swallow the
reference. `neutralise-pr-body.sh`'s own header says the same thing:
"The workflow appends its own `Fixes #NNNN` for the issue actually
being fixed after this pass, so that one is unaffected."

The suggested cheap check was run. `check-issue-links.py` against the
four most recent bot pull requests — #4121, #4122, #4124
(`issue-fix-3497`, the one named in the review) and #4130 — exits 0 on
all four, each reporting `Declared: #N` / `GitHub will close: #N`.
`closingIssuesReferences` on #4124 is non-empty.

Option (c) was taken as documentation anyway, because the premise is
worth refuting where the next reader will be standing: the
`neutralise-pr-body.sh` paragraph in `docs/developer_guide/ci.md` now
records that the workflow's own stanza is what satisfies this gate, so
nobody reaches for `BRANCH_RE` later.

**2 and 4 (documentation) — taken.** New `### The issue link check`
subsection in `docs/developer_guide/ci.md` under the merge queue
pattern: why the stanza must be in the description and nowhere else
(backticks make it a code span; a commit-message stanza closes nothing
here because the queue advances `develop` from its own staging ref),
the `X-No-Autoclose: #NNNN` escape hatch, that the reverse direction is
reported and never enforced, and the re-run step — the `pull_request`
trigger declares no `types:`, so correcting a description needs a job
re-run or a push behind it. That adding `edited` is *not* the fix is
recorded too, for the reason given in the review.

The `.github/pull_request_template.md` aside was not taken: it would
prompt every human pull request, and `gh pr create --body-file`
bypasses a template entirely, so it would not have helped the automated
path at all.

**3 (no same-repo guard) — taken as the recorded-exposure comment, not
the guard.** Adding the condition would have been actively wrong here.
`can_enqueue` treats a skipped dependency as success, so guarding the
job would silently exempt every fork pull request from the gate — the
one class of contribution nobody here can see the description of in
advance. The exposure it would buy back is nil, since the called
workflow checks out only `shakenfist/actions` and reads the pull
request through the API. The job now says so.

Items 5, 6 and 7 are informational and need nothing.

sfconductor:

The conductor will merge this at `3d5b4321dc92` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `3d5b4321dc92`, bypassing the merge queue.

PR #4125, opened 2026-09-07
Log an unresolvable node UUID once per API worker

mikalstill:
`resolve_node_uuid()` is an `sf-api` `before_request` hook. On a worker
which cannot resolve its own node UUID it warned **once per request**,
forever — the wrong bound for a fact about the process rather than
about a request. The second copy says nothing the first did not, and at
request volume it buries every other line the worker emits.

It buried the unit tests too, which is how this was noticed: no test
resolves a node UUID, and every API test drives requests through the
hook.

| test module | warnings before | after |
|---|---|---|
| `external_api.test_auth` | 6022 | 1 |
| `external_api.test_claims` | 5566 | 1 |
| `external_api.test_scopes` | 284 | 1 |
| `external_api.test_network` | 180 | 1 |

## What changed

The log line is bounded, not the work. Resolution is still attempted on
every request — a node missing from the database now may be present
later — but the failure is logged at most once per worker process.

The alternative, pinning `config.NODE_UUID` in `ShakenFistTestCase` the
way `test_health_endpoints.py` already does locally, would have quieted
the tests while leaving a misconfigured production worker emitting one
warning per request. That is why the fix is in the application rather
than in the test base.

`shakenfist/tests/external_api/test_node_uuid_resolution_logging.py`
asserts the bound the right way round: 20 requests produce exactly one
warning but 20 `Node.from_db` calls, so a future change which "fixes"
the noise by giving up on resolution fails the test. Reverting
`app.py` to the old `else:` makes the first assertion fail, which was
checked rather than assumed.

`docs/operator_guide/logging.md` gains a short section, because the
absence of repeats no longer means the condition has cleared — the
matching `Resolved node UUID in API worker` line is now what confirms
it did.

## Testing

`pre-commit run --all-files` passes in the branch worktree: all ten
hooks, including the full 4498-test `py3` suite and mypy.

https://claude.ai/code/session_01S2FS2dRJgimXTSGXRqcTZL

### DIFF
--- a/docs/operator_guide/logging.md
+++ b/docs/operator_guide/logging.md
@@ -276,6 +276,25 @@ requires the query to find *that* before it will assert that
 nothing else matched. A detector which has never been observed
 firing is not evidence of anything.
 
+## Conditions logged once per worker
+
+A few warnings describe a condition of the process rather than of
+a request, and those are logged at most once for the lifetime of
+that process. `Failed to resolve node UUID in API worker` is the
+one you are most likely to meet: an `sf-api` gunicorn worker
+resolves this node's UUID lazily on its first non-probe request,
+from the persisted UUID file or from the node's FQDN in the
+database, and warns if neither answers.
+
+Seeing that line once does **not** mean the condition has since
+cleared. The worker keeps retrying on every subsequent request --
+a node absent from the database now may be present later -- but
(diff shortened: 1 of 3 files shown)


sfconductor:
Merged at `416764926236`, bypassing the merge queue.

PR #4138, opened 2026-09-08
Close phase 2 of the CI cloud sizing plan

mikalstill:
Step 2g of [PLAN-ci-cloud-sizing](docs/plans/PLAN-ci-cloud-sizing.md), which closes the two things the phase 2 baseline could not see. Both instrument fixes this phase made have been live since 2026-09-06 (2a's `capacity_degraded` on `develop`, 2e's widened census filter in `shakenfist/actions`), so this harvests eight merge runs that used them.

It is a **classification, not a second baseline** — 32 records against the baseline's 217 — and nothing in step 2d's conclusions is recomputed from it. Every one of them is confirmed or extended.

## Blind spot 1: the guard census now holds a count

`guard.state` is `collected` on all 32 records, against `not_collected` on all 204 baseline ones.

* **3,480 denials** over 32 job-runs — median 121 per `slim-primary` job-run, 134.5 per `slim-tier`. In the same runs the CPU pre-filter dropped 283 candidates in 3,862 evaluations.
* **3,477 of the 3,480 exceed `demand` and nothing else**, split 1,789 measured-alone to 1,688 estimate-tipped. The other three exceed `cpus`, two at the `cluster` stage.
* Nothing malformed, unenforced or dimensionless, and no unrecognised stage or dimension name — the census is fully classified, not merely present.
* `census.truncated` false on all 32, so widening the filter did not push the query into its entry limit.

A denial is **not** a failed create: both placement walks catch it, try the next candidate, and re-walk with the demand clause waived. The two events that would say how often that failed are still outside the filter, so the count of creates the guard actually refused stays unmeasured — recorded as *Future work* for phase 3 rather than measured by widening the window again.

## Blind spot 2: the warm-up prefix is an empty table

`total.capacity_degraded` is present on all 3,359 samples and **false in every one**, including all 367 whose capacity rows read as absent. The prefix is contiguous from sample zero in all 32 job-runs, 9 to 14 samples, 135 to 210 seconds. Step 2f's inference from the baseline's shape is now direct measurement; no read-reliability issue is filed, and #4087 rests on two independent windows.

## A defect in the previous step's tooling

`list_runs()` stopped at the first run created before the window, documented as safe because the runs listing arrives newest first. The API does not promise an order and served this workflow's `merge_group` runs *oldest* first, so 2g's first harvest enumerated nothing and exited zero — the silently half-completed harvest the tool's docstring says it must never be. The window is now a `created=>=` filter on the API call, nothing infers an order, and `--limit` sorts explicitly. Three tests added; all three were mutation-tested (restoring the early exit, dropping the API filter, and dropping the sort each fail a distinct test).

The committed baseline is unaffected, and that was checked rather than assumed: the API returns 66 completed `merge_group` runs in its window today, exactly the number step 2d enumerated.

## Verification

* `pre-commit run --all-files` passes (all ten hooks).
* `tools/check-plan-status.py` passes.
* Definition of done item 10 is now demonstrable from CI itself — job 101911972919 of run 34178278720 prints `Placements refused by the guard: 142`, and those figures match the harvested record for that same bundle.

## Step 2h: the phase closes here

The second commit is step 2h, folded into this branch rather than a second review round. It sets the phase 2 row to `Complete` in the master plan's Execution table and the index arithmetic to `3 of 7`, and `tools/check-plan-status.py` agrees. Nothing else changes.

Phase 3 -- explicit saturation coverage -- is next, and it gates phase 4. What it inherits is listed at the end of the phase 2 plan, with one addition from this work: it should widen the census filter by the two events that say what became of a denial before writing a test which asserts on guard behaviour.

## Review round

The third commit answers the automated review. Two harvest defects it
found are fixed (a timezone-aware `--since` was mis-rendered into the
server-side filter; an empty enumeration could still truncate a
committed dataset and exit zero), the summary record gains four series
counters at `RECORD_VERSION` 2 so the two claims that closed blind spot
2 are recomputable from the committed file rather than from expiring
bundles, and `records-addendum.jsonl` was re-harvested to carry them.
Verifying that re-harvest showed the window itself was not
reproducible -- `--limit` moves with the day it runs on -- so the tool
gains `--until` and both datasets' commands now name both ends. The
re-harvest returned the same 32 records with no field changed except
the new counters.


mikalstill:
Review addressed in 7b137d874. All four action items are taken, and verifying one of them turned up a fifth defect.

**1 (FIX, timezone) — fixed.** Both boundaries are normalised to UTC before they are formatted. The finding was exactly right about the direction: a positive offset is this project's own timezone, and the server-side filter runs first so the lost hours were unrecoverable. `test_a_since_with_an_offset_is_pushed_down_in_utc` pins the shifted stamp; reverting the `astimezone` fails it.

**2 (CONSIDER, empty harvest exits zero) — taken, in both places.** An enumeration which finds no runs now raises *before* `--output` is opened, so re-running a harvest whose window has gone wrong cannot be the thing that empties the last good dataset; a set of runs which yields no records raises after the write and says the file was truncated. Four tests, including one that writes a dataset, runs an empty harvest against it and asserts the bytes are still there.

**3 (DOC, sort comment) — reworded** to what the code actually guarantees: an undated run is kept by the `--since` filter rather than dropped, sorts oldest, and a small enough `--limit` can still leave it out — which is the right way round.

**4 (DOC, capacity_degraded not traceable) — took the second option.** `summary_record()` gains four series counters at `RECORD_VERSION` 2: `capacity_degraded_samples`, `capacity_degraded_absent_samples`, `ledger_unreadable_prefix_samples` and `ledger_unreadable_prefix_seconds`. The absent counter is separate on purpose — a bundle built before step 2a is a different fact from a healthy read and must not be counted as one — and the printed report now says which of the two readings an unreadable ledger was. `records-addendum.jsonl` was re-harvested to carry them, so both headline claims recompute from the file: 0 degraded and 0 absent over 3,359 samples, and `prefix_samples == ledger_unreadable_samples` in all 32 records. The baseline stays at version 1, and its own prefix classification is now named as a third raw-bundle figure in the dataset README, matching the master plan's precedent.

**5 (INFO) — no change**, agreed on the trade-off.

**The fifth defect, found doing (4).** The window was not reproducible. `--since 2026-09-07 --limit 10` names the newest ten runs *as of the day it runs*, and two more merged within hours of the harvest, so the README's own command already enumerated a different ten. The tool gains `--until`, both boundaries are pushed down as a `created=A..B` range, and both datasets' commands now name both ends. Under the pinned window the re-harvest returned the same 32 records over the same 8 runs with **no field different** except `record_version` and the four new counters, which is also the check that version 2 changed nothing else.

Every assertion added here was mutation tested against the defect it claims to catch. `pre-commit run --all-files` passes.

sfconductor:
Merged at `5a60b2637f0b`, bypassing the merge queue.

Issue #3612, opened 2026-08-03
API body parameters are merged into handler kwargs untyped, so wrong-typed input leaks interpreter messages as 400s


mikalstill:
## Summary

`log_request` merges every key of a request's JSON body into the handler's
kwargs with no type checking, and `handle_authorization_exceptions` turns
*any* `TypeError` raised anywhere below a handler into
`400 <str(exception)>`. Together those two mean a caller can make almost any
endpoint answer with the interpreter's own error text, and in some cases with
a 500 and a recorded server exception, purely by sending a wrong-typed body
value.

Issue #3609 was one instance of this: `{'limit': '5'}` on
`GET /nodes/<node>/events` produced
`400 "'<=' not supported between instances of 'str' and 'int'"`.

## Evidence that this is systemic, not one endpoint

Fixing #3609 (PR #3610) turned up four more instances of the same shape
without looking hard:

* `event_type` on the same five events endpoints — a non-string lands in a
  protobuf string field and leaks
  `5 has type int, but expected one of: bytes, unicode`.
* `limit` beyond int32 on those endpoints — coerces fine, then raises
  `ValueError` during protobuf message construction on the gRPC path, which
  every real deployment uses. That escapes as a 500 with a `ValueError` repr
  in the body and one recorded server exception per call.
* `max_versions` / `version_id` on the artifact endpoints — `int()` was
  guarded, but only against `ValueError`, so a JSON `null` took the leaky
  `TypeError` path.
* `offset` on `POST /uploads/<uuid>/truncate` and `length` on
  `GET /instances/<ref>/consoledata` — unguarded or `ValueError`-only `int()`
  calls with the same two outcomes.

Each was fixed at its own endpoint in PR #3610. That is per-endpoint
whack-a-mole: the next endpoint to grow an integer or string body parameter
inherits the defect by default, because the default is unvalidated.

## Proposed fix

Two steps, in order:

1. Validate body parameters declaratively, using the same webargs
   `use_kwargs` schemas already used for query arguments elsewhere in
   `shakenfist/external_api/`. A handler then receives a typed value or the
   request is rejected by the framework with a consistent, non-leaky 400
   before the handler runs.

2. Once handlers can no longer be reached with a wrong-typed kwarg, narrow
   the `except TypeError` in
   `handle_authorization_exceptions` (`shakenfist/external_api/base.py`) to
   the JWT-related `TypeError`s it was actually written for. As long as it
[...]

PR #4141, opened 2026-09-08
API input validation phase 4: enforce the declarations

mikalstill:
Phase 4 of [PLAN-api-input-validation](docs/plans/PLAN-api-input-validation-phase-04-enforce.md), completing steps 4, 5 and 6. Steps 1-3 landed earlier in #4101.

`API_VALIDATION_MODE` now defaults to `enforce`, so input which does not match an endpoint's published parameter declarations is refused with `{"error": "<parameter>: <reason>", "status": 400}` rather than reaching a handler which raises `TypeError` and hands the caller the interpreter's own words. That leak is #3612 in its purest form.

## The confirmatory reading which unlocked the flip

Decision D22 said the switch could not be thrown until the two declaration bugs the warn window found were shown to be fixed. Both readings are now in the plan's measurement log:

* **Functional CI: nine to zero.** The `merge_group` run of #4101, the first full suite over the merged code, produced **zero** `unknown-parameter namespace` findings against the nine of 2026-08-13. The other three signatures are unchanged at 21, 6 and 6 — all intended rejections. Both of phase 3's guards against a false negative were re-run rather than assumed: the registry compiled line appears in all six bundles at 139 handlers, and the two cluster CI tests which drive a namespace-scoped lookup both ran and passed.
* **Metadata `value`: probed, not waited out.** D22 asked for 72 hours of absence "during which the k3s orchestration traffic has actually run", but that named the wrong thing — the traffic is `client-python-k3s` exercised by hand, not a running cluster, and it last ran 35 hours *before* the fix deployed. Since the declaration is per route, any JSON-valued metadata write exercises the same schema. A dict-valued and a list-valued write against sfcbr produced no finding, alongside a deliberate undeclared body key in the same second which did. That is positive evidence rather than a quiet channel, which is what the wait could not have given us.

Widening the reading to the whole 30-day warn window also bounds the blast radius: **all 294 findings on requests which actually succeeded are the metadata bug, now fixed. The remaining 63 already answer 400 and answer 400 after the flip.** The enforceable surface on sfcbr is empty.

## Caller-visible changes the plan did not anticipate

Both are documented in the release note rather than left to be discovered:

* **Eleven refusals which already answered 400 now carry validation's wording** instead of the handler's, because validation answers first — `key is not a string` becomes `key: Not a valid string.` The handler guards are untouched and still answer whenever the mode is not `enforce`.
* **A body key which repeats a path parameter is now refused** as a body/path collision (D18). The shipped client never sends one; three of our own tests did, which is how it surfaced.

## One commit which is not a plan step

`5a5847259` is not in the step list. Enforcing turned a finding's detail from a log line into the caller's error message, which exposed a rendering bug warn mode had kept private: marshmallow answers with a dict keyed by index when a *container's elements* are wrong, and the `str()` fallback put a Python dict repr in the response — `disk: {0: ['Not a valid mapping type.']}`. It reaches seven declarations. The fix names the element instead: `disk[0]: Not a valid mapping type.` A phase whose purpose is to stop the API answering in the interpreter's words should not ship a new way for it to answer in marshmallow's.

## Verification

* Every assertion in step 4 was mutation tested — ten mutations, each caught, each by the test which names the property. The three added for `5a5847259` likewise.
* `tools/check-api-declaration-guards.sh`: all 34 mutations caught.
* Full unit suite: 4555 tests, 0 failures. `pre-commit run --all-files` clean.
* `UNDECLARED_BY_DESIGN` is now empty, and the five metadata delete handlers no longer accept a `value` kwarg none of them read.
* All 13 Definition of done items met.

The `missing-required` filter is deliberately untouched (D17): several parameters are declared required while omitting them has always worked, so enforcing that is phase 6's decision. A request-level test pins that `enforce` plus an omitted required parameter still reaches its handler.

Operators whose callers break can set `API_VALIDATION_MODE=warn` or `off` as the rollback.

### DIFF
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -305,19 +305,28 @@ enumerate is a shape nothing checks.
 
 The declarations are compiled into marshmallow schemas at startup
 (`shakenfist/external_api/validation.py`) and checked against every
-request. **Nothing is rejected**: `API_VALIDATION_MODE` defaults to
-`warn`, which logs what would have been refused and changes no
-response. Setting it to `enforce` answers `400` in the usual
-`{"error": ..., "status": ...}` shape, and is the switch to throw once
-the warn log is understood.
+request. `API_VALIDATION_MODE` defaults to `enforce`: a finding other
+than `missing-required` answers `400` in the usual
+`{"error": "<parameter>: <reason>", "status": ...}` shape, naming the
+offending parameter, before the handler or any of its per-method
+decorators ever run. `warn` is the operator's rollback for a caller
+that breaks — it logs what would have been refused and changes no
+response — and `off` disables the layer entirely, as a further safety
+valve against unexpected log volume.
 
-A warn record carries the endpoint, the parameter, the reason, the
-offending value's **type** — never its value — and the status the
-request went on to return anyway. That last field is the interesting
-one: a finding on a request which returned 200 is a rejection
-enforcement would introduce, while one on a request which returned 404
-is a status code enforcement would merely change, because validation
-runs ahead of the per-method decorators which produce those.
+A validation record carries the endpoint, the parameter, the reason,
+the offending value's **type** — never its value — and, in `warn`
+mode, the status the request went on to return anyway. That last
+field is what told `enforce` and `warn` apart before the default was
+flipped: a finding on a request that returned 200 was a rejection
+enforcement would introduce, while one on a request that returned 404
+or 403 was a status code enforcement would merely change, because
+validation runs ahead of the per-method decorators which produce
+those. That second case is now the enforced default's own contract
+change: a request that is both malformed and refers to a missing or
+unauthorised object answers 400 rather than 404 or 403 — see the
+[v0.7 to v0.8 release notes](../release_notes/v07-v08.md) for the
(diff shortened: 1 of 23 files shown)


mikalstill:
Addressed the review round: both fix items (unbounded finding fan-out from element naming, and lexical index ordering), all three documentation items, and four of the five consider items.

The audit-coverage item was taken as a fix rather than as documentation -- a refusal now writes a `request refused by input validation` audit event against the caller's namespace, since preserving event coverage is a standing goal here. Writing it surfaced the edge the last (informational) item had traced: the parameter name would have reached a namespace event unredacted, so it now obeys the same credential-route redaction as the log line, through a shared constant.

Eleven mutations over the new assertions, each caught by the test which names the property. The one consider not taken is the cluster CI test's base class, and the reason is in the class docstring: the refusal has to hold for an ordinary caller, so the namespaced fixture is the point rather than incidental.

sfconductor:

The conductor will merge this at `a834f690c9d3` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `a834f690c9d3`, bypassing the merge queue.

Issue #4161, opened 2026-09-10
Unreachable peer node (and local nodelock socket) can still answer 500 outside proxy_request_to_node


mikalstill:
## Summary

Thirty days of `{job="shakenfist"} |= "Server error"` on the sfcbr Loki
tenant show 266 records, all logged by `suppress_exceptions_to_client`.
Nearly all of them are one of two infrastructure conditions rather than
a fault in the handler itself: a peer node's API being unreachable, or
the node's own local nodelock daemon refusing a Unix-socket connection.
Neither is the kind of fault a 500 means, and 502 or 503 would say so
instead — `proxy_request_to_node` (`shakenfist/external_api/base.py`)
already makes exactly that distinction for the one call site it
guards, since #3743.

| n | class | deepest frame |
|---|---|---|
| 260 | `ConnectionError` | `requests/adapters.py` in `send`, reached from `proxy_request_to_node` |
| 5 | `KeyError` | `util/general.py` in `get_user_agent` |
| 1 | `ConnectionRefusedError` | `util/concurrency.py` in `_node_lock_request` |

This table is finding F9 of phase 5 of the API input validation plan
(`docs/plans/PLAN-api-input-validation-phase-05-narrow.md`), which
named the defect but did not file it (decision D28: it is a phase's
worth of work, not a step's). Re-running the same query before filing
turned up two corrections and one addition, below — filing this from
the current state of the tree rather than repeating the table as
found.

## The 260 `ConnectionError` records are already fixed, for that one call site

Re-running `{job="shakenfist"} |= "Server error" |= "ConnectionError"`
over the same 30-day window: all 260 timestamps fall between
2026-08-12 and 2026-08-19. `proxy_request_to_node` started answering
503 for exactly this condition on 2026-08-21 (#3743, `2ac50193b`, PR
#3833), and there have been zero occurrences since. So the dominant
class in the sample is not live via that path any more. The fix does
not cover every place the API proxies a request to a peer, though —
see the next section — so the underlying defect class is not closed,
only its biggest instance.

## Two more unguarded proxy call sites, found by one grep pass

`grep -rn "requests\.\(get\|post\|put\|delete\|request\)("
shakenfist/`, outside tests and outside `proxy_request_to_node`, finds
two more sites making the identical mistake #3743 fixed. Neither is
covered by that fix, and neither has log evidence yet, because nobody
has hit the failure window in the last 30 days:

[...]

Issue #3743, opened 2026-08-13
consoledata proxy to an unreachable peer node escapes as an unqualified 500 'Server error'


mikalstill:
When `GET /instances/<uuid>/consoledata` lands on a node that does not host the instance, `external_api/base.py:832` proxies the request to the hypervisor node with `requests.request(...)`. If the peer's API is unreachable, the raw `requests.exceptions.ConnectionError` propagates uncaught through every endpoint wrapper: `shakenfist.external_api.base` logs ERROR "Server error" with the traceback, and the client receives an unqualified 500.

Two problems:

1. **No peer-unreachable handling on the node-to-node proxy path.** A refused or reset connection to the peer is an infrastructure condition, not a server bug in the API that received the request. It should surface as 502/503 naming the peer node, and is arguably retryable. This is the same error-conflation family as #3373/#3522 (infrastructure failure surfaced as the wrong class of error).
2. **Diagnosability.** Same shape as the closed #3433/#3630: a raw exception escaping to a content-free 500. The log line does carry a request-id, but not the peer node or proxied URL as fields — they are only recoverable by parsing the traceback text.

Observed on sfcbr, 7 events over three days (all `gunicorn: worker [sf-api]`, each with a request-id):

- 2026-08-06 09:55 — 2 events, sf-3/sf-4 (first seen, sha 8854aa9)
- 2026-08-12 09:48:35 — 1 event (ECONNRESET mid-response)
- 2026-08-13 07:02:08, 17:48:01, 20:32:38, 20:33:08 — 4 events, sf-1/sf-2; a mix of ECONNREFUSED at connect and ECONNRESET mid-response. The 20:32/20:33 pair had no adjacent deploy, so this is not purely a deploy-restart artifact.

Example (2026-08-13 20:33:08, sf-api, request-id 758ed23b-1111-40b2-870f-56d607710d2d):

```
requests.exceptions.ConnectionError: HTTPConnectionPool(host='192.168.21.55', port=13000):
Max retries exceeded with url: /instances/3bcec52d-a3e7-4124-8afb-ddfaf15dcec6/consoledata
(Caused by NewConnectionError("HTTPConnection(host='192.168.21.55', port=13000):
Failed to establish a new connection: [Errno 111] Connection refused"))
```

(192.168.21.55 is sf-5. Why sf-5's API was refusing connections at those moments is a separate question; this issue is about how the failure surfaces.)

Loki (tenant sfcbr):

```
[...]

Issue #3523, opened 2026-07-26
external_api.base swallows a bare KeyError as 'Server error'; paired content-free util.exceptions WARNING


mikalstill:
A single request on sf-4 at 2026-07-26T14:57:01+10:00 raised a bare
`KeyError` that surfaced twice, ~200µs apart, via two loggers — and
neither log line identifies the dict lookup that actually raised:

- `shakenfist.external_api.base` logged it at ERROR through the request
  wrapper as a generic "Server error" (the opaque-wrapper pattern
  already tracked as #3433), with the traceback showing only the
  wrapper frames in `external_api/base.py` (lines 601–663);
- `shakenfist.util.exceptions` logged a content-free WARNING
  `Recorded new exception` with just the exception class — the
  duplicate-logging shape already noted for InvalidStateForTask,
  ImageMissingFromCache, BlobMissing and friends, and related to the
  tracebacks-only-at-DEBUG problem in #3371. `KeyError` is a new class
  joining that list.

Found by the nightly sfcbr error-mining report for 2026-07-26 (new
signatures #4 and #5). First seen under deploy `169899d`.

## Symptom

ERROR side (traceback truncated by the facts capture before it reaches
the raise site):

```
Traceback (most recent call last):
  File "/srv/shakenfist/venv/lib/python3.13/site-packages/shakenfist/external_api/base.py", line 663, in wrapper
    return func(*args, **kwargs)
  File "/srv/shakenfist/venv/lib/python3.13/site-packages/shakenfist/external_api/base.py", line 655, in wrapper
    raise e
  File "/srv/shakenfist/venv/lib/python3.13/site-packages/shakenfist/external_api/base.py", line 652, in wrapper
    return func(*args, **kwargs)
  File "/srv/shakenfist/venv/lib/python3.13/site-packages/shakenfist/external_api/base.py", line 609, in wrapper
    return func(*args, **kwargs)
  File "/srv/shakenfist/venv/lib/python3.13/site-packages/shakenfist/external_api/base.py", line 601, in wrapper
    return func(*args, **kwargs)
  ...
```

WARNING side, same host, 195µs earlier:

```
Recorded new exception
```

(exception class `KeyError`, no message, no traceback)

## Analysis

A bare `KeyError` reaching the API request wrapper means some handler
did an unguarded dict lookup on request data or internal state — worth
finding regardless of how rare it is, since it returns an opaque 500 to
the client.

Needs:

[...]

Issue #3433, opened 2026-07-17
external_api.base logs 'Server error' at ERROR with no exception class or traceback


mikalstill:
Found by automated log analysis (the Phase 9 error-mining report) on the sfcbr home-lab cluster. Minor, but it obscures other failures.

## Symptom

On 2026-07-16, `shakenfist.external_api.base :: Server error` produced 118 ERROR-level events (sf-1…sf-4, first seen 07:26:17) whose record is just `message == "Server error"` with **no** `exception_class` or `stack_trace` attached — unlike the sibling `external_api.base` entries (`ConnectionError`, `libvirtError`), which do carry a class + traceback.

## Why it matters

These 118 events are unattributable from centralised logging alone: you can see *that* the API hit a server error but not *what*. For a log-mining workflow they are a dead end, and they mask the real underlying failures (there was a concurrent `libvirtError` burst in the same second).

## Suggested fix

Log the wrapped exception class (and ideally the traceback) on the `Server error` path in `external_api.base`, matching the detail the `ConnectionError` / `libvirtError` paths already emit.

Loki query (sfcbr tenant):

```
{job="shakenfist"} | json | level=~`ERROR|CRITICAL` | logger_name=`shakenfist.external_api.base`
```

Related: #3371 (exception logging hygiene).

*(Issue filed with Claude Code assistance from the sfcbr Phase 9 error-mining report.)*

PR #4162, opened 2026-09-10
API input validation phase 5: narrow the handlers

mikalstill:
Phase 5 of [PLAN-api-input-validation](docs/plans/PLAN-api-input-validation-phase-05-narrow.md). Phase 4 landed in #4141.

Phase 4 made the validation layer reject. This phase deletes the thing it was built to replace: the `except TypeError` arm in `handle_authorization_exceptions`, which has been answering 400 in the interpreter's own words since 2021 (`ffecee9f9`) and is the second half of the mechanism #3612 describes. It could not go earlier because it was absorbing the malformed input phase 4 now refuses.

```python
        except TypeError as e:
            _authorization_failure_log(e).info('API request rejected as malformed')
            return sf_api.error(400, str(e), suppress_traceback=False)
```

A `TypeError` is not an authorization condition, and nothing in `flask_jwt_extended` or `PyJWT` signals one with it. The three JWT arms are untouched.

## The phase grew a sixth step, because deleting the arm was not enough

Step 2 found that `suppress_exceptions_to_client` builds **every** 500 body as `'server error: %s' % repr(e)`. So the deletion on its own would have relabelled the leak rather than removed it — an undeclared body key under the `warn` rollback would have answered:

```
500 server error: TypeError("AuthEndpoint.post() got an unexpected keyword argument 'zzz'")
```

which is #3612's motivating string wearing a different status code. It is a second, independent leak site common to every exception class: it is also what put `KeyError('arch_string_raw')` and unreachable-node `ConnectionError` reprs into caller-visible bodies.

**Every 500 body is now the bare string `server error`**, for any cause, on any endpoint (D31). Nothing is lost that mattered — the `Server error` log line already carries `exception_class`, the full traceback, method and path, and the on-disk record under `/srv/shakenfist/exceptions/` carries the correlation fields. What a caller loses is stated in the release note rather than left to be discovered: you can no longer tell which exception occurred from a response body alone.

The step 2 agent did not fix this itself, correctly — it asserted what was true, exempting one marker from its `INTERPRETER_TEXT` sweep with the exemption named in the test docstring, and brought the contract decision back rather than taking it. Step 6 removes the exemption.

## What callers see

* **Under `enforce`, the default: nothing changes.** An undeclared body key is refused by name before the handler is reached.
* **Under the `warn`/`off` rollback**, such a key now answers 500 with a recorded exception instead of 400 with interpreter text (D25). The rollback buys "requests that were working keep working", not a tidier answer for the ones that were not.
* **Every 500, everywhere**, loses its exception detail.

## Also in this phase

* **#3523 is fixed and closed.** It asked for a Loki drill-down to find its raising frame; thirty days of `sfcbr` named it — `KeyError: 'arch_string_raw'` in `util/general.py:get_user_agent`, via `proxy_request_to_node` on `GET /instances/<uuid>/consoledata`. `py-cpuinfo` returns whatever its probes collected. The issue's logging half was already fixed by the #3433/#3590 work, which is why the traceback reaches the raise site at all.
* **Two dead `requires_namespace_exist_if_specified` applications removed.** Phase 4 found them and filed no issue, so the record lived only in a phase plan. All twelve applications were surveyed, not just the two named.

## Two corrections this phase made to its own plan

Recorded in the plan at source rather than only in notes, because a stale finding is how the next reader inherits an error.

* **F9 was substantially wrong.** It claimed a node being unreachable answers 500 "260 times in a month". That reads a 30-day aggregate as a rate: 252 of the 260 are a single burst on 2026-08-19, there are none after, and [`2ac50193b`](../../commit/2ac50193b) (#3743) landed 2026-08-21 and had already fixed that path. What survives is the local nodelock socket — which recurred *after* that fix — and two proxy call sites the survey missed because neither goes through `proxy_request_to_node`. Filed as #4161, scoped to what is still true.
* **Definition of done item 5 took two attempts to state.** First a `grep` for `repr(e)` that step 6's own explanatory comment falsifies; then an over-correction that would have condemned the nineteen handlers deliberately answering `str(e)` for a Shaken Fist exception class, where the message is the API's vocabulary rather than the interpreter's. Both attempts stay visible — editing whichever item fails is how a falsifiable list stops being one.

## Verification

* Every new assertion mutation tested. Restoring the deleted arm fails two tests, one of them in the frame the arm lived in, so it pins the property whatever the response is shaped into further out. Restoring `repr(e)` fails the unexempted sweep.
* The `TypeError` enumeration D25 rests on was re-run independently rather than taken from the plan, including the AST check that no handler has a parameter without a default — the one number the plan got wrong on its first attempt.
* Full unit suite: 4673 tests, 0 failures. `pre-commit run --all-files` clean.
* All 13 definition-of-done items met.

Operators whose callers break can still set `API_VALIDATION_MODE=warn` or `off`, with the caveat above.

### DIFF
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -317,6 +317,18 @@ have been refused and changes no response — and `off` disables the
 layer entirely, as a further safety valve against unexpected log
 volume.
 
+Neither rollback restores the handler-facing behaviour from before
+enforcement existed, and an undeclared body key is where that shows.
+Under `warn` or `off` the key still reaches the handler, which raises
+`TypeError` on the unexpected keyword argument, and that now answers
+`500` — recorded like any other server exception, with nothing about
+it in the response body — rather than the `400` carrying the
+interpreter's own message it used to answer. An operator choosing the
+rollback gets requests that were working kept working, not a tidier
+answer for the ones that were not; see the [v0.7 to v0.8 release
+notes](../release_notes/v07-v08.md) for the caller-visible shape of
+it.
+
 Because a refusal is answered from outside every per-method decorator,
 `log_token_use` never runs for one. The refusal writes its own
 `request refused by input validation` audit event against the caller's
--- a/docs/developer_guide/writing_an_endpoint.md
+++ b/docs/developer_guide/writing_an_endpoint.md
@@ -329,6 +329,17 @@ answer for the ones that were not; see the [v0.7 to v0.8 release
 notes](../release_notes/v07-v08.md) for the caller-visible shape of
 it.
 
+Budget for the operator-visible shape too. A refusal under `enforce`
+is an audit event; a `TypeError` under `warn` or `off` is a server
+fault, so it writes an exception record under
+`/srv/shakenfist/exceptions/` and logs at ERROR. `record_exception`
+keys those records by a hash of the traceback and the `TypeError`
+message embeds the caller's own key name, so a client sending varying
+undeclared keys writes one record per distinct key and one ERROR line
+per request — which is client input driving disk usage and quite
+possibly alerting. It is a reason to treat the rollback as temporary
+rather than a resting state.
+
 Because a refusal is answered from outside every per-method decorator,
 `log_token_use` never runs for one. The refusal writes its own
 `request refused by input validation` audit event against the caller's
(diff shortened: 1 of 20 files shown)


sfconductor:

The conductor will merge this at `3c4874b15207` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `3c4874b15207`, bypassing the merge queue.

PR #4160, opened 2026-09-10
Record where every plan's phases landed, and install the push audit block

mikalstill:
Step 4c of phase 4 ("Fleet backfill") of
[PLAN-push-audit-phase.md](https://github.com/shakenfist/development/blob/main/docs/plans/PLAN-push-audit-phase.md)
in shakenfist/development. Twenty-one plans gain a `Merged` column, two
prose-phase roadmaps gain `Merged:` lines, three failing plans are
fixed, and `PLAN-TEMPLATE.md` gains the `plan-push-audit-phase` block
at v3.

The record is what phase 5 of that plan reads to assemble an audit
range. **No consistency check reads it**, so review is the only thing
that will catch a wrong SHA -- hence the assertion table below.

## The backfill set

Nineteen carriers, derived by reading each plan's Execution table for a
phase that *is* the push audit, not by grepping for the runbook.
`PLAN-ci-cloud-sizing.md` becomes the twentieth once this branch gives
it a phase, and `PLAN-transient-capacity-refusals.md` the twenty-first.

Correctly excluded from the backfill: `PLAN-netserv.md` (has a
`### Push audit` section but no phases the check can read) and
`PLAN-sql-pushdown-filtering.md` (names the runbook, carries no audit
phase). Also checked and excluded:
`PLAN-replace-last-cluster-operation.md` phase 5, "Documentation and
final audit", which is an Explore sweep plus pre-commit rather than a
push audit.

## Phases written as prose, not table rows

The shared block this branch installs says that where a plan's phases
are prose sections rather than table rows, the record is a `Merged:`
line in the phase's own section. Three plans sit on that path and were
missed by the first pass.

`blob-storage-roadmap.md` is the material one: its phase 1 has landed,
so its phase 4 push audit had nothing to read. Reconstructed from the
merged pull request list cross-checked against the first-parent
history -- `5b1483be2` (#2929) landed the migration framework,
`47ccb2d4a` (#2930) the hash storage itself, and `3c413ddc4` (#3089)
the residual etcd usage in the scheduled tasks. Its remaining phases,
and all four of `api-query-batching-roadmap.md`'s, open with an em
dash.

`PLAN-netserv.md` has no phases at all, so rather than being silently
skipped it now records *why* there is nothing to record, and what the
shape will be when its phases are cut.

## Where the plan's own survey was wrong

Its section claimed *none* of the nineteen records a landing commit in
any shape. `PLAN-kerbside-vdi-tokens.md` already did: its phase 10
[...]

PR #4166, opened 2026-09-10
The suite waits out a transient capacity refusal, and says so

mikalstill:
Phase 2 of `PLAN-transient-capacity-refusals`, planned and implemented. Six commits: the plan, then steps 2a, 2b, 2c, 2d and 2e of its own step table.

A capacity refusal in CI is usually a statement about a moment rather than about the cluster — the capacity the refusing node is holding is very often already being returned by a sibling test that is mid-delete. The suite called `create_instance()` raw at every site, so every one of those was a hard failure. This makes them survivable and writes down how long survival took, because that number is what phase 5 decides the server-side-queue question from.

## The survey changed the design, not just the wording

The master plan's phase 2 section was written on 2026-09-08, before phase 1 executed. Two of its claims were wrong in ways that would have produced a working-looking wrapper that did the wrong thing. Both are corrected at their source in that section, with the correction marked.

**The wait predicate.** The section said to poll `/admin/resources` and proceed when `per_node[target]['cpu_available']` covers the create, or `total['cpu_available']` for an unpinned one. Both halves are wrong:

- `cpu_available` is derived from the live overcommit arithmetic (`scheduler.py:1086`), not from the capacity row's `limit_cpus` that admission actually refuses against. The endpoint publishes **both** deliberately, and the comment three lines above `cpu_limit` says in as many words that the field has no fallback precisely so a reader can see the two ledgers disagree. A wrapper reading only the published figure stops waiting during exactly the window this plan exists to describe.
- `total['cpu_available']` is a sum across nodes (`:1091-1093`) and an instance has to fit on one. It is satisfied by a cluster with one spare thread on each of ten nodes.

D8 replaces both: per node, `min(cpu_available, cpu_limit - cpu_committed)` when there is a capacity row, and the best single node rather than the total. **Deleting the three-line `cpu_limit` clause fails exactly one test**, which was verified by doing it rather than by reading it.

**The cross-repo obligation.** The section said this phase carried the same operator-push obligation the sizing plan's phase 1 did. Only the job-log print does. `smoke-cluster.yml:212` creates `/srv/ci/traces` on the primary and chowns it to the base-image user, `:550` scp's the whole directory into the artifact bundle, and the suite runs on the primary as that user — so the wait data reaches the bundle with no change in `shakenfist/actions` at all. That let the phase be restructured so **nothing depends on the cross-repo step**, and the evidence is readable from a downloaded bundle the moment this merges. Which is how phase 1's own definition-of-done item 9 was ultimately verified.

Three smaller corrections: the call-site count is 117 across 41 files, not 115 across 39, and the two directories the section named hold only 88 of them — `smoke_ci_tests/` was omitted entirely, and it is the tier that runs on the smallest cloud. `retry_while_transient` is not reusable as written. And of the two ride-along test fixes, one was already done and the other was mis-described.

## What landed

- **2a** — `wait_for_capacity()` in `retries.py` and `BaseTestCase.create_instance()`. The poll is injected as a callable so `retries.py` keeps the property its docstring claims: it imports nothing from the suite or from `shakenfist_client`, so the unit tests can load it by path. Only `InsufficientResourcesException` is waited on; the affinity 409 is a subclass of `LowResourceException` on the server, kept distinct only by clause ordering in the handler, and does not become satisfiable by waiting. Sixteen unit tests on a fake clock.
- **2b** — all 117 call sites routed, plus an `ast` guard that fails on any raw call without a `# raw-create: <reason>` marker carrying a **non-empty** reason. An allowlist that can be extended by copying the line above it is not an allowlist.
- **2c** — one JSON line per wait appended to `/srv/ci/traces/instance-waits.jsonl`, and `--waits` in `tools/ci_headroom_report.py`. One line per wait rather than per run, so concurrent stestr workers need agree on nothing and a dead worker loses at most its own in-flight line. Every write failure is swallowed: an instrument may never fail the thing it measures.
- **2d** — one convenience pin removed from `test_imagefetch.py`, its sibling repointed at `inst1['node']` where co-location really is the assertion, and the missing `addCleanup` added to `test_system_namespace.py`, which subclasses `BaseTestCase` and so has nothing tearing its instance down for it.
- **2e** — `docs/developer_guide/ci.md`, and one bullet in `AGENTS.md` for the marker convention alone.

## One call site went back to the raw client

`test_coalescing.py`'s mesh burst pins creates round-robin across the hypervisors inside a `try` that records a capacity refusal and carries on — because, as its own comment says, a forced placement onto a node the scheduler will not admit is a capacity outcome on a shared CI cluster rather than a coalescing defect. Wrapping it broke it twice: the wrapper's deadline failure is a test failure rather than an `APIException` that `except` clause could catch, so one node's refusal would have failed the whole burst; and waiting up to seven minutes per refused create would serialise a burst whose simultaneity is the thing under test.

It carries a marker saying so. This is the case D15 wrote the marker for, arriving earlier than the decision expected — D15 named the sizing plan's phase 3 saturation tests, which do not exist yet. The other burst in the same file keeps the wrapper: it has no tolerant handling and a refusal fails it outright today, so a wait can only improve it.

## Where to push back

The back brief asked which half of D8 was most likely to be wrong. The answer that came back is better than the one I would have given: the `cpu_limit` clause is sound, and the weaker half is the per-node `max` for an **unpinned** create. It reads the roster the scheduler also reads, but the scheduler's choice is subject to pre-filters the predicate does not model — queue length, `cpu_max_per_instance`, memory, disk, affinity. So an unpinned wait can be satisfied by a node the scheduler then declines, producing a create that refuses again and one logical wait split across two records.

That is bounded by the 420 s deadline and is visible in the data as a satisfied wait followed by another attempt. The right move is to look at phase 5's first real numbers before modelling more of the pre-filter, not to guess now — but a reviewer who disagrees should say so, because it is cheap to change today and expensive once the baseline is being read.

## Test plan

- [x] `pre-commit run --all-files` after every step
- [x] 16 new unit tests for the wait predicate, on a fake clock and a scripted poll
- [x] D8 mutation check run independently of the implementer: deleting the `cpu_limit` clause fails exactly one test, restoring it passes 16/16
- [x] AST guard verified by breaking it three ways — unmarked call fails, empty reason still fails, revert passes
- [x] All five `--waits` states exercised end to end: populated, malformed-among-good, empty file, absent file, flag not given. The last three report unknown, never zero
- [x] `test_the_suite_still_probes_cluster_headroom` still passes, which is what holds the load-budget prose up
- [ ] A real CI run — this branch has not been through one, and that is step 2g

The functional suite cannot run from a worktree, so 117 mechanically edited call sites are covered by the AST guard, flake8 and reading. A full CI run before the phase closes out is not optional here.

## Not in this PR

**2f**, the one-line addition to `ci_headroom_collect.sh` in `shakenfist/actions` that prints the summary in the job log. It needs a push to another repository, and that file changed on disk during this work, so it wants a fresh read rather than a remembered one. Nothing here depends on it.

**2g**, the close-out, which needs a real bundle read back with `--waits` and an Outcome section recording it — including the D8 caveat above.

## Ordering note

This PR and #4160 both edit `PLAN-transient-capacity-refusals.md`'s Execution table and the plan's row in `docs/plans/index.md`. #4160 appends a phase 7 and adds a `Merged` column; this one links phase 2's file. No phase renumbers, so there is no correctness hazard, but whichever merges second will need a mechanical rebase. #4160 is currently clean, so merging it first is the cheaper order.

https://claude.ai/code/session_015y4fDJF57hRTfkfxn16Bve

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -133,6 +133,12 @@ metrics. The headlines:
   the operator recovery path; see
   [docs/operator_guide/node_health.md](docs/operator_guide/node_health.md).
 
+- **A raw `create_instance()` call in the CI suite needs a `# raw-create:
+  <reason>` comment**, or an `ast`-based unit test fails the build; use
+  `self.create_instance()` instead unless the test must see a 507
+  directly. See
+  [docs/developer_guide/ci.md](docs/developer_guide/ci.md#creating-instances-in-the-functional-suite).
+
 - **An unlanded plan may already own the code you are changing.** A fix
   landed across a partially implemented plan has to be unpicked later,
   which is why the automated fixer reads `docs/plans/` before writing
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -133,8 +133,9 @@ metrics. The headlines:
   the operator recovery path; see
   [docs/operator_guide/node_health.md](docs/operator_guide/node_health.md).
 
-- **A raw `create_instance()` call in the CI suite needs a `# raw-create:
-  <reason>` comment**, or an `ast`-based unit test fails the build; use
+- **A raw `create_instance()` in the CI suite needs a `# raw-create:
+  <reason>` comment**, or an `ast`-based unit test fails the build. That
+  covers a bare reference passed to `assertRaises` as well as a call; use
   `self.create_instance()` instead unless the test must see a 507
   directly. See
   [docs/developer_guide/ci.md](docs/developer_guide/ci.md#creating-instances-in-the-functional-suite).
--- a/docs/developer_guide/ci.md
+++ b/docs/developer_guide/ci.md
@@ -128,6 +128,56 @@ that connects and then stops responding; without this the job would stall
 until the CI runner's own timeout killed it, with no indication of which test
 was to blame.
 
+## Creating instances in the functional suite
+
+Every functional test creates instances through `self.create_instance()` on
+`BaseTestCase` (`shakenfist/deploy/shakenfist_ci/base.py`), never through a
+client's `create_instance()` directly. The wrapper forwards its arguments
+unchanged and waits out a transient `InsufficientResourcesException` (a 507)
+rather than failing the test on it.
+
+The reason is that a 507 on a shared CI cluster is usually a statement about
+a moment, not about the cluster: the capacity the refusing node is holding is
+very often already being returned by a sibling test that is mid-delete.
+Failing the test on that makes a run's pass/fail a coin flip on sibling
+timing. The wrapper polls `/admin/resources` for room for up to
+`CLUSTER_HEADROOM_WAIT` -- 420 seconds, the same deadline
+`cluster_ci_tests/test_namespace_claims.py` already used for the same
+reason -- and only for a plain 507: an `AffinityConstraintUnsatisfiable` 409
+is never caught, because that refusal does not become satisfiable by
+waiting.
+
+If the deadline passes, which is what it means for a test that is genuinely
+asking for more than the cloud has, the test fails with the last refusal's
+body, the node the wrapper waited for (or "unpinned" if the create had no
+placement), and the `per_node` capacity roster from the last successful
+poll -- so the failure carries the diagnosis a reader needs rather than a
+bare 507.
+
+A raw call -- `self.test_client.create_instance(...)`, a local client,
+another test's client, anything that does not go through
+`self.create_instance()` -- skips the wait silently, and needs a
+`# raw-create: <reason>` comment immediately above it, with a non-empty
+reason. A unit test in `shakenfist/tests/` (`test_ci_raw_creates.py`) walks
+the suite's source with `ast` looking for exactly this, rather than
+importing the suite, which depends on `shakenfist_client` -- not a test
+dependency of this repository. An empty reason fails the same as no marker
+at all, so the allowlist cannot grow by copy-pasting an existing line.
+
+The marker exists for a caller that must see the 507 rather than have it
+waited out. `cluster_ci_tests/test_coalescing.py`'s mesh burst is the
+current example: it pins instances round robin across the hypervisors
+inside a `try` that tolerates and records a capacity refusal as a normal
+outcome on a shared cluster, and carries on -- wrapping that create would
+break it twice, since the wrapper's own deadline failure is a test failure
+rather than an `APIException` the `except` clause could catch, and waiting
+up to seven minutes per refusal would serialise a burst whose simultaneity
+is the thing under test. The other reserved use is the wrapper's own call
+inside `base.py`, which has to reach the client somehow. The CI cloud
+sizing plan's phase 3 saturation tests are expected to need the marker
+too, to assert that a genuinely full cluster refuses rather than have the
+refusal waited away.
+
 ## CI headroom instrumentation
 
 Phase 1 of `docs/plans/PLAN-ci-cloud-sizing.md` (see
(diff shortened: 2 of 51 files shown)


mikalstill:
Pushed `9bc08af9`, fixing the one unit test which failed on the first CI run.

**Root cause.** `test_the_deadline_fails_with_the_refusal_and_the_roster` patched the process-wide `time.time` with a three value script sized for the three clock reads `create_instance()` makes. That method logs between its second and third read, and on Python 3.11 `logging.LogRecord.__init__` reads `time.time()` once per record — so the log line consumed the third value and the next read raised `StopIteration`. Python 3.13 reads `time.time_ns()` there and consumes nothing, which is why it passed on a workstation and failed on CI's 3.11. Not flakiness, and not a defect in the wrapper it appeared to blame.

**Fix.** The test reached for a global patch because `create_instance()` offered no seam, so it has one now: `BaseTestCase._now()`, beside `_uniquifier()` which the same harness already overrides for the same reason. The test scripts the harness clock and touches no global, so it is correct on either interpreter rather than accidentally correct on one.

**A second problem found while fixing it.** `_WrapperHarness` had no `id()`, which `_append_capacity_wait_trace()` reads for the record's `test_id` — so every wrapper test took the swallowed exception branch instead of writing a line. The JSONL wait trace is this phase's only evidence output and had no coverage at all while appearing to have some. The harness gains `id()`, the trace file is redirected to a temporary directory rather than `/srv/ci/traces`, and two tests now assert that a wait writes one well formed line carrying every field `--waits` reads, and that an unwritable trace still never fails the create it measures.

All three changes are mutation tested: reverting the seam fails the deadline test and only that one, renaming a field in the trace record fails the well formed line test and only that one, and removing the swallow fails the unwritable trace test and only that one.

**Two things the first run settled, one it could not.** Smoke tests passed, which is the real cluster run exercising the 117 rewritten call sites that the test plan had unticked. The bundle also confirms `traces/` reaches the artifact with no change in `shakenfist/actions`, which was the D14 plumbing claim.

It did **not** produce an `instance-waits.jsonl`, because nothing waited: the run saw no client visible 507 (the 9 `507` hits in the suite log are coincidental substrings in job ids and byte counts, and the census's 13 `instance placement denied` events were each absorbed by the scheduler finding another candidate). A healthy smoke run cannot produce a wait by construction, so step 2g still needs a cluster tier run under real capacity pressure. The absent file reports as unknown rather than zero, which is the designed behaviour rather than a gap.

mikalstill:
Pushed `3a8c30bf`, addressing all four `fix` items. The six `consider` items were all cheap enough to take as well; only the one `none` item is untouched, deliberately.

**1 and 2 are one defect, and the review was right about which half was weak.** The PR text argued the weaker half of D8 was the per-node `max` for an unpinned create. It was not — the weak half was that the predicate modelled cpus at all, alone. The scheduler pre-filters on three dimensions and turns any of their `LowResourceException`s into the same 507, and CI instances are 1 vCPU but 1-2 GB with 8 GB disks, so the two dimensions D8 did not model are the two most likely to bind. That is not a rougher version of the right answer; it is a predicate that is permanently satisfied while the server is permanently refusing. `wait_for_capacity()` now takes the whole request and a node satisfies it only when every dimension fits **on that node** — the `total[cpu_available]` error one level down. `cpu_max_per_instance` and `ram_max_per_instance` (item 11) join the ledgers for the same reason, differing only in how a zero reads: a missing cpu metric publishes as 0, where the memory one is legitimately 0 on a full node. The record now names the `binding_dimension`, because a duration with no cause attached to it cannot be acted on in phase 5.

Widening the predicate shrinks the window but cannot close it — the point is that some refusals are invisible from `/admin/resources` — so the loop is paced as well. The first wait after a refusal may still return immediately, because capacity returning between the refusal and the poll is the common case and the whole reason this phase exists; every wait after that sleeps at least one poll interval. `MAX_CREATE_ATTEMPTS` is the hard ceiling you asked for, and the failure it raises says the refusal is for something the predicate cannot see rather than blaming the cluster.

**3 was worse than reported, and the fix is smaller.** The guard now walks attribute nodes rather than call nodes, which finds both shapes and finds each exactly once — a call's `func` *is* the attribute node. My first attempt kept an identity map to recover the call's line number for a call site; mutation testing showed that mutation changed nothing, and checking empirically confirmed CPython gives a call and its func the same `lineno` in every shape I could construct. So the map was dead weight and is gone. All six sites carry markers naming the specific 400/404/409 each asserts.

**4 is fixed, and finding it was fair.** A test now drives the whole wrapper loop against the real `wait_for_capacity` with a server that refuses every create while the roster reports ample room in all three dimensions, and asserts the attempts are bounded and the clock really moved between them. There is a positive control beside it, so a wrapper that simply stopped retrying would not pass.

**Verification.** Every property added here was mutation tested — eleven mutations, each broken on purpose and confirmed to fail the test that claims to enforce it and only that test. That pass is also what caught a bug I introduced: `best_node()` seeded its best-so-far with a coverage of `0.0`, so a node whose coverage was exactly zero never beat the seed and the wait recorded "nothing bound" for a create refused for cpus. Full unit suite and `pre-commit run --all-files` are green.

Items 5 through 9 and 11 are all in. Item 10 is intentional and stays.

Still outstanding and unchanged: this branch has not been through a real CI run, which is step 2g.

PR #4170, opened 2026-09-11
CI cloud sizing phase 3: saturation coverage

mikalstill:
Phase 3 of `PLAN-ci-cloud-sizing`: make the scheduler's capacity refusals
*assertable* rather than incidental, so a 507 in CI is either a test that
expected one or a genuine regression.

Seven commits, one per plan step.

## What this adds

- **`assertRefusedAtStage()`** (`shakenfist_ci/base.py`) -- a single place
  that encodes the 2026-09 refusal contract: HTTP 507 with body
  `No nodes remaining at scheduling stage <stage>`, optionally with a
  `: <detail>` suffix, which it returns. Accepts either an
  `APIException` or a `(status, body)` tuple, and deliberately does not
  go through `retries.py` -- a refusal is the expected result here, not
  a transient to be retried away.
- **Three deterministic impossible-request tests**
  (`cluster_ci_tests/test_saturation.py`) covering `sufficient_idle_cpu`,
  `sufficient_idle_memory` and `sufficient_free_disk`. Each sizes its
  request one unit beyond a *ceiling* read from `/admin/resources`
  (`cpu_limit` falling back to `cpu_hard_max`, `ram_max`, and the
  cluster-wide disk margin), never beyond headroom -- headroom grows when
  a sibling stestr worker deletes an instance, which is exactly the flake
  this phase exists to remove.
- **A node-fill test** that fills one hypervisor to its ledger limit and
  asserts the next targeted create is refused, recording via `addDetail`
  which of the three refusal paths the cluster actually took, and
  asserting the ledger returns afterwards.
- **13 unit tests** for the four capacity-stage predicates
  (`_has_sufficient_cpu`, `_has_sufficient_ram`, `_has_sufficient_disk`,
  `_has_idle_disk_bandwidth`), including both sides of each boundary.
  `test_scheduler.py` goes 94 -> 107 tests.

## Plan corrections found while implementing

Three of the plan's own facts were wrong and are corrected at source,
with `*Correction (2026-09-10):*` paragraphs rather than overwrites:

- **D24 sized memory and disk from headroom, not ceilings.** Following it
  would have manufactured the flake the phase exists to remove.
- **D27/D28 assumed one 507 message.** There are two
  (`No nodes remaining at scheduling stage ...` from the stage pre-filter,
  `no node had capacity for this instance, N candidates refused it` from
  the atomic guard) and a third possible answer, the async re-place abort
  `Requested node lacks resources`.
- **D23's `cpu_available <= 0`** is not generally true; the two figures
  come from different clocks.

The scarcity inventory in phase 0 is re-checked against current data
(step 3f), and phase 0's ask for a deterministic #3565 affinity
reproduction is explicitly declined with reasoning and handed to phase 4
as named debt rather than silently dropped.

## #3772 is reopened

The first commit corrects the plans' claim that
[#3772](https://github.com/shakenfist/shakenfist/issues/3772) is fixed.
It was closed by hand on 2026-09-09 with no comment and no claiming
commit; three recurrences postdate it, one with the #4087 warm-up fix
demonstrably live and two of three hypervisors pinned at their ledger
ceiling for a whole 1665 s run. The closure was an error and the issue
is reopened.

## Not marked Complete

Two DoD items can only be answered by a merge run -- which of the three
refusal paths a real full node gives, and the `slim-tier` skip rate -- so
the Execution table and `index.md` still say In Progress. Flipping them
now would be the "a status column is a claim, not evidence" failure.
The close-out section lists the five things that remain.

Note the branch is 8 commits behind `develop`; no conflicting files.


mikalstill:
## Review addressed in 43abd9e5b

Eight of the nine items taken, one declined with reasoning, plus one defect the review did not find.

| # | Item | Disposition |
|---|---|---|
| 1 | `[FIX]` fill placement never asserted | **Taken.** `_create_fill_instance()` now asserts `inst['node']`, the way `test_nodes.py:125` does. |
| 2 | Ledger-return failure text vs. a queue backlog | **Taken.** Skips while any of this test's own instances are still listed on the node; fails only once they are gone. |
| 3 | Whole-cluster listing on every 5s poll | **Taken.** Rate limited to once per 30s plus one read at each deadline (~36 listings down to ~7 on a full-deadline run). |
| 4 | Sizing arithmetic untestable | **Taken.** Moved to `shakenfist_ci/sizing.py`; 17 unit tests in `shakenfist/tests/test_ci_saturation.py`, loaded by path. |
| 5 | Unguarded `apiclient` symbols | **Taken.** AST guard in the same new file, pinning both exception names and `delete_instance(async_request=...)`. |
| 6 | `cpu_max_per_instance` unused | **Taken.** The CPU test skips, naming both figures, when the smallest published cap is below the computed request. |
| 7 | No-row branch and re-place discount untested | **Half taken, half declined** — see below. |
| 8 | Tautological assertion and docstring | **Taken.** |
| 9 | Polled `addDetail()` overwrites | **Taken.** `addDetailUniqueName()` for polled reads, plain name for the single pre-fill read. |

### Item 7, declined half

The `placement_filter()` re-place discount is already covered, at the `find_candidates()` level rather than the predicate level, by `test_scheduler.py`'s `test_prefilter_does_not_charge_an_instance_for_itself` (CPU, `:743`) and `test_ram_prefilter_does_not_charge_an_instance_for_itself` (RAM, `:1038`). Both sit exactly on the boundary the discount moves — `used_cpus == limit_cpus` with the instance's own vCPUs in the committed figure — so a predicate-level copy would catch no failure the existing pair misses.

The other half was a real gap and is taken: nothing asserted the no-capacity-row refusal's *reason dict* (`capacity_row_present: False`, `limit_cpus` falling back to the live `hard_max_cpus`), and the memory predicate had no P7 coverage at any level. Three tests added, taking the file to 108.

### One the review did not find

`one_unit_beyond()` justified its `math.floor` by claiming a truncating `int()` cast could round a negative reading back into servable range. It cannot: truncation rounds a negative value *towards* zero, so it returns a figure which still strictly exceeds the reading, and with `minimum` clamped at 1 the two are indistinguishable for every input. The code was right and its stated reason was wrong, which is worse than no comment. The docstring now says what is true, and the tests assert the two properties a caller actually depends on — the result strictly exceeds the reading, and stays a size a create could ask for — rather than particular return values.

### Verification

All 17 new sizing tests are mutation tested: dropping the increment, dropping the minimum clamp, reading a null `cpu_limit` as zero, sizing memory from headroom, inlining a helper back into the functional suite, losing `async_request`, and renaming a client exception are each caught by at least one test, and the clean tree passes. 4717 unit tests pass; `pre-commit run --all-files` passes all ten hooks. The review pass is written into the phase plan's Outcome section, as CLAUDE.md asks for at the end of a plan.

PR #4181, opened 2026-09-12
Add an sf_claim ansible module for namespace capacity claims

mikalstill:
Adds `sf_claim` to the shakenfist ansible collection, managing one
namespace capacity claim: `state: present` creates the claim when the
namespace holds none and re-dates it when it does.

## Why

This is step 5 of `PLAN-claim-coverage-and-sizing` phase 1 in
shakenfist/private-ci, whose decision `D10` gives mach33labs/33fl
ownership of the standing capacity claim for the static CI runner
fleet on sfcbr — 32 vCPU / 65536 MB / 1200 GB, re-dated on every
reconcile with a 30 day expiry.

The claim belongs to the play rather than to the CI conductor because
the play owns the fleet spec it is sized from, so a claim derived from
the same list that builds the instances cannot drift from the fleet it
covers. That is the one property neither alternative offers at any
price: the fleet has already shrunk once under cluster pressure
(four shakenfist static runners to two, 2026-08-12) and a
conductor-side constant would have gone on claiming for four.

`D10` also prices the alternative it rejects. `sf-client namespace
claim create` has no default expiry and the server permits only one
active claim per namespace, so the `command:` form is not one task but
a read-then-create-or-update: list as JSON, parse, branch, and
hand-write `changed_when` and `failed_when` for both branches — three
tasks and two JSON expressions whose correctness depends on the CLI's
output shape, in a play whose very first task is already
`sf_namespace`.

## What is here

- `shakenfist/deploy/collection/plugins/modules/sf_claim.py`
- `shakenfist/tests/test_ansible_sf_claim.py` — 25 unit tests, using
  the `test_ansible_sf_instance.py` idiom
- `shakenfist/deploy/ansible_module_ci/006.yml` — a lifecycle scenario,
  picked up automatically by `ansiblemoduletests.sh`
- Collection README and `docs/user_guide/ansible.md`

## Three decisions worth a reviewer's attention

**It splits `namespace` from `auth_namespace`, which deviates from the
collection's convention.** Every other module overloads a single
`namespace` parameter as both the target and the authenticating
identity — `sf_network.py:70` says "The namespace the network belongs
to / authenticate as". Claim management is admin-only, so the caller is
*always* acting on somebody else's namespace, and that overload cannot
express it. 33fl is unaffected because it authenticates from `sfrc` and
leaves all three connection parameters unset, but this is a real
divergence and it is documented in both READMEs.

**A re-date alone reports `changed`.** Extending a standing reservation
is a change to cluster state that other workloads' admission decisions
depend on, and reporting `ok` would hide the one thing the play exists
to do. A play that does not want the weekly renewal in its change count
can set `changed_when: false`; the module should not decide that for
it. The move is tested against the server's own `expires_at` from
before and after the PUT, so the control host's clock is never compared
against the cluster's — `expires_in_seconds` is a duration against the
cluster clock and `expires_at` is an absolute timestamp on it.

**A lapsed claim is replaced rather than reported as an error**, and
the ordering is deliberate: the replacement is created *first* and the
lapsed rows reaped afterwards, because the create is the half that can
be refused. An expired claim holds no `claimed_*` capacity, so nothing
is released by the reaping and nothing is lost if the create is
refused. Extra *active* claims are left alone and only logged —
deleting a claim that is holding capacity is not something to do as a
side effect of a renewal.

Note this is the opposite of what the CI conductor does with a lapse,
on purpose. The conductor never re-creates, because a re-create is a
guarded admission that can be refused mid-job for a runner already
serving somebody's work. A weekly reconcile whose entire purpose is
restoring coverage is a different case.

A 507 surfaces as a task failure carrying a structured `refusal`
(`status_code`, `text`), so a play can `rescue` a capacity refusal
explicitly rather than parsing a message, and can tell 507 (durable)
from 503 (retry).

`state: absent` is implemented plainly, with no `force` gate. The risk
is not in the typing but in the asymmetry — releasing capacity takes
effect immediately while winning it back is a fresh guarded admission a
busy cluster may refuse — so that is documented loudly rather than
guarded with a parameter that would suggest the opposite diagnosis.

## Verification

`pre-commit run --all-files` clean (all ten hooks, including
`tox -e flake8`, the full `tox -e py3` unit suite and `tox -e mypy`),
plus `tox -e docs` and a YAML parse of the `DOCUMENTATION`/`EXAMPLES`/
`RETURN` blocks — that parse caught a real bug, a `: ` inside a plain
scalar that broke the `expires_in_seconds` docs.

**Not run:** the functional cluster CI, so `006.yml` has never executed
against a real cluster; `validate-modules` (ansible-lint's config here
does not cover `plugins/modules/` and there is no sanity-test target);
and nothing was run against a live Shaken Fist API.

## Merge ordering

**This should merge before mach33labs/33fl's `static-runner-claim`.**
33fl's `ansible.cfg:13` points `library` at a sibling shakenfist
checkout, so until this lands `ansible-playbook --syntax-check
static_runner.yml` there fails with `couldn't resolve module/action
'sf_claim'` — which is exactly what it does today, and was checked.

https://claude.ai/code/session_01MQUXhVAGyR8YDji93oAKQJ

### DIFF
--- a/docs/user_guide/ansible.md
+++ b/docs/user_guide/ansible.md
@@ -21,6 +21,20 @@ The modules import the `shakenfist_client` python SDK and call the Shaken
 Fist REST API directly — they do not shell out to `sf-client`, and the
 control node never needs the Shaken Fist server package installed.
 
+`sf_claim` is the exception to that `pip3 install`. It calls the namespace
+capacity claim verbs, which are on the client's `develop` branch and are not
+in any release yet, so there is no minimum released version to ask for. A
+control node which uses `sf_claim` installs the client from git instead:
+
+```bash
+pip3 install git+https://github.com/shakenfist/client-python@develop
+```
+
+which is how the Shaken Fist CI conductor installs it. The module checks the
+client for those verbs and fails with this advice, naming the verb it could
+not find, rather than raising an `AttributeError`. Every other module works
+with the released client.
+
 ???+ note
     This example installs the Shaken Fist client in the system pip so that it
     is globally available to all Ansible users. The system pip is protected on
@@ -75,7 +89,11 @@ omitted, the module auto-discovers credentials from the environment and
 `sf_claim` is the exception, because claim management is administrator
 only and the caller is therefore acting on somebody else's namespace:
 there `namespace` names the namespace the claim covers, and the namespace
-to authenticate as is `auth_namespace`.
+to authenticate as is `auth_namespace`. Because that rename makes it easy
+to supply `api_url` and `key` while leaving `auth_namespace` unset,
+`sf_claim` treats a partly specified connection as an error rather than
+quietly falling back to the control node's ambient credentials — which
+might be pointed at another cluster entirely.
 
 ## Namespaces
 
@@ -298,6 +316,7 @@ does.
 | limit_memory_mb<br/>*integer* | The instance memory, in megabytes, the namespace may hold at once. Required when `state` is `present`. |
 | limit_disk_gb<br/>*integer* | The instance disk, in gigabytes, the namespace may hold at once. Required when `state` is `present`. |
 | expires_in_seconds<br/>*integer* | How long the claim should cover placements for, in seconds from now. Required when `state` is `present`, and must be positive. A duration rather than a timestamp: the expiry is computed from the cluster's clock, which is the only clock the expiry sweep ever compares against. Every run re-dates the claim to exactly this far in the future, which can shorten a claim as well as extend one. |
+| renew_within_seconds<br/>*integer* | Re-date the claim only when it has less than this long left to run. Optional, must be positive when set, and ignored when `state` is `absent`. See [Idempotence](#idempotence). |
 | state<br/>*string* | The state of the resource. Valid states are `present` or `absent`, defaults to `present`. |
 | auth_namespace<br/>*string* | The namespace to authenticate as, which is normally `system`. |
 
@@ -313,6 +332,20 @@ control node's clock is never compared against the cluster's. A play which
 renews on a schedule and does not want the renewal in its change count should
 set `changed_when` on the task.
 
+Re-dating on every run is the default because it is what a scheduled
+reconcile wants: every run is another chance to re-date before the claim
+lapses, so a weekly play against a thirty day expiry has four of them. A play
+which would rather converge than renew sets `renew_within_seconds`, and the
+module then leaves the claim alone — writing nothing and reporting no
+change — while it is further from its expiry than that window and its limits
+already match. Check mode says the same, so `--check` converges too. The
+remaining life is measured as the server's own `expires_at` against the
+control node's clock, which is the only comparison the module makes that
+spans two clocks and is deliberately a coarse one: skew moves the moment the
+renewal happens by the amount of the skew and decides nothing else, and never
+decides what is reported as changed. Choose a window comfortably larger than
+both the play's cadence and any plausible skew.
+
 ### A lapsed claim
 
 The server refuses to re-date a claim which is no longer active, and says to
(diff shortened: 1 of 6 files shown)


mikalstill:
All eight items addressed in `ee7bbec74`. Item by item, including the one where I did not do what was suggested.

**1 [FIX] — partial connection parameters silently ignored.** Fixed with `required_together=[['api_url', 'auth_namespace', 'key']]`. The diagnosis is right and the `auth_namespace` rename is what makes it bite: a play following the collection README passes `namespace`, `api_url` and `key`, which here sets the claim's *target* and leaves the identity unset, so the explicit connection details were discarded in favour of ambient credentials. The other four modules keep the inherited behaviour — fixing it there is a separate change with its own blast radius.

**5 [FIX] — minimum client version. Done differently, because the suggested fix does not exist.** There is no release to pin. The claim verbs landed on `client-python`'s `develop` as `34aee79` and `git tag --contains` returns nothing; the latest tag `v0.8.3` predates them. That is deliberate — `PLAN-scheduler-reservations` phase 4b decision `D7` records that no release is cut for phases 4b to 7, because no consumer installs one. So instead of inventing a pin:

- `requirements.txt`, the collection README and `docs/user_guide/ansible.md` now say the verbs are unreleased and that a control node needs `pip install git+https://github.com/shakenfist/client-python@develop`, which is how the CI conductor already installs it.
- `_require_claim_verbs()` checks the constructed client and fails naming the missing verbs and the fix, instead of letting an `AttributeError` out. It starts passing on its own once a release carries them.
- `sf_claim` added to the `requirements.txt` comment, which did still list only four modules.

**2 [CONSIDER] — lapsed rows only reaped on the create path.** Now reaped in the re-date path too. Deliberately **not** counted as `changed`: an expired claim holds no cluster capacity, so removing it alters nothing the namespace's coverage depends on. Documented in the `DOCUMENTATION` block.

**3 [CONSIDER] — positional args.** Now keywords. Four mutually type-compatible integers across a repository boundary is exactly where that matters.

**4 [CONSIDER] — every run reports `changed`, check mode never converges.** Added an optional `renew_within_seconds`, **default unset** so existing behaviour and playbooks are unchanged — the 33fl reconcile keeps the "four chances to re-date before a lapse" property `D10` asks for. With it set, a repeat run and a check-mode run both converge, so `006.yml` now carries the idempotence assertion it had to omit.

One correction on the suggestion's parenthetical: it cannot be computed "from the server's `expires_at` so no clock comparison is introduced". Remaining life needs a *now*, and the claim's external view (`namespace_claim.py:355-364`) carries no server-side one. So the comparison is unavoidable and is instead documented as the only two-clock comparison in the module, with its consequence bounded: skew moves *when* a renewal happens by the amount of the skew and decides nothing else. What is reported as `changed` stays a comparison of two of the server's own timestamps.

**6 [CONSIDER] — 503 retry in scenario 006.** Added, following what `cluster_ci_tests/test_namespace_claims.py` actually does.

**7 [CONSIDER] — failed reap returns `meta=None`.** The new claim is now threaded through as `meta`, and the documentation sentence promising a failure "leaves the namespace exactly as it was found" is narrowed, since it was untrue for that path.

**8 [CONSIDER] — 409 gets a generic message.** Now names both the below-usage and not-active cases, in the wording of `CLAIM_REFUSAL_MESSAGE` in `external_api/auth.py`.

**Verification.** Unit tests 25 → 45, all passing; `tox -e flake8` clean on both changed files; the `DOCUMENTATION`/`EXAMPLES`/`RETURN` blocks and `006.yml` re-parsed as YAML. Both `[FIX]` items were mutation-tested — removing `required_together`, and making the capability check a no-op — and each fails exactly one test.

Still not run, unchanged from the original description: the functional cluster CI, so `006.yml` has never executed against a real cluster, and nothing has run against a live Shaken Fist API.

**Merge ordering still applies.** Mach33Labs/33fl#821 was merged ahead of this and broke its daily reconcile with `couldn't resolve module/action 'sf_claim'` — `ansible.cfg` reads a sibling shakenfist checkout, so the module has to be on `develop` first. That has been reverted on 33fl master and will be re-landed once this merges.

Issue #4194, opened 2026-09-13
Agent put answers 500, not 404, for a missing blob


mikalstill:
`POST /instances/<ref>/agent/put` answers an opaque 500 instead of a 404 when the named blob does not exist, because the 404 is written with a method that does not exist.

`shakenfist/external_api/instance.py:1955`:

```python
b = Blob.from_db(blob_uuid)
if not b:
    return self.api_error(404, 'blob not found')
```

`api_error` is not a method on `api_base.Resource`, nor on the endpoint class, nor anywhere else in `shakenfist/external_api/`. Every other rejection in this file uses the module-level helper `sf_api.error(...)`, where `sf_api` is `shakenfist_utilities.api` (imported at `instance.py:17`). Confirmed against the tree:

```
>>> hasattr(api_base.Resource, 'api_error')
False
>>> hasattr(sf_api, 'error')
True
```

So the `if not b:` arm raises `AttributeError` rather than returning. Since phase 5 of `PLAN-api-input-validation` removed the broad `except TypeError` and made an unexpected handler exception a recorded 500, the caller now gets a bare `server error` with an exception record, instead of the `404 blob not found` the endpoint documents in its own `swag_from` responses.

**Who hits this:** any caller who names a blob uuid that does not resolve — a deleted blob, a typo, a uuid from another cluster. It is not limited to a malformed or omitted parameter.

**Why this is filed separately from the validation plan:** phase 6 of `PLAN-api-input-validation` is enforcing required-ness, which closes the *omission* path (`blob_uuid` absent). It does not touch this one, because a well-formed uuid that simply does not exist passes every schema check and still reaches the broken arm.

**How it was found:** the phase 6 required-parameter sweep (`shakenfist/tests/external_api/test_required_sweep.py`), which drives a real authenticated request per declaration through the whole decorator stack. The row for this parameter records `500 / faults` rather than the expected 404.

**Suggested fix:** `return sf_api.error(404, 'blob not found')`. Worth grepping for other `self.api_error` callers at the same time — this one survived because no test covered the not-found path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019M9kQBiC4WxmX6w2NPjrBe

Issue #534, opened 2020-11-12
Validate MAC address format on interface create


mikalstill:
MAC uniqueness is enforced (DB UNIQUE constraint + retry, `network/interface.py:154`), but a caller-supplied MAC is accepted without any *format* validation.

Residual: add a format check. Belongs to `docs/plans/PLAN-generic-allocator.md`.

_Rescoped 2026-06 to the residual work above; the original report is preserved in this issue edit history. See CATEGORIZED_BUGS.md on the `bug-cleanup` branch._

PR #4199, opened 2026-09-13
Enforce required parameters and scalar formats

### LINKED ISSUE #323: Don't let users create virtual networks which overlap with our reserved networks
No one will be happy if you use the same IP Range as the floating network for example...

### LINKED ISSUE #3269: The API should enforce that user data is base64 encoded in a more friendly manner
At the moment, the instance simply errors out, and the hypervisor logs something like this:

```
2026-06-13T17:50:30.769490+10:00 sf-2 sf-queues[2490]: Exception in thread 1781337029.3543353-CwAlRqsSEjI1w9sW:
2026-06-13T17:50:30.769534+10:00 sf-2 sf-queues[2490]: Traceback (most recent call last):
2026-06-13T17:50:30.769558+10:00 sf-2 sf-queues[2490]:   File "/srv/shakenfist/venv/lib/python3.11/site-packages/shakenfist/operations/node_inst_netdesc_op.py", line 118,
 in dispatch_task
2026-06-13T17:50:30.778138+10:00 sf-2 sf-queues[2490]:     self.__getattribute__(f'_{task.name}')(inst)
2026-06-13T17:50:30.778189+10:00 sf-2 sf-queues[2490]:   File "/srv/shakenfist/venv/lib/python3.11/site-packages/shakenfist/operations/node_inst_netdesc_op.py", line 261,
 in _instance_start
2026-06-13T17:50:30.778264+10:00 sf-2 sf-queues[2490]:     inst.create(iface_uuids)
2026-06-13T17:50:30.778301+10:00 sf-2 sf-queues[2490]:   File "/srv/shakenfist/venv/lib/python3.11/site-packages/shakenfist/instance.py", line 814, in create
2026-06-13T17:50:30.778398+10:00 sf-2 sf-queues[2490]:     self.power_on()
2026-06-13T17:50:30.778419+10:00 sf-2 sf-queues[2490]:   File "/srv/shakenfist/venv/lib/python3.11/site-packages/shakenfist/instance.py", line 1387, in power_on
2026-06-13T17:50:30.778686+10:00 sf-2 sf-queues[2490]:     self._make_config_drive_openstack_disk(
2026-06-13T17:50:30.778792+10:00 sf-2 sf-queues[2490]:   File "/srv/shakenfist/venv/lib/python3.11/site-packages/shakenfist/instance.py", line 1136, in _make_config_drive
_openstack_disk
2026-06-13T17:50:30.779034+10:00 sf-2 sf-queues[2490]:     user_data = base64.b64decode(self.user_data)
2026-06-13T17:50:30.779122+10:00 sf-2 sf-queues[2490]:                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-06-13T17:50:30.779138+10:00 sf-2 sf-queues[2490]:   File "/usr/lib/python3.11/base64.py", line 88, in b64decode
2026-06-13T17:50:30.779183+10:00 sf-2 sf-queues[2490]:     return binascii.a2b_base64(s, strict_mode=validate)
2026-06-13T17:50:30.779209+10:00 sf-2 sf-queues[2490]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-06-13T17:50:30.779235+10:00 sf-2 sf-queues[2490]: binascii.Error: Incorrect padding
```

### LINKED ISSUE #4195: A non-coercible agent put mode escapes as a 500
`shakenfist/external_api/instance.py:1945`, in `InstanceAgentPutEndpoint.post`:

```python
try:
    int(mode)
except ValueError:
    try:
        symbolicmode.symbolic_to_numeric_permissions(mode)
    except ValueError as e:
        return sf_api.error(406, 'invalid mode: %s' % e)
```

The intent is "accept either a numeric mode or a symbolic one, and answer 406 otherwise". The guard only catches `ValueError`, but `int()` raises `TypeError` for an argument that is not a string or a number — `int(None)`, `int([])`, `int({})`. A `TypeError` therefore escapes the handler entirely and, since phase 5 of `PLAN-api-input-validation` removed the broad `except TypeError` from `handle_authorization_exceptions`, becomes a recorded 500 with a bare `server error` body rather than the documented 406.

**Reachability depends on `API_VALIDATION_MODE`:**

* Under `enforce` (the default since phase 4), the compiled schema rejects a non-string `mode` before the handler runs, and phase 6 will reject an omitted one. So the default path is covered.
* Under `warn` or `off` — the operator's rollback — nothing types the parameter, the raw JSON value reaches the handler, and any non-coercible value lands here as a 500.

That rollback case is exactly what decision D34 of the phase 6 plan is about: a handler guard is not dead just because enforcement makes it unreachable, because `warn` and `off` hand the handler whatever the caller sent.

**How it was found:** the phase 6 required-parameter sweep (`shakenfist/tests/external_api/test_required_sweep.py`) records `500 / faults` for an omitted `mode` on this route. The omission path specifically is closed by phase 6 enforcing required-ness; this issue is the residue that enforcement does not cover.

Worth noting because the phase 6 plan quotes this parameter as its canonical example of "declared required while omitting it has always been accepted" (the `CompiledEndpoint` docstring). That was true only while the broad `except TypeError` was absorbing it. It is no longer accepted — it is a recorded 500.

**Suggested fix:** catch `(TypeError, ValueError)` on the `int()` call, and treat a non-string, non-integer mode as a 406 like any other invalid mode.

mikalstill:
Phase 6 of [`PLAN-api-input-validation.md`](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-api-input-validation.md), implementing the [merged phase plan](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-api-input-validation-phase-06-required.md) as amended by #4189.

Since phase 3 a `missing-required` finding has been generated, logged, and then filtered out of the enforceable set, because several parameters were declared required while omitting them had always worked. This phase finds out which of those declarations were telling the truth, corrects the ones that were not, and then enforces them. It also gives the semantic type tokens something to do beyond documenting themselves.

Fixes #3269
Fixes #323
Fixes #4195

## The premise turned out to be overstated

Finding F1 of the plan says all 75 body and query declarations carrying `required=True` have a handler default, so nothing is structurally required and enforcement is a contract change for all of them. That is true as far as it goes, and it reads as though 75 endpoints are about to start behaving differently.

Step 1 asked every handler what an omission actually does, by driving one real authenticated request per declaration through the whole decorator stack. The answer:

| Verdict | Count | Meaning |
|---|---|---|
| `guarded` | 63 | the handler already answers 4xx |
| `faults` | 7 | 5xx, or an exception is recorded |
| `accepted` | 6 | 2xx |

So for 63 of 76 this is a **message** change, not a behaviour change. The full table is in the plan file, pinned by `test_required_sweep.py` so a handler that starts answering differently fails CI rather than silently invalidating the evidence.

## Exactly one declaration was wrong

`shared` on `POST /artifacts` — the signature is `shared=False`, a real default, and the declaration said required. It is now `required=False`, which is the only published-API relaxation in this branch.

The other four `accepted` rows **keep** `required=True`, and this is the one judgement call worth reviewing. Each of them answers 2xx *and writes something*: an `execute` agent operation carrying a null commandline, a `get-file` with a null path, a `put-blob` that resolves a real blob and then queues it for delivery to nowhere, a DNS record pointing at `None` plus a dnsmasq rewrite. D35's rule is "required only where the handler refuses today", but its own wording carries a qualifier — "and does something sensible" — which the step brief had dropped. A 200 that commits an operation which cannot succeed is not an acceptance worth preserving, and declaring it optional would publish the breakage as the contract.

Definition-of-done item 3 stated the mechanical rule without that qualifier and so contradicted D35. It is amended in this branch with the original wording left visible.

## Three places the plan was wrong

Each was followed as far as it was right and departed from where it was not, with the reasoning recorded in the plan file.

1. **`base64.b64decode(..., validate=True)`**, which the plan names explicitly, refuses base64 wrapped at 76 columns. That is what `base64(1)` emits, so `sf-client instance create -U "$(base64 user-data.yaml)"` would have begun failing. The validator uses the default arguments — the same call the hypervisor makes at `instance.py:1930` — and still refuses the #3269 input. Nothing in CI sends wrapped base64, so this would have shipped silently.
2. **`ArtifactsEndpoint.post.url` is frequently not a URL.** `cirros` and `ubuntu:20.04` are distro shortcuts expanded by `images._resolve_image()`, and `artifact cache cirros` is a line in the functional suite. The `url` validator therefore cannot require a scheme; it is deliberately thin and its docstring says so rather than dressing it up. This is the failure mode the plan's Risks section predicted, and it was live.
3. **Step 5 guards only the floating network.** #323's title says "reserved networks" plural; its body names the floating network. Node egress and mesh addressing are not checked.

## What callers see

All of it is reverted by setting `API_VALIDATION_MODE` to `warn` or `off`.

**Better:**

* seven former 500s become 400s naming the parameter;
* `POST /networks` with no `netblock` stops answering `cannot parse netblock: None does not appear to be an IPv4 or IPv6 network`, which told the caller the server had seen a `None`;
* non-base64 `user_data` is refused at the API instead of failing on the hypervisor after the instance is scheduled and placed.

**Worse, on the pure-omission path only:** `confirm` on the three `DELETE` routes, `disk` on `POST /instances`, and `network` on interface create lose their specific messages. In each case the informative message survives for the case that carries information — `confirm: false`, `disk: []` and a malformed networkspec all still reach the handler.

**Status codes that change:** `POST`/`DELETE /networks/<n>/dns` with no `name` move from 406 to 400. A non-uuid `blob_uuid`, `upload_uuid` or `target_uuid` moves from 404 to 400, because the `uuid` validator now runs before the handler's `from_db()` lookup.

## Verification

The sweep now runs under `warn` rather than `enforce`. Its purpose is to record what the *handler* answers for an omission, and after this branch the only mode in which a handler still sees one is the rollback — which also makes it the proof of the warn-mode criterion across all 76 rows rather than the two the plan asked for.

The enforcement tests were mutation-tested by restoring the deleted filter: all four fail, and the sweep stays green. `user_data` is pinned by asserting **`Instance.new` was never called**, not by asserting the decode fails — the decode runs in a different daemon on a different machine, so asserting on the creation call is what makes the claim true of the deployed system.

4850 unit tests pass and `pre-commit run --all-files` is clean. A cluster CI case covers the netblock overlap.

## Filed, not fixed

Step 1 found two bugs that enforcement does not close, both deliberately left out of scope:

* **#4194** — `POST /instances/<ref>/agent/put` answers an opaque 500 rather than 404 for a missing blob, because the 404 is written `self.api_error(...)` and no such method exists. Reachable by any unresolvable blob uuid, not only an omitted parameter.
* **#4195** — `int(mode)` catches only `ValueError`, so a non-coercible mode escapes as a 500 under `warn`/`off`.

`#534` was closed outside this plan by #4183 before the phase started, which is recorded in the plan's Amendments section.


sfconductor:

The conductor will merge this at `21ae6f4eaaf5` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `21ae6f4eaaf5`, bypassing the merge queue.

Issue #4197, opened 2026-09-13
Re-derive the load budget for the four metrics-publish pairs exempted by transient-capacity-refusals phase 3


mikalstill:
Phase 3 of [PLAN-transient-capacity-refusals](https://github.com/shakenfist/shakenfist/blob/develop/docs/plans/PLAN-transient-capacity-refusals.md) makes `sf-resources` republish its `node_metrics` row within about five seconds of the set of running libvirt domains changing, instead of only once every 60 seconds. That closes a real gap -- the scheduler charges a node `max(measured, committed)` CPUs, the committed side is released inside the delete transaction, and the measured side used to lag by up to a minute, so a node kept refusing work for that long after it was emptied.

The cost is that the rate of every database call on the metrics publish path is now set by instance churn rather than by a fixed tick. Four `(operation, caller_daemon)` pairs are therefore marked `activity_coupled: true` in `shakenfist/data/database_load_budget.yaml`:

| Pair | Calls per publish | Status |
|------|-------------------|--------|
| `GetNodeByFqdn` / `resources` | 1 | existing entry, marked |
| `GetQueueLength` / `resources` | 12, or 17 on the network node | existing entry, marked |
| `GetNodeMetrics` / `resources` | 1 | new entry |
| `UpsertNodeMetrics` / `resources` | 1 | new entry |

`enforced()` in `shakenfist/deploy/shakenfist_ci/load_budget.py` returns false for an activity-coupled entry, so **these four pairs are no longer checked against the budget**. Two of them were enforced before this change. That is the point of this issue: the exemption is deliberate and it should not be permanent.

The two new entries carry `per_node_base_qps: 0.017`, which is derived rather than fitted -- each runs exactly once per publish and the floor on publishes is one per node per 60 s. The schema requires at least one rate term (`BudgetEntry._has_a_term()`), so the field could not be left blank; the notes say the figure is the one-per-publish floor and not a measurement.

## What would close this

Run `tools/derive-database-load-budget.py` over a measurement window on `sfcbr` that includes the new cadence, so the four pairs carry activity-coupled rates fitted against instance churn instead of a hand-derived floor. The `activity_coupled` marks survive re-derivation (`tools/derive-database-load-budget.py`, the carry-forward in `emit()`), so this is an improvement to the numbers rather than a repair.

## Worst case, for whoever picks this up

[...]

PR #4200, opened 2026-09-13
Publish node metrics when the running-domain set changes

mikalstill:
Phase 3 of [PLAN-transient-capacity-refusals](docs/plans/PLAN-transient-capacity-refusals.md). Plan file: [`PLAN-transient-capacity-refusals-phase-03-metrics-on-change.md`](docs/plans/PLAN-transient-capacity-refusals-phase-03-metrics-on-change.md).

## The problem

The CPU pre-filter charges a node `max(measured, committed)`:

```python
current_cpu = max(measured_cpus, committed_cpus)   # scheduler.py:321
```

The committed side is released inside the delete transaction. The measured side is `cpu_total_instance_vcpus` out of the node's `node_metrics` row, which `sf-resources` republished only once every 60 s. So a node which had just been emptied went on refusing work for up to a minute.

The master plan's own evidence is run 34119030297: `measured 6 / committed 0 / limit 3`, thirty-two seconds after six instances were deleted and their ledger correctly released.

The `max()` is right and is untouched — its docstring explains it is what stops a node whose ledger is full but whose instances are still fetching images from winning the load ordering and then being refused by the guard. Only the cadence of one of its two inputs changes.

## The change

`sf-resources` now polls the active libvirt domain-id set every 5 s (one `listDomainsID()`, no database access) and publishes immediately when that set differs. The 60 s publish stays as a floor, and any publish resets its clock, so a change-triggered publish suppresses a tick moments later rather than being followed by a near-duplicate.

The trigger is set **membership**, not a count and not a vCPU total. A count misses a poll interval containing one delete and one create — routine during a teardown-and-recreate, and precisely where a stale low measurement lets the scheduler over-admit. A vCPU total adds nothing: there is no CPU hotplug path.

Because a publish rewrites the whole row, this also refreshes `memory_available`, the KSM ratio input, `disk_free_instances` and the disk-busy rate — three further pre-filter stages read the same row, and two of them have no ledger to fall back on.

## Commits

| | |
|---|---|
| `9ffb387b8` | `get_active_domain_ids()` — one `listDomainsID()`, no per-domain lookup |
| `20b9929f7` | the cadence: two pure gate functions plus the loop wiring |
| `ac7266184` | functional test that measurement follows a delete |
| `68faec1f7` | load budget, derive-tool fix, regenerated Prometheus rules |
| `ed6af3fd6` | documentation |

## What a reviewer should look at hardest

**The publish is not cheap, and the plan originally said it was.** A publish is fifteen database round trips (twenty on the network node), twelve of them `get_work_queue_length()` reads that have nothing to do with capacity. Worst case at full churn is 3/s per node, about 17/s on the six-node reference cluster.

Decision D20 rejects the obvious optimisation — skipping the queue reads on a change-triggered publish, a five-fold saving — because `node_queue_waiting` is read by `_has_reasonable_queue_state()` and by `summarize_resources()`'s skip condition, so a row carrying two different ages is not inert. That decision is written down with the arithmetic so it can be reversed on measurement rather than on opinion. **If you disagree with anything here, this is the thing to disagree with.**

**Four budget pairs lose CI enforcement.** `activity_coupled` switches `enforced()` off, and two of the four were enforced before. [#4197](https://github.com/shakenfist/shakenfist/issues/4197) tracks re-deriving them over a window that includes the new cadence. `GetObjectState`/`resources` is deliberately **not** marked — it turned out not to be on the publish path at all (it comes from the health-check thread), and its budget note claimed otherwise and is corrected here.

**This does nothing for a cold cluster.** A domain set that is empty and stays empty never changes, so the first publish after start-up still waits out the 60 s floor. The two documentation pages that discuss the warm-up window say so explicitly rather than implying the window closed.

## Testing

- 16 new unit tests. Two mutation checks were run and reported: comparing set lengths instead of membership fails exactly one test, and reverting the wiring to the bare 60 s gate fails the loop test.
- Four of the tests drive `_run_inner()` itself with libvirt and the database mocked, so the wiring is covered and not just the decision.
- One functional test asserting `cpu_measured` falls by an instance's vCPUs within 20 s of a delete. It waits for the domain to actually boot before taking a baseline, and asserts a relative drop so a sibling instance landing on the node does not break it. **This one needs a real CI run to confirm the timing holds** — that reading is step 3f's closeout.
- Full unit suite (4,837 tests) and `pre-commit run --all-files` pass on the rebased branch.

## Incidental fixes the phase forced

- `emit()` in `tools/derive-database-load-budget.py` wrote `entry['measured']` unconditionally, so a schema-legal entry with no measurement raised `KeyError` on the next re-derivation. A budget file that cannot be regenerated is not a budget file.
- `examples/prometheus-database-load-rules.yaml` is regenerated, since a test asserts it is byte-for-byte what the budget renders to.

https://claude.ai/code/session_01Czfe1Jt9Tbh7cwhWBgMV9s


mikalstill:
All five addressed in `c1c4f7f26`. All five were real; two were defects rather than polish.

**1 — Functional test races its own baseline.** Correct, and worse than described: the baseline was read after `_await_instance_create()`, which returns as soon as the instance reaches `created` (set right after `power_on()`), while the figure has to travel through a 5 s domain poll, a publish *and* `SCHEDULER_CACHE_TIMEOUT`. A baseline taken early is the node's idle measurement, and `baseline - cpus` is then a value the node can never return to — so the test would have failed at its deadline for the opposite of the reason it exists.

Fixed as suggested: read the node's measurement *before* creating anything, wait for the rise to include the started instance, and only then start the drop clock. The two waits share a `_await_cpu_measured()` helper that takes the expectation as text, so publish-on-start and publish-on-delete fail with different messages.

**2 — The two new budget entries would be silently deleted by the next re-derivation.** Correct, and this is the one that mattered. My D26 fix stopped a `KeyError` and I mistook that for the disease; not raising is not the same as surviving. `main()` builds `entries` only from pairs clearing the cut, `emit()` iterates `entries`, and *a pair with no measurement is exactly a pair no measurement saw*. The marks, notes and #4197 would have gone with them.

I took the "no `measured:` block" route rather than a new `hand_added` field: `to_entry()` always sets a measurement, so its absence already means "written by hand for a call the fit could not see", and it needs nothing new to remember. A pair which **was** fitted and has fallen below the cut is still dropped — it has a measurement saying it used to matter and no longer does, which is what the cut is for.

Two `EmitRoundTripTestCase` cases added, one for each half. Deleting the carry-forward fails exactly one of them; I checked.

**3 — `emit()` carries a measurement forward, contradicting its own comment.** Right, and the reasoning is right: every other carried value is judgement that stays true across a re-measurement, but a measurement is a fact about one window and the emitted `_doc.window` names a different one. Fallback dropped.

Writing the round-trip test then surfaced a fourth thing in the same function, which I've fixed here too: `wrapped_block()` used `textwrap.fill` defaults, so a wrap inside `one-per-publish` came back as `one-per- publish` — a folded scalar turns the newline into a space, and the text degrades a little more on every re-derivation. Now breaks on neither hyphens nor long words.

**5 — Test comment misstates what the threshold protects against.** Correct, and I had repeated the error in my own PR description. Subtracting the instance's vCPUs tolerates *pre-existing* load; it does nothing about a sibling arriving mid-window. Now picks the emptiest hypervisor as `test_cluster_resources_charges_unbooted_placements` does, and the comment says what is true, including that a mid-window sibling remains the known flake source.

**4 — Stale figures.** Fixed in the master plan paragraph and in the phase plan's Risks and Back brief. The only surviving mention of ten round trips / seven queue reads is the note recording what the first draft got wrong, which is deliberate.

While in that file I also corrected the neighbouring `test_cluster_resources_charges_unbooted_placements`, whose comment stated the 60 s cadence as current — this PR makes that stale, and its point survives the change anyway (a placed-but-unbooted instance stays invisible to `cpu_measured` however promptly the row is republished).

`pre-commit run --all-files` and the full unit suite pass.

sfconductor:

The conductor will merge this at `ca5d8da60b5e` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `ca5d8da60b5e`, bypassing the merge queue.

Issue #4201, opened 2026-09-13
Reservation clamps engage silently, hiding a node that reserves all of its CPU


mikalstill:
`_compute_reservations()` (`shakenfist/daemons/resources/main.py:102-129`) has two safety clamps, and both are silent:

```python
'cpu_schedulable': max(1, cpu_threads - cpu_reservation_threads),
...
memory_reserved_mb = min(memory_reserved_mb, memory_total_mb // 2)
```

Both are correct. Neither says anything when it engages. A node whose thread reservation meets or exceeds its thread count publishes `cpu_schedulable` 1 and is indistinguishable — in metrics, in events, and in `/admin/resources` — from a node that genuinely has one spare thread. The same goes for a RAM reservation above half the machine.

## This is reachable in practice, and it is reached in our own CI

Shaken Fist's CI under-cloud runs 4 vCPU / 12 GB nodes, and `examples/_shared/site.yml:352-363` bumps network and database nodes to 4 reserved threads and `max(2.0, 10% of RAM) + 4.0` GB:

```yaml
node_cpu_reservation_threads: >-
  {{ (1 + ((node_is_network_node or node_is_database_node) | ternary(1, 0))) * 2 }}
```

So every CI infra hypervisor reserves **100% of its CPU** and **51% of its RAM**, and is rescued by the clamps. Measured from the rendered `/etc/sf/config` and published metrics of every node in run 34681505274:

| Topology | Node | Roles | Reserved threads | `cpu_schedulable` | `limit_cpus` |
|---|---|---|---|---|---|
| slim-tier | primary | hypervisor, network, database | 4 | 1 | 3 |
| slim-tier | sf1 | hypervisor, database | 4 | 1 | 3 |
| slim-tier | sf2 | hypervisor | 2 | 2 | 6 |
| slim-primary | sf1 | hypervisor, network | 4 | 1 | 3 |
| slim-primary | sf2–sf5 | hypervisor | 2 | 2 | 6 |

The RAM side saturates too: all three slim-tier nodes publish `memory_reserved_mb` 5980, which is `11960 // 2`, not the 6144 the playbook asked for.

Identically sized guests, and a scheduling ledger of 3 against 6, with nothing anywhere saying why. Two of slim-tier's three hypervisors sat at 100% of their ledger for an entire run as a result.

## Why it is worth an event

[...]

PR #4202, opened 2026-09-13
Plan and gate phase 4 of CI cloud sizing

mikalstill:
Plans phase 4 of `PLAN-ci-cloud-sizing`, runs the three steps that do not need a merge run, and corrects two comments that misled the planning survey.

Three commits, reviewable in order. The only code change is in the third, and it is comments.

## What this is for

Phase 4 re-shapes the nested CI clouds against the phase 2 peak-demand data. `slim-tier` has run at a 19% pass rate with two of its three hypervisors pinned at 100% of their admission ledger for entire runs.

## The survey got something wrong, and the correction is the interesting part

The plan's F2 claimed no 4-thread CPU reservation existed anywhere. **It does.** The per-host reservation is computed by the deployment playbook CI actually runs:

```yaml
# examples/_shared/site.yml:359-363
node_cpu_reservation_threads: >-
  {{ (1 + ((node_is_network_node or node_is_database_node) | ternary(1, 0))) * 2 }}
```

So a network or database node reserves **4** threads and a plain hypervisor 2. On a 4 vCPU guest that leaves `max(1, 4-4)` = 1 schedulable thread and a ledger of `floor(1 x 3.0)` = 3, against 2 and 6 on a plain hypervisor. Proven from the rendered `/etc/sf/config` of every node in the phase 3 merge run, not derived:

| Topology | Node | Roles | Reserved | `cpu_schedulable` | Ledger |
|---|---|---|---|---|---|
| slim-tier | primary | hyp, net, db | 4 | 1 | 3 |
| slim-tier | sf1 | hyp, db | 4 | 1 | 3 |
| slim-tier | sf2 | hyp | 2 | 2 | 6 |
| slim-primary | sf1 | hyp, net | 4 | 1 | 3 |
| slim-primary | sf2–sf5 | hyp | 2 | 2 | 6 |

The survey missed it because the value appears in neither the collection nor the CI repository: the role default of 2 is only the fallback for a caller that does not set the variable, and this caller always sets it. **The master plan's original explanation was right and the survey's contradiction of it was wrong**, which the plan now records rather than quietly deletes — F2 is struck through and kept, because it is the reason the phase has a gate at all, and D1 says plainly that the gate rested on a false premise and is not entitled to claim vindication.

Two comments actively encouraged the wrong conclusion, and the third commit fixes them:

- `scheduler.py:_schedulable_threads()` said "there is no longer an infra-role bump" — true of the synthetic fallback it annotates, false as a statement about the system.
- `config.py`'s RAM reservation described the templating as "folding in any historical infra-role bump", where *historical* reads as *since removed*.

## What the steps produced

**4a (gate).** The mechanism above, verified. It also found that the RAM reservation saturates its clamp on the same nodes — all three tier nodes publish `memory_reserved_mb` 5980, which is `11960 // 2`, not the 6144 the playbook asked for.

**4b.** F2, the master plan's phase 4 section and the `index.md` row corrected to cite the line that sets the value. Four smaller survey corrections also landed at source: only one `-released` topology exists and it is not structurally a copy of the other; `functional-tests.yml` is in this repository and holds no node lists; the three-node structural minimum is enforced by two tests counting all nodes, and `slim-tier` already sits at `test_network_lifecycle`'s two-hypervisor minimum; and the candidate table's `vCPU` column is under-cloud spend, not scheduling capacity.

**4c.** The reshape prepared as a reviewable diff in the plan — three lines, `cpu: 4` → `cpu: 6` in `ci-topology-slim-tier.yml` — with predictions derived from 4a's arithmetic rather than guessed:

| Node | Reserved | `cpu_schedulable` | Ledger | Demand bound |
|---|---|---|---|---|
| primary | 4 | 1 → 2 | 3 → 6 | 0.75 → 1.50 |
| sf-1 | 4 | 1 → 2 | 3 → 6 | 0.75 → 1.50 |
| sf-2 | 2 | 2 → 4 | 6 → 12 | 1.50 → 3.00 |

Cluster ledger 12 → 24 for +6 under-cloud vCPU and no extra RAM. The "before" column reproduces the measured 3/3/6 exactly, which is the model checking itself. It carries a three-way falsification statement separating "the model is wrong" from "the model is right and the demand bound binds" from "the binding dimension moved", because those point at different next actions.

## Decisions a reviewer should push on

**D3 keeps `slim-tier` at 6 vCPU, and the case for 8 is real.** Six only lifts the infra nodes to where plain hypervisors already sit; eight would give 12 / 12 / 18 and triple rather than double the demand bound on the nodes that refuse. Six wins on cost (+6 vCPU against +12 on the binding resource) and on the measured p90 of 10 committed vCPU against a ledger of 12, with the revision condition written down: if the merge runs show demand refusals persisting at 6, the answer is 8, decided on evidence.

A cheaper third option — setting `node_cpu_reservation_threads` per host in CI inventory, which the playbook explicitly supports — is recorded as **rejected**, because it departs from the production reservation values the master plan deliberately keeps in CI.

## Filed, not fixed

**#4201** — both reservation clamps engage silently, so a node reserving 100% of its CPU is indistinguishable from one with a spare thread. The body states explicitly that the clamps must not be removed, the CI reservation must not be lowered to dodge them, and the scheduler must not be taught to tolerate a zero ledger: the defect is that they are quiet. Out of this phase's scope.

## What is still owed

4d–4g are blocked on applying `Prepared changes` to `shakenfist/actions` — a separate repository — and on the merge runs after it. Note that `functional-tests.yml:485` references `smoke-cluster.yml@main` unpinned, so the new shape takes effect on the next run including in-flight ones; apply it when the queue is quiet.

Two deviations from the step table are declared in the plan's Outcome rather than left to be discovered: 4b ran inline instead of as a sub-agent (its content changed once 4a landed, and it touched files the session already held), and 4a–4c landed as one commit instead of three (their edits interleave in the same regions).

## Verification

`pre-commit run --all-files` passes all ten hooks on each commit. `tools/check-plan-status.py` agrees. The Definition of done items that are checkable without a merge run were run rather than asserted.


sfconductor:
Merged at `a5f1e29457ea`, bypassing the merge queue.

PR #4231, opened 2026-09-16
Adopt PyJWT 2.14.

mikalstill:
Renovate's bump to PyJWT 2.14.0 (#4218) failed CI with every federated exchange
answering 401. Three separate changes in 2.14 were involved, so this branch takes
the bump and everything it breaks together rather than leaving the version change
to Renovate — `cooldown_duration` does not exist in 2.13, so there is no spelling
of the fix that works under both. **#4218 becomes redundant once this lands.**

### 1. The JWKS mock silently stopped intercepting

`PyJWKClient` no longer fetches through `urllib.request.urlopen`; it builds an
opener so it can refuse redirects. Three test modules patched `urlopen` to serve
locally generated keys, and all three fell through to the real network: they made
HTTPS requests to `token.actions.githubusercontent.com`, got GitHub's genuine key
set, did not find the test's `key-1`, and failed as though the exchange endpoint
were refusing valid tokens.

The patch now goes on `urllib.request.OpenerDirector.open`, which both spellings
end up calling, and it lives in one place (`shakenfist/tests/fake_jwks.py`) so the
next rearrangement upstream is one edit rather than three. It stays a transport
level fake deliberately: patching `PyJWKClient.fetch_data` would be simpler and
would disable the key set caching several of these tests exist to check.

### 2. A cooldown under the unknown-kid refetch

2.14 will not force a JWKS refetch more than once per `cooldown_duration`
(default 30s). That is a real change to what `/auth/federated` does, and worth
having: the endpoint takes no credential, so before it every invented key id a
stranger sent bought another JWKS fetch of the issuer, made by us, under that
issuer's lock.

Taken deliberately rather than inherited. New
`FEDERATION_JWKS_ROTATION_COOLDOWN_SECONDS`, default 30, passed explicitly, with
the cost it buys written down in the operator guide: a rotation landing inside the
window is not recognised until the window elapses.

### 3. Stricter base64url decoding

The 48 spellings of an RS256 signature that used to verify are now one.
`test_re_encoding_the_signature_does_not_change_the_identity` was written as
"spellings PyJWT will verify", which 2.14 takes to one — it would have quietly
become a test of nothing. It now asserts the property over the spellings
themselves, since what we need does not depend on how strict the decoder of the
day is.

### Verification

- Full unit suite under 2.14: 4813 passed, 0 failed.
- `pre-commit run --all-files` clean.
- Both new assertions mutation tested: hardcoding the cooldown to zero fails all
  three cooldown tests, and restoring the old `urlopen` patch target fails all six
  caching tests.


mikalstill:

Round 1 (`0c69722b2`) addressed all eight actionable items: 4 `fix`, 2
`document`, 2 `consider`. Every claim the review made about 2.14 was
checked against the installed wheel, and every assertion it called
non-discriminating was mutation tested rather than read.

**`fix`**

1. **JWKSCache docstring** — confirmed 2.14's `get_signing_key` holds a
   per-client `RLock` across the refetch, so the "PyJWKClient does not
   collapse concurrent refetches" paragraph was stale. Rewritten to say
   what the per-issuer lock still uniquely buys: cross-issuer isolation.
2. **Stampede test** — confirmed it passes with `JWKSCache`'s lock
   removed entirely, so the "an unlocked implementation genuinely
   stampedes" claim was false. Claim dropped, and
   `test_a_slow_issuer_does_not_delay_a_healthy_one` added for the
   property only our code provides. Mutation tested: collapsing the
   per-issuer locks onto one shared lock fails it.
3. **`test_rate_limiting_precedes_the_jwks_fetch`** — measured directly:
   at cooldown 30 an unknown kid after a successful exchange costs 0
   fetches, at 0 it costs 1. `FederatedExchangeTestCase.setUp` now turns
   the cooldown off, mirroring `FederationTestCase`.
4. **Config bounds** — `ge=0` on the cooldown and `ge=1` on
   `FEDERATION_JWKS_CACHE_SECONDS`. Both constructors run per exchange,
   so a mistyped value was a 500 per request; now a `ValidationError` at
   config load.

**`document`** — cooldown reworded as measured from *any* successful
fetch in both the config description and the operator table, with a
paragraph on cache-expiry refetches starting the window; new "The JWKS
URI must be the final URL" section plus an upgrade note for the
no-redirects change.

**`consider`** — both taken. The vacuous `assertGreater` replaced with
the value assertion, and `FakeJWKSTransportTestCase` added: two tests
that fail loudly, naming `fake_jwks.py`, when PyJWT moves its fetch
again. Mutation tested against the old `urlopen` target.

**One disagreement.** Item 3 was slightly overstated. I mutated the
endpoint to move the rate limit below `validate_token`, and the test
failed at *both* cooldown values — its status-code assertion still
caught the reordering. It was the fetch-count assertion, the half the
test is named for, that had gone inert. The fix stands.

Full unit suite 4816 passed / 0 failed (was 4813; three new tests),
`pre-commit run --all-files` clean.

sfconductor:

The conductor will merge this at `8e82e04ebf8c` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `8e82e04ebf8c`, bypassing the merge queue.

Issue #4223, opened 2026-09-15
A null object reference resolves to an arbitrary object


mikalstill:
## Summary

`from_db_by_ref(None, namespace)` does not answer "not found". It
resolves to an arbitrary active object in the namespace, as though the
caller had named it.

The three pushed-down implementations —
`Network.from_db_by_ref` (`shakenfist/network/network.py:189`),
`Instance.from_db_by_ref` (`shakenfist/instance.py:452`) and
`Artifact.from_db_by_ref` (`shakenfist/artifact.py:243`) — all build

```python
criteria = ObjectFilterCriteria(
    states=list(cls.ACTIVE_STATES),
    namespace=criteria_namespace,
    name=object_ref,
)
```

and the criteria object reads a `None` name as *no name filter* rather
than as a name of `None`. So the query returns every active object of
that type in the namespace. One match is returned as the answer; two or
more raise `MultipleObjects`, which surfaces as
`400 multiple networks have the name "None" in namespace "..."`.

## How a caller reaches it

Every `from_db_by_ref` call site in `shakenfist/external_api/` is fed a
route segment, which can never be `None` — except one.
`_netdesc_safety_checks` (`shakenfist/external_api/instance.py:358`)
passes `netdesc['network_uuid']`, which comes from the request body. It
checks that the key is *present*, not that it has a value:

```python
if 'network_uuid' not in netdesc:
    return sf_api.error(400, 'network specification is missing network_uuid')
```

so `{"network_uuid": null}` walks straight through.

Both netdesc call sites are affected: instance create
(`instance.py:829`) and interface hotplug (`instance.py:1153`).

## Reproduction

Against a namespace holding exactly one network, with an instance that
has an interface added after creation:

```
POST /instances/<instance_ref>/interfaces
{"network": {"network_uuid": null}}
```

answers **200** and creates a network interface attached to whichever
network the namespace happens to hold. Measured through the whole
decorator stack with the fixture from
`shakenfist/tests/external_api/test_required_sweep.py`, at the default
`API_VALIDATION_MODE=enforce`.

Directly:

```python
>>> Network.from_db_by_ref(None, 'system')
<Network 12345678-1234-4321-8234-000000000002 name='sweepfixturenet'>
```

`''` correctly answers `None`. `False` and `0` raise a pydantic
`ValidationError` from `ObjectFilterCriteria`, which is a 500.

## Why this is a regression rather than an old bug

The generic `baseobject.from_db_by_ref`
[...]

Issue #4167, opened 2026-09-10
Declared-but-nullable parameters turn caller mistakes into recorded 500s


mikalstill:
Every declared API parameter compiles to a marshmallow field with `allow_none=True`, and `required` is deliberately not enforced (decision D17 of `PLAN-api-input-validation`). So on the **default** `enforce` path, a caller can send `{}` or an explicit `null` for any declared parameter and the handler is entered with `None`.

Any handler that then dereferences that parameter without a type guard — `in`, `len()`, `.startswith()`, `.lower()`, subscripting, iteration — turns an ordinary caller mistake into a 500: an exception record under `/srv/shakenfist/exceptions/`, an ERROR `Server error` log line, and (since #4162) a response body of `server error` that tells the caller nothing.

Until #4162 the `except TypeError` arm in `handle_authorization_exceptions` absorbed the `TypeError` subset of this and answered 400 — badly, in the interpreter's own words, which is why it was deleted. Deleting it made the class visible rather than creating it.

## The instance already fixed

`InstancesEndpoint.post` defaulted `name=None`, and `validators.hostname(None, ...)` *returns* a falsy `ValidationError` rather than raising, so execution reached `contains_domain = '.' in name`:

```
TypeError: argument of type 'NoneType' is not iterable
```

`POST /instances {}` and `POST /instances {"name": null}`. Fixed in #4162 with a guard and tests, after the automated review found it.

## What is still open

Three more in the same handler, all verified, none of them regressions from #4162 — they raise `ValidationError` or `AttributeError`, which the deleted arm never caught, so they behaved identically before it:

1. **`cpus`, `memory`, `uefi`, `secure_boot` as `null` or omitted** → pydantic `ValidationError` from `InstanceData(...)` inside `Instance.new` → `_db_create`. Fixing this means the handler enforcing required-ness for parameters the schema declares required, which is exactly the D17 question this phase deferred. The handler already does that for `disk` (`instance must specify at least one disk`), so there is precedent either way.
2. **`disk[].base` set to a truthy non-string** (`5`, `true`) → `AttributeError: 'int' object has no attribute 'lower'` in `shakenfist/util/general.py:noneish`.
3. **`network[].network_uuid` set to a non-string** → `AttributeError: 'int' object has no attribute 'replace'` in `shakenfist/util/general.py:valid_uuid4`, via `_netdesc_safety_checks`.

[...]

PR #4232, opened 2026-09-16
API input validation phase 7: structured parameter schemas

### LINKED ISSUE #3612: API body parameters are merged into handler kwargs untyped, so wrong-typed input leaks interpreter messages as 400s

mikalstill:
Phase 7 of [PLAN-api-input-validation](docs/plans/PLAN-api-input-validation.md).
It teaches the type vocabulary to describe what is *inside* a structure, and
then describes three of them.

Until now `dict` compiled to `fields.Dict()` and `arrayofdict` to
`fields.List(fields.Dict())`, neither carrying a value schema. Every key inside
a diskspec, networkspec or videospec was therefore unvalidated in every mode,
including `enforce`.

## What the survey measured

Not read — sent. Roughly sixty probes went through the real decorator stack at
`enforce`. Eleven nested values were a recorded 500, and nine more were
accepted in silence: an unknown key, a negative disk size, a `video.model` of
the string `"yes"`. The population turned out smaller than the master plan's
row suggested — six `dict`/`arrayofdict` declarations across four operations,
not the open-ended set the row implied.

## What changed

`_field()` gained an object branch, so a rendered `properties` block compiles
to a nested marshmallow schema. The branch keys on the presence of
`properties`, not on the type being `object`, so a bare `{'type': 'object'}`
still compiles to `fields.Dict` — which is what keeps `metadata` and
`bound_claims` free-form. `additionalProperties: false` becomes
`unknown=RAISE`.

`ARGTYPES` gained `diskspec`, `arrayofdiskspec`, `networkspec`,
`arrayofnetworkspec` and `videospec`, built from three module-level constants
so a shape declared twice cannot drift — the array forms nest the same Python
object as the scalar forms.

Twenty-one nested values which were a fault or a silent acceptance now answer
400 naming the key and its index, `disk[1].size: Not a valid integer.` Two
declarations stay out and are documented as staying out: `metadata` is
free-form by design, and `bound_claims` is already guarded by hand with better
messages than a schema would produce.

## What was deliberately not narrowed

Five values that look like they should have been narrowed are still accepted.
A validator narrower than its handler is a breaking change wearing the clothes
of a correctness fix, and the published specification has to stay a true
statement about what the server backs. `network[].model` and `video.model` in
particular render raw into `libvirt.tmpl`, and the user guide recommends a
value the API reference omits, so neither can carry an enum yet.

Three real narrowings were taken and written down: a `disk[].type` outside
`disk`/`cdrom`, a caller-supplied `blob_uuid` inside a diskspec, and a diskspec
asking for neither a `size` nor a `base`.

## Two guards are handler guards, on purpose

The size-or-base guard and the explicit-null `network_uuid` guard sit in the
handler rather than in the schema, so `API_VALIDATION_MODE=warn` does not roll
them back. The null `network_uuid` case is why: it previously resolved to an
arbitrary network in the namespace, and on interface hotplug answered 200 and
created an interface on a network the caller never named. Handing that back
under the operator's rollback switch would be wrong.

The underlying defect is the lookup function's, not the vocabulary's — the
pushed-down `from_db_by_ref` reads a `None` name as *no name filter* — so it is
filed as #4223 and left for its own fix rather than masked by this phase's
schema.

## Verification

- `test_nested_sweep.py`, 99 rows across `enforce`/`warn`/`off`.
- `tools/mutate-nested-sweep.sh`, 12 mutations: 12 applied, 0 survived.
- `test_api_reference_specs.py` parses the API reference's three spec sections
  and asserts their key sets match `ARGTYPES`, so the docs cannot drift from
  the enforced schema.
- A cluster CI case boots against every documented spec key.
- Full suite passes; `pre-commit run --all-files` clean across all 11 hooks.

## Issue disposition

Fixes #3612. #528 was closed separately once its body was rewritten to the
scope that was actually still open — its own text described `use_kwargs`
coverage that phase 4 had already overtaken. #4167 is reduced, not closed: its
items 2 and 3 now answer 400, but `uefi`/`secure_boot` are `required=False`
booleans that no element schema reaches, and `POST /instances {"uefi": null}`
was re-measured as still a recorded 500. The issue carries that measurement as
a comment.

## A note on the phase itself

Decision D46 was rewritten mid-phase — marshmallow's `strict=True` refuses a
query parameter, which arrives as a string on the wire — and the rewrite
silently invalidated three other statements reasoning from it. Four separate
steps rediscovered that independently. A plan carries no index from a decision
to its dependents, which is the phase's recorded lesson and the reason its
close-out re-read the decisions as a set.


mikalstill:

All six actionable items are addressed in 2e01b505f, including the three marked `consider`.

**1 (fix), the videospec null.** Confirmed and fixed: the guards are value tests now, spelt `is None` the way `_netdesc_safety_checks` is. `is None` rather than the suggested truthiness test, deliberately — null is what the reference's nullability sentence promises about, and refusing `""` or `0` as well would be a narrowing nothing has asked for and which D50's list does not carry. A non-mapping videospec needed an explicit shape guard in front of them: the presence tests answered 400 for one by accident, since `'model' not in ['vga']` is a membership test which happens to be True. `video.model.null` and `video.memory.null` are now refused in all three columns, and a new test asserts a null `vdi` is *defaulted* rather than stored, which no status code can see. D42's guard count moves from two to three; every dependent statement of that count was corrected.

**2 (consider), `float`.** Taken. The handler's read now goes through `validation.declared_boolean()`, keyed on `fields.Boolean.truthy`/`falsy` so the check and the read cannot drift. Coercing rather than narrowing the field, because the shipped CLI and the ansible collection both parse their own command lines into a real bool, so narrowing to strict `bool` would have refused `"yes"` — which works today — for no gain. `net.float.false_string` and `net.float.true_string` are swept, the meaning is pinned by a test that checks whether an interface actually got a floating address, and the census table is corrected.

**3 (consider), the `_schema` collapse.** Taken, using the suggested discriminator: `_flatten_messages` now takes the supplied container and collapses only when the element at that path is not a `Mapping`. `{"disk": [{"size": 8, "_schema": 1}]}` answers `disk[0]._schema: Unknown field.`; a non-mapping element still answers `disk[0]: Not a valid mapping type.` Both directions have a compiler test and a mutation.

**4 (consider), the harness.** Taken. It runs the sweep once before the first mutation and aborts if it is not green, and `check()` now requires the run to have actually failed rather than only to have mentioned the row.

**5 (doc), the user guide.** Both sections now say only the listed keys are accepted, link to the reference, and say a `-D` spec must ask for something.

**6 (doc), the collection's `video`.** Filed as #4236 with `automated-fix-attempted` applied at filing, referenced from the plan's *What is left*, and the census finding notes that the close-out audit missed the omission.

**7** needs no action, agreed.

The fourth missing-test scenario was real: a `base` of the literal string `"none"` is read as no base by `util_general.noneish`, so it is a sixth spelling of the diskspec which asks for nothing. Swept as `disk.base_none_only`, named in the release note, and documented in both the reference and the user guide.

Sweep 99 → 104 rows; mutations 12 → 16, all 16 caught. Full suite 5011 tests and `pre-commit run --all-files` clean.

sfconductor:
Merged at `2e01b505fa6e`, bypassing the merge queue.

Issue #4240, opened 2026-09-17
swagger_helper() response declarations cannot express a response header


mikalstill:
`swagger_helper()` renders response declarations as three-tuples,
`(httpcode, description, sample)`, in `shakenfist/external_api/base.py`. The
`sample` becomes `examples['application/json']` and that is the whole of it:
there is no way to express a response *header*, and no way to attach a body
*schema* rather than an example.

This became concrete in `PLAN-transient-capacity-refusals` phase 4. `POST
/instances` now returns `Retry-After: 15` on its two scheduling `507`s, along
with `stage` and `transient` fields in the body. The body shape could be
published as a sample, but the header — the part a client is actually meant to
read and act on — can only be described in English prose in the description
string. A generated client cannot discover it.

Phase 4 decided (D32) not to fix this in passing, and the reasoning still
stands:

- Response declarations are validated at import time, and a malformed one
  raises `InvalidAPIDeclaration` so that **sf-api does not start**. Getting a
  format change wrong is fatal rather than cosmetic.
- Changing the tuple arity touches every endpoint in the API, not just the one
  that needed a header.

That makes it a piece of work with its own scope and its own migration, most
naturally a phase of `PLAN-api-input-validation`, which already owns the
published-specification contract.

Worth noting for whoever picks this up: `PLAN-api-input-validation`'s D4
decided to keep the `{'error': ..., 'status': ...}` body and add "no structured
field-keyed body". That decision is about *validation* failures and about
keying a body by request field name. It does not forbid a status-specific field
on a `507`, and it has nothing to say about response headers.

Until then, the `Retry-After` contract is published in prose and asserted by
`assertRefusedAtStage()` in the functional suite, and is documented at
`docs/operator_guide/capacity_refusals.md`.

PR #4241, opened 2026-09-17
Retry-After and a machine-readable transient capacity refusal

mikalstill:
Phase 4 of [PLAN-transient-capacity-refusals](docs/plans/PLAN-transient-capacity-refusals.md),
the server half. It changes what a capacity refusal *says*, not when one
happens: nothing here touches the scheduler's admission decisions, the
capacity ledger, or the refusal rate.

Today both scheduling `507`s from `POST /instances` are bare prose, so a
caller wanting to tell a refusal it should wait out from one it should not
has to pattern-match an English sentence. After this, both carry
`Retry-After: 15` and a body of
`{"error": ..., "status": 507, "stage": ..., "transient": true}`.

Phase 3 is what makes the hint honest. Before it, a node kept refusing for up
to a metrics period after its instances were destroyed, so any `Retry-After`
short enough to be useful would have been a lie. Its closeout measured the
drop at 5.3-5.6 s across three topologies, so 15 s has roughly three times
margin. The 15 is `BaseOperation.defer()`'s default delay, so the number a
client is told to wait and the number the server waits before re-examining
deferred work are defined once.

### What is marked, and what deliberately is not

`POST /instances` has four `507` branches, not the two the master plan named.
Only the two *scheduling* ones are marked: the filter pre-check, and the
capacity guard refusing every candidate. The two `CongestedNetwork` refusals
stay bare (D29) — an exhausted address pool has a horizon shaped by the
deletion halo, not by an instance teardown, so the same constant would be a
lie there. The `409` affinity refusal and the `404` are untouched, and the
fragile except-clause ordering that keeps `AffinityConstraintUnsatisfiable`
above `LowResourceException` is unchanged.

The stage rides on the exception rather than being parsed back out of the
message (D30). The scheduler already has it as a local at the raise site;
re-deriving it from `str(e)` would make the prose load-bearing at exactly the
moment we add a structured field so it stops being. No message text changed,
so `assertRefusedAtStage()`'s regexp still matches and its four callers in
`test_saturation.py` are byte-for-byte unchanged.

### The client retry is off, and the CI suite leaves it off

This reverses an instruction in the master plan, which said the suite turns it
on. It must not (D33): the suite's clients use `ASYNC_PAUSE`, so a blind retry
would spend up to 60 s inside each call the phase 2 wrapper makes, invisible
to the wait records phase 5 reads as its denominator; the wrapper retries
under a fresh instance name because the refused instance is
`enqueue_delete_due_error`'d, where a blind retry replays an identical body;
and the saturation assertions must never retry the refusal they are asserting.
The master plan's section was corrected in the planning commit.

### One plan correction worth reading

The definition of done claimed a `shakenfist-utilities` upgrade changing the
error body would fail the new unit test. Running that check rather than
reading it showed it cannot: because the helper replaces the body rather than
patching it, our fields are never derived from the upstream dict, and renaming
`status` to `status_code` upstream left every assertion passing. The real
exposure is the inverse — this `507` silently becoming the only error in the
API shaped unlike the other 305 `sf_api.error()` call sites — so
`test_sf_api_error_shape_canary` pins the upstream shape directly, and was
confirmed to be the only test that rename fails. Two other checks were greps
that did not close as written. All three now say what they mean.

### Known gaps, recorded rather than fixed

- Nothing asserts the capacity-guard branch's contract:
  `assertRefusedAtStage()`'s regexp only matches the filter branch. Provoking
  the guard needs a concurrent placement race, which is `test_saturation.py`'s
  territory.
- `swagger_helper()` renders responses as three-tuples and cannot express a
  response header, so `Retry-After` is published in prose. Widening the tuple
  is import-time-fatal if got wrong and touches every endpoint, so it is
  [#4240](https://github.com/shakenfist/shakenfist/issues/4240) against
  `PLAN-api-input-validation` instead of a side quest here.
- `assertRefusedAtStage()`'s header assertion is conditional, because the
  suite installs the client from PyPI and no released client carries
  `APIException.headers` yet. The comment names the commit and says no version
  is assigned, rather than inventing one.

### Companion

The client half is a separate PR against `client-python`, per D34 —
server first, since the suite's header assertion degrades rather than fails
against an older client. Step 4h, the closeout, is a separate PR after this
merges.

`pre-commit run --all-files` passes all eleven hooks on the final tree,
including mypy, the full unit suite, doc-link resolution and the plan-status
check.


sfconductor:
Merged at `71193ce6fc1e`, bypassing the merge queue.

PR #4247, opened 2026-09-17
Allow reserving an address in a network

mikalstill:
## What

Adds a manual address reservation to IPAM: `POST` and `DELETE` on
`/networks/<network_ref>/addresses/<address>`, backed by a new
`ReservationType.MANUAL`.

A reservation takes the address out of the pool and does nothing else — no
interface, no DHCP entry, no routing — and is held until it is explicitly
released, because nothing in the cluster owns it.

## Why

Shaken Fist allocates from a network's block at random and only knows about the
addresses it allocated itself. An address something outside the cluster owns — a
keepalived VIP inside a guest, written into a deployment's own configuration — is
free as far as IPAM is concerned, so eventually it is handed to an interface and
two things answer for it.

That is not hypothetical. In kerbside-patches CI run
[35202102331](https://github.com/shakenfist/kerbside-patches/actions/runs/35202102331)
the VIP a kolla deployment configures for itself (`10.0.2.3`) was allocated to the
CI runner's own interface on the test network. kolla-ansible's prechecks ping the
VIP, the runner answered, and the deploy refused to start. Sibling jobs in the same
run drew `.29`, `.98`, `.128` and `.223` and passed — a one-in-253 collision.

## Contract

- An address already in use is refused with `409` rather than taken over. A
  deletion-halo reservation *is* evicted, since the caller named this address
  rather than asking for any free one (the same reasoning as #4059).
- Only a manual reservation can be released this way; a gateway or an instance's
  address answers `403`, because handing it out again while it is still in use is
  the collision this exists to prevent.
- Releases go to the deletion halo like any other, so an address is not
  immediately re-allocatable.
- Clusters advertise the `reserve-addresses` capability, so a client can tell a
  cluster that refuses the reservation from one that never offered it.

## Testing

- Ten endpoint tests in `test_network.py`, including the property the feature
  exists for: reserve an address, then 250 random allocations never return it.
- Four cluster CI tests in `test_network_lifecycle.py`, including an instance that
  asks for the reserved address by name and gets a `409`.
- Full `tox -e py3`, `tox -e docs` and `pre-commit run --all-files` pass.

## Rollout

**Merge ordering**: shakenfist/client-python#400 (the `sf-client` verbs)
has already merged, which is what the cluster CI tests here need.
`setup-test-environment` checks out client-python's default branch and
builds a wheel from it rather than installing `shakenfist-client` from
PyPI, so `reserve_network_address` / `release_network_address` are present
in the merge queue and the tests need no capability gate.

shakenfist/actions#83 (the CI playbooks, which fail soft against a cluster
without this) depends on this landing and the cluster being upgraded.

Operators upgrading past this commit must run `sf-ctl ensure-mariadb-schema`
before rolling the daemons: it adds a value to the
`ipam_reservations.reservation_type` ENUM, and enum growth deliberately does
not move a table version, so `sf-database` will start against an
unreconciled database and fail every reservation at runtime instead. This is
called out in the release note.


mikalstill:
Round 1 addressed in 5cc3813.

**Item 1 (fix)** — confirmed and fixed. A `MANUAL` reservation in the floating network really was reaped: it is on none of the reaper's protected lists, and its owner is the floating network itself, which is alive, so the mid-teardown guard did not save it either. Protected explicitly rather than refusing the floating network at the API, since holding a floating address for a device Shaken Fist does not manage is the same need the feature exists for. New unit test, mutation-tested by removing the guard — it fails with `a manual reservation in the floating network was reaped`.

**Item 3 (fix)** — no code needed. shakenfist/client-python#400 has merged, and `setup-test-environment` checks out client-python's default branch and builds a wheel rather than installing from PyPI, so both methods are present in the merge queue. Merge ordering is now stated in the PR description.

**Item 2 (document)** — release note now says the release grows the `reservation_type` ENUM and that `sf-ctl ensure-mariadb-schema` must run first. I did not take the version-bump alternative: `mariadb.py`'s header documents the opposite convention deliberately, and `_native_enum_columns()` exists precisely so enum growth needs no hand-written migration. A one-off bump here would contradict that, so the requirement is documented instead.

**Considers taken** — 4 (non-atomicity recorded next to the type check), 5 (guarded re-read, 409 instead of a 500), 6 (both audit events moved ahead of the work so refused attempts are recorded; `comment` carried in the POST event), 8 (250 → 50 allocations), 9 (import block sorted).

**Considers declined** — 7: checked, and `_compile_parameters()` collects path parameters as names only, so format validators run on body and query alone; the `ipv4` token would publish a check the server does not run there. 11: the decorator chain forces `= None` defaults, so honest annotations are `Optional[str]` and contradict what the handlers require, and `sf_api.error()` is untyped so the return is `Any` — better done when the file joins the mypy rollout wholesale.

All seven listed test-coverage gaps now have tests, including the reaper case, IPv6 → 400, network/broadcast → 409, cross-namespace → 404 (positive-controlled so the 404 is genuinely from ownership), non-string `comment` → 400, and halo release → 403.

mikalstill:
Round 2 addressed.

**Item 1 (fix)** — confirmed, and the diagnosis was right in a way round 1's
fix was not. `requires_network_active` really does make the floating-network
half of this feature unreachable: `_apply_create_on_network_node()`
early-returns for the floating network before the `STATE_CREATED` transition,
so it sits in `initial` forever. Checked against a live cluster, which reports
`state: initial`. The reaper protection added last round was therefore
defending something that could never exist.

Fixed with a new `requires_network_not_dead` decorator rather than by
special-casing `FLOATING_NETWORK_UUID`, because `created` was the wrong gate
for either handler, not just for the floating network: a reservation only
touches IPAM, which exists from `Network.new()`, so an ordinary network in the
window between `new()` and the net worker applying the create was equally
wrongly refused. Dead is still refused, and for a sharper reason than
tidiness — `Network.__init__` builds a *deleted* network's IPAM with
`in_memory_only=True`, so a reservation against one is accepted, held in
process memory and lost. That is recorded in the decorator's docstring.

Four new unit tests, each mutation-tested: restoring
`requires_network_active` fails the floating-network and not-yet-created tests
with `200 != 406`, and making the dead check a no-op fails the dead-network
test with `406 != 200`.

**Item 3 (fix)** — no code needed, as last round. shakenfist/client-python#400
has merged, and `setup-test-environment` builds a wheel from client-python's
default branch rather than installing from PyPI, so both methods are present
in the merge queue. The ordering is in the PR description.

**Item 6 (document)** — the reservation types table in
`docs/operator_guide/database.md` gains the `manual` row, plus the suggested
paragraph on the halo contrast: a `deletion-halo` entry is evicted by a
request naming that address, and a `manual` reservation is not.

**Item 5 (consider) — taken.** `comment` is bounded at 255 characters in the
handler, matching the namespace metadata bound in `auth.py`, with a 400 on
overflow and tests at and over the limit. Not a `pattern` in the declaration's
constraints dict: the supported keys are `minimum`/`maximum`/`pattern`, and
the only pattern that expresses a length cap is `^.{0,255}$`, whose `.` does
not match a newline under ECMA-262 — that would publish a rule the server does
not enforce, which is the thing the declaration checker exists to prevent. Not
on `IPAMReservation` either: that model also reads existing rows back, so a
bound there would turn an old long comment into a read failure.

**Item 2 (consider) — deferred, and now tracked.** Agreed on the analysis. The
fix needs an `expected_type` on the `ReleaseAddress` RPC, which means a proto
change, regenerated stubs, changes in `mariadb.py`, the database handler and
`ipam.release()`, plus a decision about an `sf-database` from before the field
silently ignoring it during a mixed-version window. Filed as #4261 with the
suggested `WHERE` clause and the skew note; the handler comment now points at
the issue rather than standing alone.

**Item 4 (consider) — declined.** The suggested 200-on-re-POST returns the
existing reservation when it is `MANUAL` and owned by this network, but *every*
manual reservation in a network is owned by that network — `user_uuid` is the
network UUID, which is what the round-1 test output shows. So the test cannot
distinguish "this is my earlier call" from "somebody else in my namespace got
here first": it would answer 200 to a caller who did not make the reservation,
and hand them somebody else's comment as if it were their own. The 409 is the
honest answer to a question the server cannot resolve. What a retrying client
actually needs is documented instead, in
`docs/developer_guide/api_reference/networks.md`: treat the 409 as ambiguous
and read the reservation back with `GET /networks/{network_ref}/addresses`,
where a manual reservation carrying its own comment is the confirmation it
wanted. If per-caller ownership is wanted later it needs a real owner on the
reservation, not an inference from the network UUID.

**Test coverage** — the conditional cluster CI assertion is tightened.
`assertNotEqual('manual', ...)` also accepted the address turning into some
other *live* reservation type, which is not a release; it now asserts the
address is either absent or specifically a `deletion-halo`, the only other
shape a release may legitimately leave behind. The floating-network gap is
covered by the new unit tests above. The remaining listed gaps (a non-system
caller in the owning namespace, a concurrency test for item 2, a congestion
sweep test) ride with #4261 or are covered by the existing cluster CI test,
which authenticates as an ordinary namespace.

**Item 8's nit** — accurate, and left as is: the DDL gains `MANUAL` because
`sa.Enum` stores names, while the user-facing string really is `manual`, which
is what the release note is describing.

Full `tox -e py3` (5046 tests), `tox -e docs` and `pre-commit run --all-files`
pass.

sfconductor:

The conductor will merge this at `e87948adb840` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `e87948adb840`, bypassing the merge queue.

PR #4289, opened 2026-09-21
Close out CI cloud sizing phase 4.

mikalstill:
Closes out phase 4 of [PLAN-ci-cloud-sizing.md](docs/plans/PLAN-ci-cloud-sizing.md), reading the merge runs on the reshaped `slim-tier` and taking the three decisions that were waiting on them. Two commits: the measurement, then the close-out.

`shakenfist/actions#88` merged as `f78576e` on 2026-09-19, taking all three `slim-tier` nodes from `cpu: 4` to `cpu: 6`. Six `Debian 12 tier` merge runs have completed on that shape, read here against the four immediately before it.

## 4d: the prediction held, and then stopped mattering

|  | Before (4 runs) | After (6 runs) |
|---|---|---|
| Ledger (primary, sf-1, sf-2) | 3, 3, 6 | **6, 6, 12** |
| Node-runs at `peak frac` 1.000 | 12 of 12 | 3 of 18 |
| `sufficient_idle_cpu` drops | 30, 23, 34, 28 | 5, 4, 9, 4, 4, 14 |
| Guard refusals | 179, 181, 173, 182 | 147, 153, 164, 176, 185, 172 |

The ledger prediction was exact in every run, which rules out falsification reading 1: the infra-role reservation is the absolute 4 threads 4a measured. The allocation bound was relieved decisively — saturation went from every node in every run to 3 node-runs in 18, and scheduler drops at `sufficient_idle_cpu` fell by a factor of 4.3. The guard's refusal count moved about 7%, inside the run-to-run spread, and stayed `demand`-dimension alone. That is falsification reading 2.

**The finding the plan did not predict is in the split underneath that count:**

|  | Load alone already over | Feedforward carried it over |
|---|---|---|
| Before | 152, 170, 160, 156 (**85-94%**) | 27, 11, 13, 25 |
| After | 53, 89, 48, 114, 116, 101 (**29-65%**) | 93, 64, 116, 62, 69, 71 |

Doubling the bound did what the arithmetic said: measured load is under it far more often. The refusal count held anyway, because `expected_demand` accumulates `cpus × SCHEDULER_DEMAND_PER_VCPU` per admitted placement — the guard admits until demand fills whatever bound it is given, then refuses at the same rate. **The refusal count is close to invariant under node size; more vCPU changes which placements are refused, not how many.** That argues for the sibling plan's guard work over D3's revision to 8 vCPU.

No structural minimum broke: `test_network_lifecycle`, `test_affinity`, `test_binary_affinity_prefers_the_tagged_node` and `test_database_tier` each report a verdict other than `SKIPPED` in all six runs, read from stestr output rather than inferred from job colour. Two of the six jobs failed, both only on `test_cluster_resources_measured_drops_after_delete`, which also failed before the reshape — 2 of 6 after against 1 of 4 before.

## 4e: `slim-primary` is deliberately not reshaped

A decision, not a deferral. Three `Debian 12 cluster` runs from the same window:

| Run | Ledgers | Guard refusals | Dimension | `sufficient_idle_cpu` drops |
|---|---|---|---|---|
| `35485863474` | 3, 6, 6, 6, 6 | 225 | demand, 225 alone | 12 |
| `35490705886` | 3, 6, 6, 6, 6 | 210 | demand, 210 alone | 8 |
| `35495432148` | 3, 6, 6, 6, 6 | 203 | demand, 203 alone | 17 |

There is no allocation starvation to relieve — `p90 frac` sits at 0.33 to 0.50 and the drops are already at the level `slim-tier` only reached *after* its reshape. Its refusals are higher in absolute count than the reshaped tier's on a *larger* cluster ledger (27 against 24), which is 4d's invariance restated on a second topology. And the spend is larger: five nodes at +2 vCPU is ten under-cloud vCPU across two hand-checked diffs, since F4 records the two `slim-primary` files are not structurally identical.

One real finding carried to phase 5 rather than answered here: its network node sits at ledger 3 and saturated in one of the three runs. That is the infra-reservation pathology, but the census says `demand` rather than `cpus` is what refuses, so reshaping four other nodes around it is the risk this phase's own risk table names first.

## 4f: tier timeout 70 → 60

`Run functional tests` took 42, 29, 36, 27, 32 and 30 minutes across the six runs — mean 32.7, worst 42. Sixty leaves 30% headroom over the worst observed run, matching the other three matrix entries. The comment above it explained the bump by the tier having half `slim-primary`'s hypervisor capacity, which the reshape made untrue; it is rewritten to say what is still true (three nodes rather than five) and to record the durations the 60 is sized against.

**This is the one behavioural change in the PR.** Six runs is a small sample, and the comment it replaces claimed "30-50 minutes" from a similar one. If a slow run trips the new limit, that is where it will show.

## 4g: close-out

Phase 4 goes `Complete` in the master plan's Execution table and the index row moves to 5 of 8. *What phase 5 inherits* records four things: the headroom band cannot be expressed as a function of node size; the allocation side now has assertable slack (`p90 frac` 0.17 to 0.58 against the proposed 0.70 upper bound); the infra-reservation pathology survives on `slim-primary`; and `slim-tier` has no slack left above its structural minimum.

The `Merged` cell records `870a5fbec` (#4202), which landed 4a to 4c. This close-out cannot know its own merge commit — that is appended when phase 5 is planned, per the convention that a phase does not record its own merge.

## Verification

- `pre-commit run --all-files` — 11 hooks Passed, exit 0.
- `python3 tools/check-plan-status.py` — statuses, index arithmetic and phase links agree.
- All eleven Definition of done items run rather than read. Items 3, 4, 9 and 10 re-run at close-out: three `ci-topology-slim-*` files in `shakenfist/actions`, two `len(nodes) < 3` guards in `test_scheduler.py`, `timeout_minutes: 60`, and no configuration default moved.

Related: shakenfist/actions#88 (the reshape), #4271 and shakenfist/private-ci#68 (the claim-sizing cross-references).


sfconductor:

The conductor will merge this at `2bb602a5db3a` the next time no merge group is being built, so that bypassing the queue does not discard somebody else's CI run. The arming lapses after 6 hours, or immediately if a new commit is pushed.

sfconductor:
Merged at `2bb602a5db3a`, bypassing the merge queue.

